PR #44, opened 2019-04-25
Add unit tests for org.dromara.soul.common.utils.UUIDUtils

JohnLBergqvist:
Hi,

I've analysed your code base and noticed that `org.dromara.soul.common.utils.UUIDUtils` in the `soul-common` module is not fully tested.

I've written some tests for the functions in this class with the help of [Diffblue Cover](https://www.diffblue.com/opensource).

In addition, I noticed that instantiating `UUIDUtils` via the public constructor will never cause the block at line 83 to be entered. 

Hopefully these tests should help you detect any regressions caused by future code changes. If you would find it useful to have additional tests written for this repository, I would be more than happy to look at other particular classes that you consider important.

### DIFF
--- a/soul-common/pom.xml
+++ b/soul-common/pom.xml
@@ -43,6 +43,34 @@
             <artifactId>slf4j-api</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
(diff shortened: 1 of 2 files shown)


yu199195:
thanks，please look at other particular classes

Issue #511, opened 2020-12-04
refactor chinese to english in soul.sql


yu199195:
Hi :
All communitys, we should  refactor chinese to english in soul.sql 

you can find it in :   https://github.com/dromara/soul/blob/master/script/soul.sql  and  https://github.com/dromara/soul/blob/master/soul-admin/src/main/resources/META-INF/schema.sql 

The Chinese description in it is changed to English.

Please read the Contribution Guideline before submitting the PR  .  thanks !

yiwenlong:
get

evasnowind:
assign me

funpad:
手都这么快的么

PR #545, opened 2020-12-04
refactor chinese to english in soul.sql and schema.sql(#511)

evasnowind:
tested the  new sql in my local database,  no error. please check.


dengliming:
Repeated with other PR.

Issue #570, opened 2020-12-05
Only use SecretProperties or remove it


dengliming:
## Describe the bug

This bean is defined as below code:
```
/**
     *  egister cipherUtils in spring ioc.
     *
     * @return CipherUtils
     */
    @Bean
    @ConditionalOnProperty(prefix = "soul.aes.secret", value = "key", havingValue = "2095132720951327")
    public SecretProperties secretProperties() {
        return new SecretProperties();
    }
```
But we currently only use in this way. and the key is inconsistent.
`org.dromara.soul.admin.service.impl.DashboardUserServiceImpl`
`org.dromara.soul.admin.controller.DashboardUserController`
```
 @Value("${aes.secret.key:2095132720951327}")
    private String aesKey;
```

## Environment

- Soul version(s): [e.g. 2.2.1]

## Expected behavior

Keep consistent. Only use `SecretProperties`

## Debug logs

<details>
<summary>Click to expand</summary>

```
replace this line with your debug logs
```
</details>

jjnnzb:
Get this task.

dengliming:
@jjnnzb Thanks for your contribution.

dengliming:
closed by [https://github.com/dromara/soul/pull/593](https://github.com/dromara/soul/pull/593)

PR #593, opened 2020-12-06
Fix issue #570, only use SecretProperties.

jjnnzb:
Only use SecretProperties, keep consistent, and modify related tests.


soul-admin/src/main/java/org/dromara/soul/admin/config/SecretConfiguration.java
```diff
@@ -17,26 +17,30 @@
...
 @Configuration
 public class SecretConfiguration {
-
+    
     /**
-     *  Register cipherUtils in spring ioc.
+     * Register key for CipherUtils in spring ioc.
      *
-     * @return CipherUtils
+     * @return key
```
dengliming, line 37: SecretProperties

jjnnzb: 好的，改一下

dengliming (approved):

PR #609, opened 2020-12-07
[#607]add springmvc test case.

tydhot:
Make sure that:


soul-client/soul-client-http/soul-client-springmvc/src/test/java/org/dromara/soul/client/springmvc/init/SpringMvcClientBeanPostProcessorTest.java
```diff
@@ -36,28 +44,51 @@
...
+        Assert.assertEquals(registerNum, 1L);
     }
 
     @Test
-    public void testNormalBeanProcess() {
+    public void testNormalBeanProcess() throws InterruptedException {
+        CountDownLatch countDownLatch = new CountDownLatch(1);
+        registerNum = 0L;
         springMvcClientBeanPostProcessor.postProcessAfterInitialization(new Object(), "normalBean");
+        countDownLatch.await(5, TimeUnit.SECONDS);
```
dengliming, line 90: If you don't call `countDownLatch.countDown(); ` It will always await for 5s.

tydhot: but I just call countDownLatch.await(), it is keep await and never run again .

dengliming (approved):

PR #614, opened 2020-12-08
[#600] modifies all entity data classes constructed as Builder schemas

nuo-promise:
Make sure that:


soul-admin/src/main/java/org/dromara/soul/admin/dto/SelectorDTO.java
```diff
@@ -81,4 +85,21 @@
...
+        this.name = name;
+        this.matchMode = matchMode;
+        this.type = type;
+        this.sort = sort;
+        this.enabled = enabled;
+        this.loged = loged;
+        this.continued = continued;
+        this.handle = handle;
+        this.selectorConditions = selectorConditions;
+    }
```
tydhot, line 104: why not use the annotation AllArgsConstructor?

nuo-promise: Yeah, this one doesn't have a parent class and I don't have to do that, so Let me change it

PR #636, opened 2020-12-09
Refactor RpcTypeEnum & RuleHandle generate method.

yiwenlong:
* Try make RpcTypeEnum simple and clear.
* Add RuleHandle interface.
* Add RuleHandle factory to generate RuleHandle object with given RpcTypeEnum.


soul-common/src/main/java/org/dromara/soul/common/dto/convert/rule/impl/DivideRuleHandle.java
```diff
@@ -0,0 +1,63 @@
...
+    @Override
+    public RuleHandle init(final String path) {
+        this.loadBalance = RuleHandleConstants.DEFAULT_LOAD_BALANCE.getName();
+        this.retry = RuleHandleConstants.DEFAULT_RETRY;
+        return this;
+    }
+
+    @Override
+    public String toJson() {
+        return JsonUtils.toJson(this);
```
yu199195, line 61: use gsonUtils

soul-common/src/main/java/org/dromara/soul/common/dto/convert/rule/RuleHandle.java
```diff
@@ -17,33 +17,25 @@
...
 
     /**
-     * loadBalance.
-     * {@linkplain org.dromara.soul.common.enums.LoadBalanceEnum}
+     * It will be call when RuleHandleFactory to generate a RuleHandle object.
+     * @param path  path.
+     * @return      it's self.
      */
-    private String loadBalance;
+    RuleHandle init(String path);
```
yu199195, line 34: the method is createDefualt(String path)

yiwenlong: Do you mean it is more appropriate to use `createDefualt` instead of `init` as the method name?

soul-common/src/main/java/org/dromara/soul/common/dto/convert/rule/impl/DivideRuleHandle.java
yiwenlong: ok

yu199195 (approved):

PR #637, opened 2020-12-09
Add retry support of issue 70

yzhuomaoy:
Add retry support using webClient and fix issue 70 which  mention.


tydhot:
checkstyle is failed

PR #638, opened 2020-12-09
Refactor: remove the `else` statement, try make the code clear.

yiwenlong:
Remove the `else` statement, try make the code clear.


tydhot:
There may be many scenes like this. Can you sort out places like this completely

yiwenlong:
> There may be many scenes like this. Can you sort out places like this completely

Good idea, I will modify as suggested.

tydhot (approved):

PR #642, opened 2020-12-09
Add webClient retry support of dividePlugin set, and issue #70 resolved

yzhuomaoy:
Add webClient retry support of dividePlugin set, and issue 70 resolved


tydhot:
Today there is something wrong with the ci server, you can merge the latest code and push again, I have make the await time in test case longer.

yu199195 (approved):

PR #645, opened 2020-12-09
Refactor: remove some `else if` statements, try make the code looks m…

yiwenlong:
Remove some `else if` statements, try make the code looks more simple and clear.


soul-common/src/main/java/org/dromara/soul/common/utils/VersionUtils.java
```diff
@@ -84,11 +84,10 @@ public static String getVersion(final Class<?> cls, final String defaultVersion)
...
             while (file.length() > 0 && !Character.isDigit(file.charAt(0))) {
                 i = file.indexOf("-");
-                if (i >= 0) {
-                    file = file.substring(i + 1);
-                    continue;
+                if (i < 0) {
+                    break;
                 }
-                break;
+                file = file.substring(i + 1);
```
nuo-promise, line 90: Can We use recursion here？

yu199195 (approved):

Issue #471, opened 2020-12-02
Function list for dubbo


KevinClair:
### Function List for dubbo
* Dubbo Tag Route；
* Dubbo support SoulException;
* Dubbo support ValidationException;

#### Is your feature request related to a problem? Please describe

No!

#### Describe the solution you'd like
* When api throws SoulException(“User's infomation has exist.”)，response:
``` java
{
    "code": 500,
    "message": "User's infomation has exist.",
    "data": "User's infomation has exist."
}
```
* When ValidationFilter validate the request param of method,throws ValidationException：
``` java
{
    "code": 500,
    "message": "Property must not be empty",
    "data": "Property must not be empty"
}
```
* [Dubbo tag route](http://dubbo.apache.org/zh/docs/v2.7/user/examples/routing-rule/)

yu199195:
每个功能一个pr，都提交一下吧

KevinClair:
#### 功能列表
- [x] Dubbo标签路由
- [x] soul对dubbo服务自定义异常的支持
- [x] soul对dubbo服务参数校验异常的支持

Issue #615, opened 2020-12-08
add test case for module soul-client


tydhot:
- [ ] soul-client-alibaba-dubbo @0x12FD16B
- [ ] soul-client-apache-dubbo @nuo-promise
- [x] soul-client-springcloud @plutokaito
- [x] soul-client-sofa @HoldDie 

Please read the [Contribution Guideline](https://dromara.org/website/en-us/docs/soul/contributor.html) before submitting the PR

You can take reference soul-client-springmvc's test case.

branchen:
get soul-client-alibaba-dubbo

plutokaito:
get soul-client-springcloud

lw1243925457:
get soul-client-apache-dubbo

nuo-promise:
get soul-client-sofa

tydhot:
> get soul-client-sofa

猫大不是说有别的安排给你么。。。

HoldDie:
get soul-client-sofa

HoldDie:
那我来吧，嘎嘎

HoldDie:
啪的一下子，很快奥！

PR #736, opened 2020-12-14
Dubbo support ValidationException#471

KevinClair:
//  Feature #471 
### How to use it.
* Set the validation of @Service,for example
```java
@Service(validation = "soulValidation")
public class MainServiceImpl implements IMainService {

    @Override
    @SoulDubboClient(path = "/test", desc = "test method")
    public String test(@Valid HelloServiceRequest name) throws SoulException {
        if (true){
            throw new SoulException("Param binding error.");
        }
        return "Hello " + name.getName();
    }
}
```
* The request param
```java
public class HelloServiceRequest implements Serializable {

    private static final long serialVersionUID = -5968745817846710197L;

    @NotEmpty(message = "name cannot be empty")
    private String name;

    @NotNull(message = "age cannot be null")
    private Integer age;

    public String getName() {
        return name;
    }

    public void setName(String name) {
        this.name = name;
    }

    public Integer getAge() {
        return age;
    }

    public void setAge(Integer age) {
        this.age = age;
    }
}
```
* Send request
```java
{
    "name": ""
}
```
* Response
```java
{
    "code": 500,
    "message": "Internal Server Error",
    "data": "name cannot be empty,age cannot be null"
}
```

Make sure that:


dengliming (commented):
Can you also add some testcases for your custom `Validator`?

KevinClair:
> Can you also add some testcases for your custom `Validator`?

But I think this mission has been claimed here #615 ,so do I need add a new one?

dengliming:
Yes. You can add test for `AlibabaDubboClientValidator ` and `ApacheDubboClientValidator `

KevinClair:
> Yes. You can add test for `AlibabaDubboClientValidator ` and `ApacheDubboClientValidator `

OK, I will try it.

dengliming (approved):

PR #751, opened 2020-12-15
Refactor: Move the build do object method to its own classes. 

yiwenlong:
I do this refactor with the following reason:
* The responsibilities of the class should be made clearer. Service component should focus on the service flow and the DO classes should known the details that how to build the object of its self.


soul-admin/src/main/java/org/dromara/soul/admin/entity/AppAuthDO.java
```diff
@@ -62,12 +64,12 @@
...
 
     /**
-     * build appAuthDO.
+     * Build AppAuthDO object with given AppAuthDTO object.
      *
      * @param appAuthDTO {@linkplain AppAuthDTO}
      * @return {@linkplain AppAuthDO}
      */
-    public static AppAuthDO buildAppAuthDO(final AppAuthDTO appAuthDTO) {
+    public static AppAuthDO buildWith(final AppAuthDTO appAuthDTO) {
```
yu199195, line 72: this method name maybe refator

yiwenlong: Do you think `create` will be a good method name?

lmhmhl: 茴香豆的茴有几种写法

yu199195 (approved):

Issue #510, opened 2020-12-04
Add test cases for soul-admin controller


dengliming:
**_controller_**
- [ ] AppAuthController @magicalxiaochen
- [x] ConfigController @HoldDie
- [x] DashboardUserController @operaxxx
- [x] IndexController @lw1243925457
- [x] MetaDataController @fcwalker
- [x] PlatformController @wangwenjie666
- [x] PluginController @guanyushen
- [x] PluginHandleController @charlesgongC
- [x] RuleController @shijie666
- [x] SelectorController @cchenxi
- [x] SoulClientController @branchen

Please read the [Contribution Guideline](https://dromara.org/website/en-us/docs/soul/contributor.html) before submitting the PR

HoldDie:
1

lw1243925457:
1

operaxxx:
get

magicalxiaochen:
get AppAuthController

branchen:
get IndexController

shijie666:
get RuleController

wangwenjie666:
get PlatformController

fcwalker:
get MetaDataController

branchen:
get SoulClientController

cchenxi:
get SelectorController

guanyushen:
get  PluginController

charlesgongC:
get PluginHandleController

dengliming:
done

PR #753, opened 2020-12-16
Add test cases for AppAuthController of soul-admin(#510)

magicalxiaochen:
Fix: #510

### DIFF
--- a/soul-admin/src/main/java/org/dromara/soul/admin/query/AppAuthQuery.java
+++ b/soul-admin/src/main/java/org/dromara/soul/admin/query/AppAuthQuery.java
@@ -41,10 +41,6 @@ public class AppAuthQuery implements Serializable {
 
     private String phone;
 
-    private Integer currentPage;
-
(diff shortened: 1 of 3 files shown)


soul-admin/src/main/java/org/dromara/soul/admin/query/AppAuthQuery.java
```diff
@@ -41,10 +41,6 @@
 
     private String phone;
 
-    private Integer currentPage;
-
-    private Integer pageSize;
-
```
dengliming, line 47: Why remove this?

magicalxiaochen: Because it repeats with PageParameter

dengliming: OK

dengliming (approved):

PR #761, opened 2020-12-16
Optimize the setup selectors of divide plugin

dengliming:

Make sure that:

### DIFF
--- a/soul-admin/src/test/resources/META-INF/schema.sql
+++ b/soul-admin/src/test/resources/META-INF/schema.sql
@@ -14,11 +14,7 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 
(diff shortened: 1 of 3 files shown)


soul-admin/src/test/resources/META-INF/schema.sql
```diff
@@ -14,12 +14,8 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 
-SET MODE MySQL;
-
 CREATE Schema IF NOT EXISTS  `soul`;
 
-USE `soul`;
```
yu199195, line 21: not use soul databases？

dengliming: no need this for h2

PR #850, opened 2020-12-22
Issue#849 Rewrite RandomLoadBalance class, and sort the upstreamList beforehand. 

lmhmhl:


tydhot:
I think this is a good opportunity to improve the unit testing of this strategy

dengliming:
Can you also add some test to cover your changes?

PR #918, opened 2020-12-30
compatible with h2 storage

dmsolr:

Make sure that:

1. compatible with h2
2. change default storage to H2


soul-admin/src/main/resources/application-local.yml
```diff
@@ -25,16 +25,19 @@ spring:
...
     suffix: .html
   datasource:
-    url: jdbc:mysql://localhost:3306/soul?useUnicode=true&characterEncoding=utf-8
-    username: root
-    password:
-    driver-class-name: com.mysql.jdbc.Driver
+    url: jdbc:h2:mem:~/soul;DB_CLOSE_DELAY=-1
+    username: sa
+    password: sa
+    driver-class-name: org.h2.Driver
```
dengliming, line 31: If we can support like this would be better. 
```
datasource:
  select: mysql
  mysql:
    url: jdbc:mysql://localhost:3306/soul?useUnicode=true&characterEncoding=utf-8
    username: root
    password:
    driver-class-name: com.mysql.jdbc.Driver
  h2:
    url: jdbc:h2:mem:~/soul;DB_CLOSE_DELAY=-1
    username: sa
    password: sa
    driver-class-name: org.h2.Driver
```

dengliming:
Let's separate multiple database configurations for now.

dengliming (approved):

PR #932, opened 2021-01-03
refactor client register code

Severezh:

Make sure that:

### DIFF
--- a/soul-admin/src/main/java/org/dromara/soul/admin/controller/SoulClientController.java
+++ b/soul-admin/src/main/java/org/dromara/soul/admin/controller/SoulClientController.java
@@ -17,10 +17,11 @@
 
 package org.dromara.soul.admin.controller;
 
+import org.dromara.soul.admin.dto.MetaDataDTO;
 import org.dromara.soul.admin.dto.SpringCloudRegisterDTO;
 import org.dromara.soul.admin.dto.SpringMvcRegisterDTO;
-import org.dromara.soul.admin.dto.MetaDataDTO;
 import org.dromara.soul.admin.service.SoulClientRegisterService;
+import org.dromara.soul.common.constant.RegisterConstants;
 import org.springframework.web.bind.annotation.PostMapping;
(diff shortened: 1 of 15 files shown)


soul-admin/src/main/java/org/dromara/soul/admin/controller/SoulClientController.java
```diff
@@ -52,7 +53,7 @@ public SoulClientController(final SoulClientRegisterService soulClientRegisterSe
      * @param springMvcRegisterDTO the spring mvc register dto
      * @return the string
      */
-    @PostMapping("/springmvc-register")
+    @PostMapping(RegisterConstants.HTTP_PATH)
```
tydhot, line 56: why spring mvc is http path?

soul-common/src/main/java/org/dromara/soul/common/constant/RegisterConstants.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * constants for register.
+ *
+ * @author severez
+ */
+public final class RegisterConstants {
+
+    public static final String COMMON_CONTEXT_PATH = "/soul-client";
+
+    public static final String HTTP_PATH = "/springmvc-register";
```
tydhot, line 29: I don't think http path can be "/springmvc-register".

soul-common/src/main/java/org/dromara/soul/common/enums/RpcTypeEnum.java
```diff
@@ -37,46 +38,48 @@
     /**
      * Http rpc type enum.
      */
-    HTTP("http", true),
+    HTTP("http", RegisterConstants.HTTP_PATH, true),
```
tydhot, line 41: The RPC type here should not be related to the specific path, especially when HTTP can't correspond to a specific path

tydhot (changes requested):

tydhot:
I think the rough changes here are not very appropriate and need more careful thinking

PR #976, opened 2021-01-14
Open httpclient plugin in soul-bootstrap pom.xml

### LINKED ISSUE #975: Little question for soul-bootstrap pom.xml
## Question

In soul-bootstrap pom.xml 
image

I think soul-spring-boot-starter-plugin-httpclient It's a very important plugin. 

It shouldn't put in 
 < ! --if you use http proxy start this -- >
  ………………
  < ! --if you use http proxy end this -- >

Because For beginners, it may be commented out together with plugin-divide. 

so I think It's should be move to other place.

lxl910128:

fix #975 

If It's not a bug , please close me


dengliming (approved):
Agree with keeping it open in pom.xml. But it's not a bug.

PR #982, opened 2021-01-15
fix pom property mismatch(#981)

### LINKED ISSUE #981: jdk.version property in soul's pom mismatch
## Describe the bug
in the root pom, soul define `jdk.version` in <properties>:
```xml
    <properties>
        <jdk.version>1.8</jdk.version>
......
</properties>
```

But use `${java.version}` to give the version of java compile in maven plugin:
```xml
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-compiler-plugin</artifactId>
                <version>${maven-compiler-plugin.version}</version>
                <configuration>
                    <source>${java.version}</source>
                    <target>${java.version}</target>
                    <encoding>${project.build.sourceEncoding}</encoding>
                </configuration>
            </plugin>
```

They are mismatched.

## Environment

- Soul version(s): master branch

## Expected behavior
recommend to use `java.version`  in all places.
Because  right now `jdk.version` is not used but `java.version` is used in root pom and soul-examples's pom.

evasnowind:
Fixes #981  

- [X] You have read the [contribution guidelines](https://dromara.org/en-us/docs/soul/contributor.html).
      - this issue need not new test case. If not , please let me know. Thx.
- [X] Your local test passed `mvn clean install -Dmaven.javadoc.skip=true`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -83,7 +83,7 @@
     </parent>
 
     <properties>
-        <jdk.version>1.8</jdk.version>
+        <java.version>1.8</java.version>
         <hutool.version>4.0.8</hutool.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <jasypt.version>1.9.2</jasypt.version>


dengliming (approved):
LGTM  /thanks.

Issue #990, opened 2021-01-16
the selector added through soul-admin is not checked


augfool:
## Describe the bug

The survival of the divide selector added through soul-admin is not checked.
UpstreamCheckService#check

## Environment

- Soul version(s): [2.2.1]

## Steps to reproduce
1. start soul-admin, soul-bootstrap
2. add divide selector  that cannot be connected through soul-admin
3. GET http://localhost:9195/http/order/findById?id=3 
![image](https://user-images.githubusercontent.com/7943753/104817562-fb393d00-585c-11eb-936a-56571cfc1f11.png)

## Expected behavior

The selector data table is updated and the selector update event is published.

## Debug logs

![image](https://user-images.githubusercontent.com/7943753/104818022-e5794700-585f-11eb-858e-1f3f0dc265e4.png)

<details>
<summary>Click to expand</summary>

```
replace this line with your debug logs
```
</details>

augfool:
pr: https://github.com/dromara/soul/pull/991

PR #991, opened 2021-01-16
fix the divide selector added through soul-admin is not checked (#990)

augfool:
* delete UpstreamCheckService dependence on SelectorService
    * reduce one select operation in UpstreamCheckService#updateSelectorHandler
    * modify SelectorServiceTest, UpstreamCheckServiceTest

Make sure that:

### DIFF
--- a/soul-admin/src/main/java/org/dromara/soul/admin/service/impl/SelectorServiceImpl.java
+++ b/soul-admin/src/main/java/org/dromara/soul/admin/service/impl/SelectorServiceImpl.java
@@ -44,9 +44,11 @@ import org.dromara.soul.admin.vo.SelectorConditionVO;
 import org.dromara.soul.admin.vo.SelectorVO;
 import org.dromara.soul.common.dto.ConditionData;
 import org.dromara.soul.common.dto.SelectorData;
+import org.dromara.soul.common.dto.convert.DivideUpstream;
 import org.dromara.soul.common.enums.ConfigGroupEnum;
 import org.dromara.soul.common.enums.DataEventTypeEnum;
 import org.dromara.soul.common.enums.PluginEnum;
+import org.dromara.soul.common.utils.GsonUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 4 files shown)


soul-admin/src/main/java/org/dromara/soul/admin/service/impl/UpstreamCheckService.java
```diff
@@ -168,21 +182,25 @@ private void check(final String selectorName, final List<DivideUpstream> upstrea
...
-            if (upstreams == null) {
-                selector.setHandle("");
-                selectorData.setHandle("");
-            } else {
-                String handler = GsonUtils.getInstance().toJson(upstreams);
-                selector.setHandle(handler);
+        SelectorDO selectorDO = selectorMapper.selectByName(selectorName);
+        if (Objects.nonNull(selectorDO)) {
+            String handler = CollectionUtils.isEmpty(upstreams) ? "" : GsonUtils.getInstance().toJson(upstreams);
+            List<ConditionData> conditionDataList = selectorConditionMapper
```
yu199195, line 188: what is mean？

soul-admin/src/main/java/org/dromara/soul/admin/service/impl/SelectorServiceImpl.java
```diff
@@ -264,4 +271,15 @@ private SelectorData buildSelectorData(final SelectorDO selectorDO) {
         return SelectorDO.transFrom(selectorDO, pluginDO.getName(), conditionDataList);
     }
 
+    private void updateDivideUpstream(final SelectorDO selectorDO) {
```
yu199195, line 274: what is mean？

soul-admin/src/main/java/org/dromara/soul/admin/service/impl/UpstreamCheckService.java
augfool: 1. Remove the dependency of the Class UpstreamCheckService on the Class SelectorService and modify it to rely on the Class Mapper, because the Class SelectorService will depend on the Class UpstreamCheckService.
2. The value of the selector is already known when the selectorData is generated, reducing one query operation.

soul-admin/src/main/java/org/dromara/soul/admin/service/impl/SelectorServiceImpl.java
augfool: The divide selector added through the web console is updated to UpstreamCheckService#UPSTREAM_MAP.

PR #1022, opened 2021-01-19
Add final for `execute` method.

yiwenlong:
`execute` method is designed as a template method, so the final keyword should be used to prevent subclasses from overriding the implementation of the `execute` method.


tydhot (commented):
why

yiwenlong:
> why

Prevent subclass override.

PR #1048, opened 2021-01-24
[soul-plugin] refractor: make soul-plugin-base more clean

cottom:
Now `soul-plgin-base` have some children plugin's specific process logic. This PR is trying to move such logic into children plugin to make the base plugin more pure.

Make sure that:


soul-plugin/soul-plugin-base/src/main/java/org/dromara/soul/plugin/base/AbstractSoulPlugin.java
```diff
@@ -99,6 +97,14 @@
         return chain.execute(exchange);
     }
 
+    protected Mono<Void> handleSelectorIsNull(final String pluginName, final ServerWebExchange exchange, final SoulPluginChain chain) {
```
yu199195, line 100: not used in the AbstractSoulPlugin?

cottom: old function signature. used as before.

image

PR #1058, opened 2021-01-25
delete redundant code, refactor onRuleChanged method

### LINKED ISSUE #1049: about nacos-sync-data, maybe some code is redundant.
## Question

// your question here;
when I use nacos-sync-data, if I change auth data at admin, then changed data will be send NaosDataChangedListener#updateAuthMap method invoke, but in this method,the configInfo param is old data, and the AUTH_MAP also old, there old data change equal no change, so, what use is this code.

Please pay attention on issues you submitted, because we maybe need more details. 
If no response anymore and we cannot reproduce it on current information, we will **close it**.
image

Zhoutzzz:
fixes #1049 

This PR does two things:

Firstly, the redundant method named 'updatexxxMap' in NacosDataChangeLlistener listener is deleted. These methods are not useful in data synchronization.

Second, in 'onRuleChanged' method, old data will be cached. In the long run, there will be more and more rule related cache data, which will affect the performance. The cache processing logic of this method has been modified. When new data comes, it will be directly added to the cache, and when the original data is modified, it will be covered.

Finally, the modified code has been verified and the function is normal.

Make sure that:


soul-admin/src/main/java/org/dromara/soul/admin/listener/nacos/NacosDataChangedListener.java
```diff
@@ -233,20 +223,35 @@ public void onRuleChanged(final List<RuleData> changed, final DataEventTypeEnum
...
+                            .parallelStream()
+                            .collect(Collectors.toMap(RuleData::getId, v -> v));
+                    currentData.put(rule.getId(), rule);
+                    RULE_MAP.put(rule.getSelectorId(),
+                            currentData
+                                    .values()
+                                    .parallelStream()
+                                    .sorted(RULE_DATA_COMPARATOR)
+                                    .collect(Collectors.toList()));
                 });
```
dengliming, line 237:
```
                    List<RuleData> ls = RULE_MAP
                            .getOrDefault(rule.getSelectorId(), new ArrayList<>())
                            .stream()
                            .filter(s -> !s.getId().equals(rule.getId()))
                            .sorted(RULE_DATA_COMPARATOR)
                            .collect(Collectors.toList());
                    ls.add(rule);
                    RULE_MAP.put(rule.getSelectorId(), ls);
```

dengliming (changes requested):

soul-admin/src/main/java/org/dromara/soul/admin/listener/nacos/NacosDataChangedListener.java
Zhoutzzz: ok,thanks you review my code #

dengliming (approved):

PR #1080, opened 2021-02-01
delete the RestController,just use Controller

stephenshen1993:
It is unnecessary to judge restcontroller in spring mvcclientbean postprocessor, because restcontroller is also a controller,
so delete the RestController,just use Controller


soul-client/soul-client-http/soul-client-springmvc/src/main/java/org/dromara/soul/client/springmvc/init/SpringMvcClientBeanPostProcessor.java
```diff
@@ -74,9 +73,8 @@ public Object postProcessAfterInitialization(@NonNull final Object bean, @NonNul
             return bean;
         }
         Controller controller = AnnotationUtils.findAnnotation(bean.getClass(), Controller.class);
-        RestController restController = AnnotationUtils.findAnnotation(bean.getClass(), RestController.class);
```
yu199195, line 77: after， also support RestController  Annotation？

stephenshen1993: Yes, it does.  It will work like this:
image-20210202135138034

dengliming:
Can you also add a test case to cover your changes? Thanks.

stephenshen1993:
> Can you also add a test case to cover your changes? Thanks.

I will try it tonight.

PR #1099, opened 2021-02-07
validate username uniqueness on create or update scenario

zendwang:

Make sure that:


soul-admin/src/main/java/org/dromara/soul/admin/controller/DashboardUserController.java
```diff
@@ -110,17 +110,22 @@ public SoulAdminResult detailDashboardUser(@PathVariable("id") final String id)
...
             item.setPassword(AesUtils.aesEncryption(item.getPassword(), key));
-            Integer createCount = dashboardUserService.createOrUpdate(item);
+            int createCount = dashboardUserService.createOrUpdate(item);
             log.info("dashboard user created, info:[{}]", item);
-            return SoulAdminResult.success(SoulResultMessage.CREATE_SUCCESS, createCount);
+            if (0 < createCount) {
+                return SoulAdminResult.success(SoulResultMessage.CREATE_SUCCESS, createCount);
+            } else {
+                return SoulAdminResult.error(SoulResultMessage.DASHBOARD_USER_NAME_EXIST_ERROR);
+            }
```
nuo-promise, line 122: I don't think I need to change this

soul-admin/src/main/java/org/dromara/soul/admin/controller/DashboardUserController.java
```diff
@@ -132,9 +137,13 @@ public SoulAdminResult updateDashboardUser(@PathVariable("id") final String id,
...
         dashboardUserDTO.setPassword(AesUtils.aesEncryption(dashboardUserDTO.getPassword(), key));
-        Integer updateCount = dashboardUserService.createOrUpdate(dashboardUserDTO);
+        int updateCount = dashboardUserService.createOrUpdate(dashboardUserDTO);
         log.info("dashboard user updated, id:[{}], info:[{}]", id, dashboardUserDTO);
-        return SoulAdminResult.success(SoulResultMessage.UPDATE_SUCCESS, updateCount);
+        if (0 < updateCount) {
+            return SoulAdminResult.success(SoulResultMessage.UPDATE_SUCCESS, updateCount);
+        } else {
+            return SoulAdminResult.error(SoulResultMessage.DASHBOARD_USER_NAME_EXIST_ERROR);
+        }
```
nuo-promise, line 146: Updating logically means modifying something that is already there

soul-admin/src/main/java/org/dromara/soul/admin/service/impl/DashboardUserServiceImpl.java
```diff
@@ -88,6 +88,10 @@ public DashboardUserServiceImpl(final DashboardUserMapper dashboardUserMapper, f
     @Override
     @Transactional(rollbackFor = Exception.class)
     public int createOrUpdate(final DashboardUserDTO dashboardUserDTO) {
+        DashboardUserDO existedDashboardUserDO = dashboardUserMapper.selectByUserName(dashboardUserDTO.getUserName());
+        if (Objects.nonNull(existedDashboardUserDO) && !existedDashboardUserDO.getId().equals(dashboardUserDTO.getId())) {
+            return -1;
+        }
```
nuo-promise, line 94: image

Wouldn't it be better to change it this way, so that the Controller doesn't have to change

nuo-promise (commented):
some advice

soul-admin/src/main/java/org/dromara/soul/admin/controller/DashboardUserController.java
```diff
@@ -110,17 +110,22 @@ public SoulAdminResult detailDashboardUser(@PathVariable("id") final String id)
      */
     @PostMapping("")
     public SoulAdminResult createDashboardUser(@RequestBody final DashboardUserDTO dashboardUserDTO) {
-        String key = secretProperties.getKey();
+        final String key = secretProperties.getKey();
         return Optional.ofNullable(dashboardUserDTO).map(item -> {
             item.setPassword(AesUtils.aesEncryption(item.getPassword(), key));
-            Integer createCount = dashboardUserService.createOrUpdate(item);
+            int createCount = dashboardUserService.createOrUpdate(item);
```
tydhot, line 116: why

zendwang: the return data type of member method  "createOrUpdate"  is primitive data type "int".so I think we shouldn't convert to wrapper data type "Integer".

soul-admin/src/main/java/org/dromara/soul/admin/service/impl/DashboardUserServiceImpl.java
```diff
@@ -90,7 +90,7 @@ public DashboardUserServiceImpl(final DashboardUserMapper dashboardUserMapper, f
     public int createOrUpdate(final DashboardUserDTO dashboardUserDTO) {
         DashboardUserDO existedDashboardUserDO = dashboardUserMapper.selectByUserName(dashboardUserDTO.getUserName());
         if (Objects.nonNull(existedDashboardUserDO) && !existedDashboardUserDO.getId().equals(dashboardUserDTO.getId())) {
-            return -1;
+            return 0;
         }
```
nuo-promise: If it's an update operation, is there a problem here？

nuo-promise (commented):
Update and insert are in the same function. If it's an update, you first check if there's any data and it's true, then you can't perform the update operation

dengliming (dismissed):
I think it would be better to add a `unique index`. We can merge it for now.

zendwang:
> I think it would be better to add a `unique index`. We can merge it for now.

agree with your idea. but we also should tell user to change input when error happened.

dengliming (approved):

PR #1114, opened 2021-02-21
optimize client-springmvc registration time

en-o:
optimize client-springmvc registration time
It could be an example！
When registering, the project may not start successfully, but it just collides with polling, resulting in invalid registration

 [Please see the detailed description of the](https://www.yuque.com/tanning/epv4c9/bwbs88)

 

yu199195:
no， we also support springmvc， not only springboot

Issue #1188, opened 2021-03-24
nacos2.0.0 adapt


naah69:
adapt 2.0.0 version of nacos

PR #1189, opened 2021-03-24
nacos2.0.0 adapt

naah69:
issue #1188 

Make sure that:

### DIFF
--- a/soul-admin/pom.xml
+++ b/soul-admin/pom.xml
@@ -28,7 +28,7 @@
     <artifactId>soul-admin</artifactId>
 
     <properties>
-        <nacos-client.version>1.2.0</nacos-client.version>
+        <nacos-client.version>2.0.0</nacos-client.version>
     </properties>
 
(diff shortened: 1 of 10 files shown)


yu199195:
whats the modify?

yu199195:
@naah69 I don't understand the meaning of your modification, Can you explain it to me

soul-admin/pom.xml
```diff
@@ -28,7 +28,7 @@
     <artifactId>soul-admin</artifactId>
 
     <properties>
-        <nacos-client.version>1.2.0</nacos-client.version>
+        <nacos-client.version>2.0.0</nacos-client.version>
```
yu199195, line 31: maybe move to root pom.xml

naah69: it only change version of  nacos-client to provide performance of nacos.

commit c5bbe3c fix error when nacos username or password is null in config file.

PR #1194, opened 2021-03-25
fix sql can't execute when mysql database name has been changed

naah69:

Make sure that:

### DIFF
--- a/soul-admin/src/main/resources/META-INF/schema.sql
+++ b/soul-admin/src/main/resources/META-INF/schema.sql
@@ -14,9 +14,7 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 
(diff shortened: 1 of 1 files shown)


naah69:
the zk and nacos is in the other pr.
wait it ,and merge again,then the build will pass

soul-admin/src/main/resources/META-INF/schema.sql
```diff
@@ -14,9 +14,7 @@
 -- See the License for the specific language governing permissions and
 -- limitations under the License.
 
-CREATE DATABASE  IF NOT EXISTS  `soul`  DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci ;
 
-USE `soul`;
```
dengliming, line 19: If the database itself doesn't exist, how to do?

dengliming (changes requested):

yu199195:
no， this pr we can not merged

PR #1214, opened 2021-04-01
spring cloud client support config servlet.context-path and fix "null/**" path for metadata when soul`s contextPath is null

xxd763795151:

1: When application.yml config server.servlet.context-path，the rule path does not contains the server.servlet.context-path. So I get fixed it.
2: When soul.client.props.contextPath is null, the path for metadata is "null/**". I get fixed it.

There are four scene:
if the uri for controller is "uri/**", 

case 1: server.servlet.context-path is null, soul.client.props.contextPath is "springCloud"， so the rule path: springCloud/uri/**, the path for metadata is springCloud/**;

case 2: server.servlet.context-path is "app", soul.client.props.contextPath is "springCloud"， so the rule path: springCloud/app/uri/**, the path for metadata is springCloud/**;

case 3: server.servlet.context-path is "app", soul.client.props.contextPath is null， so the rule path: app/app/uri/**, the path for metadata is app/**;

case 3: server.servlet.context-path is null, soul.client.props.contextPath is null， so the rule path: /**, the selector is rule, the path for metadata is /**;

Make sure that:

- [ ] You have read the [contribution guidelines](https://dromara.org/projects/soul/contributor/).

### DIFF
--- a/soul-client/soul-client-http/soul-client-springcloud/src/main/java/org/dromara/soul/client/springcloud/init/SpringCloudClientBeanPostProcessor.java
+++ b/soul-client/soul-client-http/soul-client-springcloud/src/main/java/org/dromara/soul/client/springcloud/init/SpringCloudClientBeanPostProcessor.java
@@ -63,6 +63,8 @@ public class SpringCloudClientBeanPostProcessor implements BeanPostProcessor {
     private final String host;
 
     private final String port;
+
+    private final String servletContextPath;
     
     /**
      * Instantiates a new Soul client bean post processor.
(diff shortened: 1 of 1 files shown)


yu199195:
env.getProperty("server.servlet.context-path", "")   if is not config？

xxd763795151:
> env.getProperty("server.servlet.context-path", "") if is not config？

@yu199195 if it is not config, the values is empty  string(""). then it do not affect merge path:
```String path = contextPath + this.servletContextPath + prePath + soulSpringCloudClient.path();```

xxd763795151:
@yu199195 Sorry， I forgot thinking about this scene: if the server.servlet.context-path is not null but is "/"，the path will occur a question.
```contextPath + this.servletContextPath + prePath + soulSpringCloudClient.path();```
I want to fix it just now. And I found you have remove it from the path:
``` String path = contextPath + prePath + soulSpringCloudClient.path();```, and the comment is "fix error"， what`s about the error， I want fix it , because I have found some project which have config the server.servlet.context-path want to use gateway, however I can not ask the developer of these project to remove the config item about servlet contxt path. Or please you fix it later , then it can support config server.servlet.context-path. Thanks!!!

PR #1274, opened 2021-04-14
fix and finish ConcurrentRateLimiterAlgorithm

WYUSIG:
old version has **two** files of concurrent_request_rate_limiter.lua and request_rate_limiter.lua (location: soul-web and soul-plugin-ratelimiter),
 so i delete soul-web one, 

and the old concurrent_request_rate_limiter will throw Exception for that:

2021-04-14 22:47:35 [lettuce-nioEventLoop-6-1] ERROR org.dromara.soul.plugin.ratelimiter.handler.SoulReactiveScriptExecutor - Redis execute exception: Error in execution; nested exception is io.lettuce.core.RedisCommandExecutionException: ERR Error running script (call to f_7b19c6e0eeb8bbe111771ec3974993c7d0445f94): @user_script:30: @user_script: 30: **Wrong number of args calling Redis command From Lua script**
(Because of  old concurrent_request_rate_limiter.lua file has code "**redis.call("setex", key, timestamp)**", it not allow do it in the zset)

ConcurrentRateLimiterAlgorithm: **"Just add a random token to a set at the start of a request and remove it from the set when you're done."**
> https://gist.github.com/ptarjan/e38f45f2dfe601419ca3af937fff574d#file-1-check_request_rate_limiter-rb-L11-L34

this is the test I finished


WYUSIG:
org.dromara.soul.plugin.ratelimiter.executor.RedisRateLimiterScriptsTest#concurrentLuaTest：
List<String> keys = Stream.of("test-concurrent").collect(Collectors.toList());

It doesn't match the reality

WYUSIG:
There is no current limiting group based on rule ID. my other idea is to send the request ID to the concurrent ratelimiter algorithm through ThreadLocal

WYUSIG:
If the last cleaning request ID fails, it will always occupy the request capacity. Another idea I have is to clean the scope 100 seconds ago

yu199195:
hi,  why remove request_rate_limiter.lua ?

WYUSIG:
Because  soul-plugin-ratelimiter already has one

yu199195:
request_rate_limiter.lua used by  TokenBucketRateLimiterAlgorithm

WYUSIG:
soul-web resources\META-INF\scripts has one request_rate_limiter.lua, soul-plugin-ratelimiter resources\META-INF\scripts also has one request_rate_limiter，so i remove the soul-web one

soul-plugin/soul-plugin-ratelimiter/src/main/java/org/dromara/soul/plugin/ratelimiter/algorithm/ConcurrentRateLimiterAlgorithm.java
```diff
@@ -38,4 +43,11 @@ protected String getScriptName() {
     protected String getKeyName() {
         return "concurrent_request_rate_limiter";
     }
+
+    @Override
+    public List<String> getKeys(final String id) {
+        String prefix = getKeyName() + ".{";
+        String tokenKey = prefix + "}.tokens";
```
yu199195, line 50: the key can used uuId?  NOT request id

WYUSIG: Does RateLimiterPlugin generate a UUID and pass it in? Because RateLimiterPlugin needs to remove by this ID

yu199195:
@WYUSIG  hi, i think  Override getKeys ,and then used uuid , like SlidingWindowRateLimiterAlgorithm.java ,be better. thanks!

yu199195:
I see what you mean, thanks for PR, I will reconstruct it later

Issue #1266, opened 2021-04-14
Connection refused during graceful shutdown and startup


jerbo99:
(springboot http service, registered with zookeeper) The upstream service gracefully shutdown, or when it is just started, there is a "Connection refused" in the soul-bootstrap.

Suggestion: When the upstream service gracefully shutdown, first go offline from the registration center, wait 1-2 seconds before rejecting new requests, and finally stop the service until the existing requests are processed or timeout; when starting, delay register until the service is started.

yu199195:
can you fix it?

jerbo99:
I think that the code in persistInterface can be change to async and delay n seconds. But I don't know how to judge that the service has been startup completed, to judge when to execute. I haven't figured out how to change the graceful shutdown.

jerbo99:
In soul-register-common, set the started flag through ApplicationRunner, async persistInterface judges whether it can be register according to the flag.

jerbo99:
Registration has been decoupled by queue, no longer need to be async

jerbo99:
The problem of connection refused during shutdown was solved by adding a shutdown hook to close zkClient and delaying other hooks for n seconds.

PR #1276, opened 2021-04-15
Connection refused during startup

jerbo99:
Fixes part of #1266
Delay registration at startup until the service startup is complete


yu199195:
in register common module  can not use springboot,  we also must be supprot spring 3.X  or other

jerbo99:
> in register common module can not use springboot, we also must be supprot spring 3.X or other
understood.

PR #1278, opened 2021-04-15
cache json deserialization for performance of ContextPathMappingPlugin

kimmking:
cache json deserialization for performance of ContextPathMappingPlugin


Asxing (approved):

yu199195:
This is a bad change, and we want to initialize it when we change it, and clear it when we delete it

PR #1324, opened 2021-04-25
Update MatchOperatorJudge judge method to matches

qurenneng:
Update MatchOperatorJudge judge method to matches


soul-plugin/soul-plugin-base/src/main/java/org/dromara/soul/plugin/base/condition/judge/MatchOperatorJudge.java
```diff
@@ -35,6 +35,6 @@ public Boolean judge(final ConditionData conditionData, final String realData) {
         if (Objects.equals(ParamTypeEnum.URI.getName(), conditionData.getParamType())) {
             return PathMatchUtils.match(conditionData.getParamValue().trim(), realData);
         }
-        return realData.contains(conditionData.getParamValue().trim());
+        return realData.matches(conditionData.getParamValue().trim());
```
dengliming, line 38: You can use `RegExOperatorJudge`.

dengliming (changes requested):

dengliming:
IMO, We already have `RegExOperatorJudge`. This change is unnecessary.

Issue #1321, opened 2021-04-25
dubbo plugin "Generic invocation" will throw exception when dubbo version below 2.7.3


ttttangzhen:
## Question
dubbo "Generic invocation" support "async invoke" on version 2.7.3. if we use dubbo version 2.7.2, it will throws exception.

the dubbo plugin "generic invocation" method may change to sync type?

yu199195:
the apache dubbo ?   add dubbo version ,

ttttangzhen:
> the apache dubbo ? add dubbo version ,

yes, the apache dubbo，that means telling users to use apache dubbo higher than version 2.7.3?

ttttangzhen:
`CompletableFuture<Object> future = genericService.$invokeAsync(metaData.getMethodName(), pair.getLeft(), pair.getRight());`

apache dubbo only supports this method after version 2.7.3,
so that, users who use soul apache dubbo plugin need use apache dubbo higher than version 2.7.3

KevinClair:
You can wrap the `$invoke` method with `CompletableFuture`.

ttttangzhen:
> You can wrap the `$invoke` method with `CompletableFuture`.

I don't understand😂

ttttangzhen:
@KevinClair I can fix this problem , I think it don't need 'async call' , I wan't to change it to 'sync call ' , It is OK ?

KevinClair:
> @KevinClair I can fix this problem , I think it don't need 'async call' , I wan't to change it to 'sync call ' , It is OK ?

Yes，you can send a pull request. However, I mean if it's not work in the version of `apache dubbo`  less than 2.7.3, you can wrap the `$invoke` method with `CompletableFuture`.
```java
public interface GenericService {

    /**
     * Generic invocation
     *
     * @param method         Method name, e.g. findPerson. If there are overridden methods, parameter info is
     *                       required, e.g. findPerson(java.lang.String)
     * @param parameterTypes Parameter types
     * @param args           Arguments
     * @return invocation return value
     * @throws GenericException potential exception thrown from the invocation
     */
    Object $invoke(String method, String[] parameterTypes, Object[] args) throws GenericException;

    default CompletableFuture<Object> $invokeAsync(String method, String[] parameterTypes, Object[] args) throws GenericException {
        Object object = $invoke(method, parameterTypes, args);
        if (object instanceof CompletableFuture) {
            return (CompletableFuture<Object>) object;
        }
        return CompletableFuture.completedFuture(object);
[...]

PR #1335, opened 2021-04-26
fix: change apache dubbo $invokeAsync to $invoke(#1321)

ttttangzhen:

change apache dubbo $invokeAsync to $invoke(#1321) and update test case

Make sure that:


yu199195:
$invokeAsync   will be remove？

ttttangzhen:
> $invokeAsync will be remove？

yes,I think there don't need async, because the original way is to block the value immediately after calling the asynchronous, so I remove async method.

PR #1376, opened 2021-05-02
fix RateLimiterAlgorithm#callback execution sequence

### LINKED ISSUE #1375: RedisRateLimiter#isAllowed “doFinally” execution sequence problem
After my test, I found that，RedisRateLimiter#isAllowed “doFinally” execution sequence is :

1.WebClientPlugin#execute send request

2.doFinally execution RateLimiterAlgorithm#callback

3.WebClientResponsePlugin#execute write response

I think a request needs to write response ，then execution RateLimiterAlgorithm#callback, because

Concurrent requests limiter：

“Because Redis is so fast, doing the naive thing works. Just add a random token to a set at the start of a request and remove it from the set when you're done. If the set is too large, reject the request.”

from https://gist.github.com/ptarjan/e38f45f2dfe601419ca3af937fff574d#file-1-check_request_rate_limiter-rb-L11-L34

-------------------------------------------------------
so the doFinally It should be written in RateLimiterPlugin#doExecute，If so, then execution sequence is :

1.WebClientPlugin#execute send request

2.WebClientResponsePlugin#execute write response

3.doFinally execution RateLimiterAlgorithm#callback

WYUSIG:
fixes #1375 move the RateLimiterAlgorithm callback action to RateLimiterPlugin#doExecute doFinally.


WYUSIG:
Is the code not good enough? Can you tell me, I can do better.

dengliming (dismissed):

dengliming:
@WYUSIG Thanks for contribution. I took a quick glance at the changes and looks good to me. Let's wait for more reviews from other committers. : )

tydhot (approved):

soul-plugin/soul-plugin-ratelimiter/src/main/java/org/dromara/soul/plugin/ratelimiter/executor/RedisRateLimiter.java
```diff
@@ -64,11 +72,15 @@
...
+                        exchange.getAttributes().put(RATE_LIMITER_ALGORITHM, rateLimiterAlgorithm);
+                        exchange.getAttributes().put(KEYS, keys);
+                        exchange.getAttributes().put(SCRIPT_ARGS, scriptArgs);
+                    }
                     Long tokensLeft = results.get(1);
                     return new RateLimiterResponse(allowed, tokensLeft);
                 })
-                .doOnError(throwable -> log.error("Error determining if user allowed from redis:{}", throwable.getMessage()))
-                .doFinally(signalType -> rateLimiterAlgorithm.callback(script, keys, scriptArgs));
+                .doOnError(throwable -> log.error("Error determining if user allowed from redis:{}", throwable.getMessage()));
```
tydhot, line 83: It has nothing to do with your changes here, but I think the expression of this diary is 农not very clear.

WYUSIG: After my test, I found that，RedisRateLimiter#isAllowed “doFinally” execution sequence is :

1.WebClientPlugin#execute send request

2.doFinally execution RateLimiterAlgorithm#callback

3.WebClientResponsePlugin#execute write response

I think a request needs to write response ，then execution RateLimiterAlgorithm#callback, because

Concurrent requests limiter：

“Because Redis is so fast, doing the naive thing works. Just add a random token to a set at the start of a request and remove it from the set when you're done. If the set is too large, reject the request.”

from https://gist.github.com/ptarjan/e38f45f2dfe601419ca3af937fff574d#file-1-check_request_rate_limiter-rb-L11-L34

so the doFinally It should be written in RateLimiterPlugin#doExecute，If so, then execution sequence is :

1.WebClientPlugin#execute send request

2.WebClientResponsePlugin#execute write response

3.doFinally execution RateLimiterAlgorithm#callback

soul-plugin/soul-plugin-ratelimiter/src/main/java/org/dromara/soul/plugin/ratelimiter/executor/RedisRateLimiter.java
```diff
@@ -41,15 +42,22 @@
 @Slf4j
 public class RedisRateLimiter {
     
+    public static final String RATE_LIMITER_ALGORITHM = "rateLimiterAlgorithm";
+    
+    public static final String KEYS = "keys";
+
+    public static final String SCRIPT_ARGS = "scriptArgs";
```
tydhot, line 49: I think the key of the attribute here maybe  need a prefix like soul or ratelimiter to make the meaning clearer.

tydhot (changes requested):

soul-plugin/soul-plugin-ratelimiter/src/main/java/org/dromara/soul/plugin/ratelimiter/RateLimiterPlugin.java
```diff
@@ -61,18 +64,26 @@ public int getOrder() {
...
                         Object error = SoulResultWrap.error(SoulResultEnum.TOO_MANY_REQUESTS.getCode(), SoulResultEnum.TOO_MANY_REQUESTS.getMsg(), null);
                         return WebFluxResultUtils.result(exchange, error);
                     }
                     return chain.execute(exchange);
+                }).doFinally(s -> {
+                    RateLimiterAlgorithm<?> rateLimiterAlgorithm = (RateLimiterAlgorithm<?>) exchange.getAttributes().get(RedisRateLimiter.RATE_LIMITER_ALGORITHM);
+                    if (rateLimiterAlgorithm != null) {
+                        List<String> keys = (List<String>) exchange.getAttributes().get(RedisRateLimiter.KEYS);
+                        List<String> scriptArgs = (List<String>) exchange.getAttributes().get(RedisRateLimiter.SCRIPT_ARGS);
+                        rateLimiterAlgorithm.callback(null, keys, scriptArgs);
```
WYUSIG, line 85: Do I need to change the parameters of the callback method？because I don't think scripts will be passed to the method

soul-plugin/soul-plugin-ratelimiter/src/main/java/org/dromara/soul/plugin/ratelimiter/executor/RedisRateLimiter.java
tydhot: I just happened to find that the grammar of this log is not very right. Can you fix it by the way?

soul-plugin/soul-plugin-ratelimiter/src/main/java/org/dromara/soul/plugin/ratelimiter/executor/RedisRateLimiter.java
```diff
@@ -44,9 +44,9 @@
     
     public static final String RATE_LIMITER_ALGORITHM = "rateLimiterAlgorithm";
     
-    public static final String KEYS = "keys";
+    public static final String RATE_LIMITER_KEYS = "keys";
```
dengliming, line 47: I think you misunderstood @tydhot 's meaning. That should be something like that:
`public static final String RATE_LIMITER_KEYS = "_rate_limiter_keys";`

soul-plugin/soul-plugin-ratelimiter/src/main/java/org/dromara/soul/plugin/ratelimiter/executor/RedisRateLimiter.java
WYUSIG: I found that spring cloud gateway's error log is the same
```java
catch (Exception e) {
	/*
         * We don't want a hard dependency on Redis to allow traffic. Make sure to set
         * an alert so you know if this is happening too much. Stripe's observed
         * failure rate is 0.01%.
         */
	log.error("Error determining if user allowed from redis", e);
}
```
https://github.com/spring-cloud/spring-cloud-gateway/blob/main/spring-cloud-gateway-server/src/main/java/org/springframework/cloud/gateway/filter/ratelimit/RedisRateLimiter.java

line 284

soul-plugin/soul-plugin-ratelimiter/src/main/java/org/dromara/soul/plugin/ratelimiter/algorithm/RateLimiterAlgorithm.java
```diff
@@ -49,10 +49,9 @@
     /**
      * Callback string.
      *
-     * @param script the script
      * @param keys the keys
      * @param scriptArgs the script args
      */
-    default void callback(final RedisScript<?> script, final List<String> keys, final List<String> scriptArgs) {
+    default void callback(final List<String> keys, final List<String> scriptArgs) {
```
WYUSIG, line 55: RedisScript can get by this.getScript(), so i remove it

soul-plugin/soul-plugin-ratelimiter/src/main/java/org/dromara/soul/plugin/ratelimiter/executor/RedisRateLimiter.java
```diff
@@ -41,15 +42,22 @@
 @Slf4j
 public class RedisRateLimiter {
     
+    public static final String RATE_LIMITER_ALGORITHM = "rateLimiterAlgorithm";
+    
+    public static final String RATE_LIMITER_KEYS = "keys";
+
+    public static final String RATE_LIMITER_SCRIPT_ARGS = "scriptArgs";
+
```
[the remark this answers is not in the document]
WYUSIG: This code has been fixed

Issue #1397, opened 2021-05-08
If the content-type is not set in the request header，WebClientPlugin will set "application/json" for default


PandaUncle:
## Describe the bug
If the content-type is not set in the request header，WebClientPlugin will set "application/json" for default. This will trigger some exceptions。
for example:
My controller function is like this:
@GetMapping(value="/test/name")
public Object getOjb() {} 
If the content-type is not set in the request header, an execption will throw:
org.springframework.web.HttpMediaTypeNotSupportedException: Content type 'application/json' not supported。
And I don't think gateway should change origin request header。

## Environment
- Soul version(s): 2.3.1-master
- My service base on spring boot 2.1.5

ttttangzhen:
I think so, usually, the browser will set content-type to "application/x-www-urlencoded".
If shenyu set the default value to "application/json", it may be inconsistent with expectations

yu199195:
can you fix it?

PandaUncle:
> can you fix it?

ok, I'll fix it this weekend

PandaUncle:
> > can you fix it?
> 
> ok, I'll fix it this weekend

fixed in pr #1428

PR #1428, opened 2021-05-09
fix issue #1397

PandaUncle:
Reason: don't modify the header's content type, keep it as it is


tydhot (approved):

tydhot:
@yu199195 Maybe you should review it too.

yu199195:
if user have contextType？  this can not be remove

PandaUncle:
> if user have contextType？ this can not be remove

it will work well, because the origin request headers are added! see here:
WebClientPlugin.java line 114
httpHeaders.addAll(exchange.getRequest().getHeaders());

PR #1585, opened 2021-06-05
[Test #1569] Add metrics reporter test for ut#1569

tangtian8:

Make sure that:


shenyu-metrics/shenyu-metrics-spi/src/test/java/org/apache/shenyu/metrics/reporter/MetricsReporterTest.java
```diff
@@ -0,0 +1,98 @@
...
+        MetricsReporter.registerHistogram("name", new String[]{"labelNames"}, "document");
+        MetricsReporter.registerHistogram("name", "document");
+    }
+
+    @Test
+    public void testCounterIncrement() {
+        MetricsReporter.counterIncrement("name", new String[]{"labelValues"});
+        MetricsReporter.counterIncrement("name");
+        MetricsReporter.counterIncrement("name", new String[]{"labelValues"}, 1);
+    }
```
dengliming, line 79: Personally, I don't think we should write tests only for improving the coverage. We need to verify that the program works as we want it to. It's not just calling methods.

tangtian8: Thank you, I'll improve

tangtian8: > IMO, I don't think we should write tests only for improving the coverage. We need to verify that the program works as we want it to. It's not just calling methods.

Thank you, I'll improve

Issue #1606, opened 2021-06-11
LocalDataSourceLoader custom read init_script path


zhaiweij:
## Task

add custom read init_script path to LocalDataSourceLoader

Task List:

- [ ] add custom read init_script path to LocalDataSourceLoader

Please read the [Contribution Guideline](https://dromara.org/projects/soul/contributor/) before submitting the PR

PR #1608, opened 2021-06-11
[ISSUE #1606] LocalDataSourceLoader custom read init_script path

zhaiweij:
LocalDataSourceLoader custom read init_script path #1606

Make sure that:

- [ ] You have read the [contribution guidelines](https://dromara.org/projects/soul/contributor/).


shenyu-admin/src/main/java/org/apache/shenyu/admin/config/properties/DataBaseProperties.java
```diff
@@ -35,5 +35,7 @@
 
     private String initScript;
 
+    private String customInitScript;
```
flycash, line 38: Actually, I think it's not a good idea that introduce new field named `customInitScript` because this field is similar to `initScript`. Users may feel confused about those two fields. 
My proposal is using some special prefix. For example, if there is no prefix, means that read file from "resources"; if the path has some special prefix, like "file:/opt/xxx" means that we read from file system. 
If we want, we could support read those files from remote server, for example supporting "http://xxx/xxx/file".

But actually, it's not a good idea either.

zhaiweij: Can also add a enableCustomInitScript

flycash: I am afraid not, it makes this config  more complicated. I just think that `init_script` is a common name, so we'd better not to introduce new field.

zhaiweij: Yes, the configuration field is still init_script, add an  enbale field, can you read the custom path

zhaiweij: Yes, the configuration field is still init_script, add an enable, can you read the custom path

flycash (approved):
Ok, I think all solutions, including my proposal, are not good enough :( since we have to introduce more complexity to keep compatibility.

So just keep this PR and don't need to do any modification.

Issue #1700, opened 2021-07-02
Prepare for the first Apache release


tuohai666:
It's time to start creating the first Apache release which is one of the most important milestones in the incubator journey.
I have checked the status of the codes, licences, binaries, etc. And found that there are some items need to be finished before releasing.
The following is a to do list, we can track the status here.

- [x] **Release Guide**
The initial version can be found here: https://shenyu.apache.org/zh/community/release/
It would be updated several times before finish. Once the guide is stable, the English version need to be added.

- [x] **Enable Nexus Access For ShenYu**
The progress can be tracked via: https://issues.apache.org/jira/browse/INFRA-22072

- [x] **Request for ShenYu DockerHub repository**
The progress can be tracked via: https://issues.apache.org/jira/browse/INFRA-22070
Available now:
https://hub.docker.com/repository/docker/apache/shenyu-bootstrap
https://hub.docker.com/repository/docker/apache/shenyu-admin

- [x] **Create the LICENSE, NOTICE, DISCLAIMER files for back end codes and binaries.**

- [x] **Create the LICENSE, NOTICE, DISCLAIMER files for front  end codes and binaries.**

- [x] **Exclude the dependency of mysql-connector-java when packaging.**

- [x] **Change the binaries of bootstrap.jar and admin.jar to .tar.gz**
In the tarball, there would be a start/stop script and a ext-lib to allow users put 3rd party libs

- [x] **How to sync/package the dashboard codes when packaging?**
Maybe submodule will resolve this problem.
Result:
Use submodule for src packaging.
Manually sync the front end codes for now.

- [x] **Refine the shenyu-dist module to assemble codes, binaries artifacts and licenses.**

- [x] **Finish the release notes**
Please update the release notes here: https://github.com/apache/incubator-shenyu/blob/master/RELEASE-NOTES.md

- [x] **Clean the source code of the shenyu-dashboard repo**
https://github.com/apache/incubator-shenyu-dashboard/pull/134
Remove doc directory if possible.
Remove mock directory if possible.
Remove or update favicon.png if possible.

- [x] **Add shenyu-dashboard as a sub project of shenyu**

- [x] **Add license headers of shenyu-dashboard**
https://github.com/apache/incubator-shenyu-dashboard/pull/135

hermeshephaestus:
@tuohai666 i want to do this! You can assign the task to me!

tuohai666:
@hermeshephaestus Which one would you like to do?

[...]

PR #1724, opened 2021-07-06
[ISSUE #1700] Exclude mysql-connector-java from shenyu-admin.jar

tuohai666:
For #1700


flycash:
I guess that `examples-http` does not have dependencies `mysql-connector`

tuohai666:
> I guess that `examples-http` does not have dependencies `mysql-connector`

Thanks for your comments. I have noticed this problem that should be resolved later.

yu199195 (approved):

PR #1802, opened 2021-07-21
Improve token security

dengliming:
- Add verifyToken
- expiredSeconds configurable
- Using different secretKey for token generator

Make sure that:

- [ ] You have read the [contribution guidelines](https://shenyu.apache.org/community/contributor/).

### DIFF
--- a/shenyu-admin/src/main/java/org/apache/shenyu/admin/config/properties/JwtProperties.java
+++ b/shenyu-admin/src/main/java/org/apache/shenyu/admin/config/properties/JwtProperties.java
@@ -29,5 +29,5 @@ import org.springframework.stereotype.Component;
 @ConfigurationProperties(prefix = "shenyu.jwt")
 public class JwtProperties {
 
-    private String key;
+    private Long expiredSeconds = 24 * 60 * 60 * 1000L;
 }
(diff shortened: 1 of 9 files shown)


shenyu-admin/src/main/java/org/apache/shenyu/admin/config/properties/JwtProperties.java
```diff
@@ -29,5 +29,5 @@
 @ConfigurationProperties(prefix = "shenyu.jwt")
 public class JwtProperties {
 
-    private String key;
+    private Long expiredSeconds = 24 * 60 * 60 * 1000L;
```
tydhot, line 32: make it configurable maybe is better

SaberSola (approved):

shenyu-admin/src/main/java/org/apache/shenyu/admin/config/properties/JwtProperties.java
dengliming: It already is. see https://github.com/dengliming/soul/blob/3df1fe73875ef500c0088823649ca76909fad7c0/shenyu-admin/src/main/resources/application.yml#L91

PR #1832, opened 2021-07-30
Using lambda and CompleteFuture to decrease logic codes (#)

Technoboy-:

### DIFF
--- a/shenyu-common/src/main/java/org/apache/shenyu/common/healthcheck/HealthCheckTask.java
+++ b/shenyu-common/src/main/java/org/apache/shenyu/common/healthcheck/HealthCheckTask.java
@@ -180,8 +180,8 @@ public final class HealthCheckTask implements Runnable {
         return checkStarted.compareAndSet(false, true);
     }
 
-    private void waitFinish()  {
-        CompletableFuture.allOf(futures.toArray(new CompletableFuture[]{})).thenAccept(__-> futures.clear());
+    private void waitFinish() {
+        CompletableFuture.allOf(futures.toArray(new CompletableFuture[]{})).thenAccept(__ -> futures.clear());
     }
 
     private void putEntityToMap(final UpstreamWithSelectorId entity) {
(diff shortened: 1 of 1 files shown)


shenyu-common/src/main/java/org/apache/shenyu/common/healthcheck/HealthCheckTask.java
```diff
@@ -139,15 +138,8 @@ private void doHealthCheck() {
...
-            String key = entry.getKey();
-            List<DivideUpstream> value = entry.getValue();
-            for (DivideUpstream upstream : value) {
-                CompletableFuture<UpstreamWithSelectorId> future = CompletableFuture.supplyAsync(() -> check(key, upstream), executor);
-
-                futures.add(future);
-            }
-        }
+        map.forEach((k, v) -> v.forEach(i -> futures.add(
+                CompletableFuture.supplyAsync(() -> check(k, i), executor).thenAccept(d -> putEntityToMap(d)))));
```
dengliming, line 142: `.thenAccept(d -> putEntityToMap(d)` That's not the original logic.

Technoboy-: The original logic is done in waitFinish.   That's not the correct way to use CompletableFuture.

Technoboy-: The original way to do putEntityToMap is in waitFinish, it's after `check` right ?

dengliming: Can you help to remove `waitFinish` method and use the correct way?

Technoboy-: Why remove `waitFinish` ? it's waiting for the futures to done.

dengliming: My mistake. You already removed ` putEntityToMap(entity); `from `waitFinish `

dengliming (approved):

PR #1875, opened 2021-08-05
modify Transfer class

HessTina-YuI:

Make sure that:

- [ ] You have read the [contribution guidelines](https://shenyu.apache.org/community/contributor/).


shenyu-admin/src/main/java/org/apache/shenyu/admin/transfer/ConditionTransfer.java
```diff
@@ -43,12 +44,11 @@
      * @return the condition data
      */
     public ConditionData mapToSelectorDO(final SelectorConditionDO selectorConditionDO) {
-        if (selectorConditionDO == null) {
+        if (Objects.isNull(selectorConditionDO)) {
```
dengliming, line 47: Maybe Optional is better?

HessTina-YuI: agree

yu199195 (approved):

PR #1883, opened 2021-08-05
feature : add integrated test for spring cloud plugin

### LINKED ISSUE #1882: feature : integrated test for spring cloud plugin
I think There is a big difference between services that use spring cloud and services that do not use spring cloud

We should add a mould to integration test on service witch use spring cloud

I will do it.

lxl910128:
Fixes  #1882
 integrated test for spring cloud plugin
 just test shennyu can route request to spring cloud service
Plugin testing will be added in the future


shenyu-integrated-test/shenyu-integrated-test-spring-cloud/admin-config/schema.sql
```diff
@@ -0,0 +1,21 @@
+-- noinspection SqlNoDataSourceInspectionForFile
```
flycash, line 1: This file please move to src/test/resources.

flycash (changes requested):
Please extract common java files to a new common module.

PR #1912, opened 2021-08-10
refactor : replace orElse with orElseGet to avoid calling too many times

Technoboy-:
##Motivation
- Replace orElse with orElseGet to avoid calling too many times if the else is not a constant variable.


SaberSola:
goog job！
can you fixe your pr name like this : refactor: replace orElse with orElseGet to avoid calling too many times

Technoboy-:
> goog job！
> can you fixe your pr name like this : refactor: replace orElse with orElseGet to avoid calling too many times

ok, thanks for reviewing. done.

dengliming (approved):

Issue #1918, opened 2021-08-11
client Will try to connect in an infinite loop, causing the interface registration to fail


zhaoguhong:
When using the client, if the host or port is configured incorrectly

[org.apache.shenyu.client.core.disruptor.subcriber.ShenyuClientMetadataExecutorSubscriber#executor](https://github.com/apache/incubator-shenyu/blob/master/shenyu-client/shenyu-client-core/src/main/java/org/apache/shenyu/client/core/disruptor/subcriber/ShenyuClientMetadataExecutorSubscriber.java)

Will try to connect in an infinite loop, causing the interface registration to fail, and no logs will be output

![image](https://user-images.githubusercontent.com/28058794/129002739-9b5cd443-616c-4c4d-b7c2-b1a417d745ca.png)

I think it can be optimized
1. The connection fails to be connected after multiple retries, and the corresponding log is output to facilitate troubleshooting
2. If multiple retries fail, the retry interval can be increased (for example, if the retry fails after three minutes, the retry interval is increased to 30s)
---

在使用 client 的时候，如果 host 或者 port 配置错误

[org.apache.shenyu.client.core.disruptor.subcriber.ShenyuClientMetadataExecutorSubscriber#executor](https://github.com/apache/incubator-shenyu/blob/master/shenyu-client/shenyu-client-core/src/main/java/org/apache/shenyu/client/core/disruptor/subcriber/ShenyuClientMetadataExecutorSubscriber.java) 

会无限循环尝试连接，导致接口注册失败，并且不会输出任何日志

我觉得可以优化下
1. 多次重试连接不上输出对应的日志便于排查问题
2. 如果多次重试失败，可以加大重试时间间隔（比如重试了三分钟仍旧失败，重试间隔加大到 30s）

yu199195:
@zhaoguhong  very good ,can you fix it ?

zhaoguhong:
@yu199195 Ok, I'll do it

midnight2104:
done

PR #1922, opened 2021-08-12
[ISSUE #1918] fix client : retry the connection add log and increase sleep time

zhaoguhong:
fix client : retry the connection add log and increase sleep time #1918


tydhot:
welcome! thanks

tydhot:
@yu199195 @zhaoguhong I think we should make it termination after a large loop. It is a waste of thread.

zhaoguhong:
> @yu199195 @zhaoguhong I think we should make it termination after a large loop. It is a waste of thread.

I agree. It can be terminated after three minutes

SaberSola (approved):

SaberSola:
Hi,Do you have plan to add a time limit like three minutes on this PR

yu199195 (approved):

zhaoguhong:
> Hi,Do you have plan to add a time limit like three minutes on this PR

Yes,But I still have a question, if the connection fails after three minutes，Should terminate or ignore the error and continue to register?

PR #1962, opened 2021-08-21
password encryption and password change logic for the admin user

qurenneng:
password encryption and password change logic for the admin user
1.The password in the detaildashboarduser method is set to null
2.Remove the compatible old login logic code from the login method

### DIFF
--- a/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/DashboardUserServiceImpl.java
+++ b/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/DashboardUserServiceImpl.java
@@ -256,18 +256,7 @@ public class DashboardUserServiceImpl implements DashboardUserService {
     private DashboardUserVO loginByDatabase(final String userName, final String password) {
         String key = secretProperties.getKey();
         String iv = secretProperties.getIv();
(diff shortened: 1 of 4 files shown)


shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/DashboardUserServiceImpl.java
```diff
@@ -256,18 +256,7 @@ private DashboardUserVO loginByLdap(final String userName, final String password
...
-                    .id(dashboardUserVO.getId())
-                    .userName(dashboardUserVO.getUserName())
-                    .password(AesUtils.aesEncryption(dashboardUserVO.getPassword(), key, iv))
-                    .role(dashboardUserVO.getRole())
-                    .enabled(dashboardUserVO.getEnabled()).build();
-            createOrUpdate(dashboardUserDTO);
-        } else {
-            dashboardUserVO = findByQuery(userName, AesUtils.aesEncryption(password, key, iv));
-        }
+        DashboardUserVO dashboardUserVO = findByQuery(userName, AesUtils.aesEncryption(password, key, iv));
```
yu199195, line 259: why del this ？

yu199195:
pls refactor this pr title

Issue #2025, opened 2021-09-04
feature plugin: RequestId plugin, generate request id for every request


qicz:
Generate  `request id` for every request, and put to request header and response header. the downstream can fetch `request id` from `request header`, the frontend can fetch `request id` from `response header`.

KevinClair:
Move it to [RequestPlugin](https://shenyu.apache.org/zh/docs/plugin-center/http-handle/request-plugin)

qicz:
> Move it to [RequestPlugin](https://shenyu.apache.org/zh/docs/plugin-center/http-handle/request-plugin)

It's different, the `request id` plugin is the global and the highest sort than others.

PR #2027, opened 2021-09-05
[ISSUE #2025] feature plugin: add request id plugin.

qicz:
# Intro
- Generate  `request id` for every request, and put to the request header. the downstream can fetch `request id` from `request header`.
  > ### the `request id plugin` is the highest sort than others.

# Custom generator
implement the `org.apache.shenyu.plugin.api.request.id.RequestIdGenerator` and put to `spring ioc` can custom `request id generator`, default is `uuid`.

Use the `org.apache.shenyu.plugin.api.request.id.ShenyuRequestIdWrap#getRequestId` API at any time to get the `request id` anywhere in the application.

### DIFF
--- a/shenyu-common/src/main/java/org/apache/shenyu/common/config/ShenyuConfig.java
+++ b/shenyu-common/src/main/java/org/apache/shenyu/common/config/ShenyuConfig.java
@@ -17,6 +17,7 @@
 
 package org.apache.shenyu.common.config;
 
+import org.apache.shenyu.common.enums.ThreadShareDataEnum;
 import org.springframework.util.StringUtils;
 
 import java.util.ArrayList;
@@ -629,6 +630,8 @@ public class ShenyuConfig {
                     add("token");
                     add("username");
                     add("client");
+                    // request id
+                    add(ThreadShareDataEnum.REQUEST_ID.getName());
                 }
             };
         }
--- a/shenyu-common/src/main/java/org/apache/shenyu/common/enums/PluginEnum.java
+++ b/shenyu-common/src/main/java/org/apache/shenyu/common/enums/PluginEnum.java
@@ -31,144 +31,149 @@ public enum PluginEnum {
     GLOBAL(10, 0, "global"),
 
     /**
+     * RequestId plugin enum.
+     */
+    REQUEST_ID(20, 0, "request-id"),
+
+    /**
      * Sign plugin enum.
      */
-    SIGN(20, 0, "sign"),
+    SIGN(30, 0, "sign"),
 
     /**
      * Jwt plugin enum.
      */
-    JWT(30, 0, "jwt"),
+    JWT(40, 0, "jwt"),
 
     /**
      * OAuth2 plugin enum.
      */
-    OAUTH2(40, 0, "oauth2"),
+    OAUTH2(50, 0, "oauth2"),
 
     /**
      * Waf plugin enum.
      */
-    WAF(50, 0, "waf"),
+    WAF(60, 0, "waf"),
 
     /**
      * Rate limiter plugin enum.
      */
-    RATE_LIMITER(60, 0, "rate_limiter"),
+    RATE_LIMITER(70, 0, "rate_limiter"),
 
     /**
      * Param mapping plugin enum.
      */
-    PARAM_MAPPING(70, 0, "param_mapping"),
+    PARAM_MAPPING(80, 0, "param_mapping"),
 
     /**
      * Context path plugin enum.
      */
-    CONTEXT_PATH(80, 0, "context_path"),
+    CONTEXT_PATH(90, 0, "context_path"),
 
     /**
      * Rewrite plugin enum.
      */
-    REWRITE(90, 0, "rewrite"),
+    REWRITE(100, 0, "rewrite"),
 
     /**
      * Cryptor request plugin enum.
      */
-    CRYPTOR_REQUEST(100, 0, "cryptor_request"),
+    CRYPTOR_REQUEST(110, 0, "cryptor_request"),
 
     /**
      * Redirect plugin enum.
      */
-    REDIRECT(110, 0, "redirect"),
+    REDIRECT(120, 0, "redirect"),
 
     /**
      * Request plugin enum.
      */
-    REQUEST(120, 0, "request"),
+    REQUEST(130, 0, "request"),
 
     /**
      * ModifyResponse plugin enum.
      */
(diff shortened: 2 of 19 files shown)


tydhot (commented):
I found there are some same changes in another pr. It looks a little complicated

qicz:
> I found there are some same changes in another pr. It looks a little complicated

another is the `request` plugin? the `request` plugin is the request to replace logic, change the headers or parameters.

the `request id` plugin provides an id for every request.  

When a request arrives, a `request id` will be created and put in the `ThreadLocal`, and returned to the user in the Response.

tydhot:
To be honest, I don't think we need to set up a plugin to do this function. This function can be used as a configuration item in the request plugin as a function supplement. Setting up a separate plugin for this small function is not a good start, which may lead to confusion and redundancy in subsequent plugin management.btw, I don't think the request id in single gateway layer has the value of backtracking and trace like the order id or traceId

qicz:
> To be honest, I don't think we need to set up a plugin to do this function. This function can be used as a configuration item in the request plugin as a function supplement. Setting up a separate plugin for this small function is not a good start, which may lead to confusion and redundancy in subsequent plugin management.btw, I don't think the request id in single gateway layer has the value of backtracking and trace like the order id or traceId

the `request id` is a service for every request that has any selector and rule. but the `request` is for a request that has the selector and rule. to my knows, maybe the `request id` works like `traceId`, because business users are not cared for `traceId`, except `request id`. every people know about what is a request call.

from aws:
```
public String requestId()
Returns the AWS request ID contained in this response metadata object. AWS request IDs can be used in the event a service call isn't working as expected and you need to work with AWS support to debug an issue.
Can be overriden by subclasses to provide service specific requestId

Returns:
The AWS request ID contained in this response metadata object.
```

tydhot:
right, maybe put it in global plugin as a configuration is better?  
but if you didn't pass this id to downstream app, it is meaningless. 
And if the downstream application does not continue to pass this field through though,  I am afraid this field is meaningless too. And this ID is not integrated with other APM systems, even if it is transmitted, it will not provide enough useful information.

qicz:
> right, maybe put it in global plugin as a configuration is better?
> but if you didn't pass this id to downstream app, it is meaningless.
> And if the downstream application does not continue to pass this field through though, I am afraid this field is meaningless too. And this ID is not integrated with other APM systems, even if it is transmitted, it will not provide enough useful information.

@tydhot  the `global` is turn on default. but the `request id` can turn off when the shenyu user is not dependency the starter. 
I think, the `request id` like the `traceId` work, or like `operate log` work. if like the `operate log`, the downstream application just logging the `request id` and error info.

tydhot:
> I think, the `request id` like the `traceId` work, or like `operate log` work. if like the `operate log`, the downstream application just logging the `request id` and error info.
I am afraid it can't be. If the downstream application does not continue to pass down with this ID, this ID has no meaning. It is just a id in gateway, it means nothing. If you think it is a `traceId`, you should insert it to its invoke context in `dubbo`, `sofa` plugin too. And I did't find you did it in your pr. Finally, if this is a traceid and we should contain it in response, we can directly integrate the existing APM framework. This is what we are doing.

qicz:
> > I think, the `request id` like the `traceId` work, or like `operate log` work. if like the `operate log`, the downstream application just logging the `request id` and error info.
> 
> I am afraid it can't be. If the downstream application does not continue to pass down with this ID, this ID has no meaning. It is just a id in gateway, it means nothing. If you think it is a `traceId`, you should insert it to its invoke context in `dubbo`, `sofa` plugin too. And I did't find you did it in your pr. Finally, if this is a traceid and we should contain it in response, we can directly integrate the existing APM framework. This is what we are doing.

the `request id` can logging the business error info and the `shenyu` error too. but if the APM framework integrated, the `traceId` and `request id` can relate in `shenyu`, but I think, the `traceId` does not care of the business users.

for the `dubbo` and `sofa`, if the `request id` is turn on default, can be this.

tydhot:
> > > I think, the `request id` like the `traceId` work, or like `operate log` work. if like the `operate log`, the downstream application just logging the `request id` and error info.
> > 
> > 
> > I am afraid it can't be. If the downstream application does not continue to pass down with this ID, this ID has no meaning. It is just a id in gateway, it means nothing. If you think it is a `traceId`, you should insert it to its invoke context in `dubbo`, `sofa` plugin too. And I did't find you did it in your pr. Finally, if this is a traceid and we should contain it in response, we can directly integrate the existing APM framework. This is what we are doing.
> 
> the `request id` can logging the business error info and the `shenyu` error too. but if the APM framework integrated, the `traceId` and `request id` can relate in `shenyu`, but I think, the `traceId` does not care of the business users.
> 
> for the `dubbo` and `sofa`, if the `request id` is turn on default, can be this.

business user will not care about a random id just live in gateway either.

tydhot:
I am not opposed to adding the function of requestId, but I am not in favor of using it as a plugin, because the capability provided by this function is limited

qicz:
> > > > I think, the `request id` like the `traceId` work, or like `operate log` work. if like the `operate log`, the downstream application just logging the `request id` and error info.
> > > 
> > > 
> > > I am afraid it can't be. If the downstream application does not continue to pass down with this ID, this ID has no meaning. It is just a id in gateway, it means nothing. If you think it is a `traceId`, you should insert it to its invoke context in `dubbo`, `sofa` plugin too. And I did't find you did it in your pr. Finally, if this is a traceid and we should contain it in response, we can directly integrate the existing APM framework. This is what we are doing.
> > 
> > 
> > the `request id` can logging the business error info and the `shenyu` error too. but if the APM framework integrated, the `traceId` and `request id` can relate in `shenyu`, but I think, the `traceId` does not care of the business users.
> > for the `dubbo` and `sofa`, if the `request id` is turn on default, can be this.
> 
> business user will not care about a random id just live in gateway either.

if the downstream application crashed that can not be logging error info.

qicz:
> I am not opposed to adding the function of requestId, but I am not in favor of using it as a plugin, because the capability provided by this function is limited

@tydhot  fine, maybe this pr just reset plugin's order and provide the `ThreadShareUtils`.

the `ThreadShareUtils` can share data for the same thread with a different name and two different threads with the same name.

tydhot:
@qicz I think u can move the requestid to global plugin as a configuration item, make it can be printed in log.

qicz:
> @qicz sorry ,Sometimes the spring cloud integration test will fail. could you commit again ? I am fixing this problem

maybe integration test is fixed by maodaren.

PR #2040, opened 2021-09-08
Optimize "GlobalErrorHandler" and refactor "SpringUtils".

qicz:
Optimize "GlobalErrorHandler" for the correct response status from `ResponseStatusException` and refactor "SpringUtils"， renname `cfgContext` to `applicationContext` and modify the type to `ApplicationContext`.


tydhot (changes requested):
I think we need more reliable libraries. You can implement the parts you need to reference in Shenyu

qicz:
> I think we need more reliable libraries. You can implement the parts you need to reference in Shenyu

I will fix this.

yu199195 (approved):

tydhot (approved):

PR #2049, opened 2021-09-10
[type refactor] add nacos sync starter pom placeholder.

qicz:
append the `shenyu-spring-boot-starter-sync-data-nacos` placeholder to `bootstrap` pom.

### DIFF
--- a/shenyu-bootstrap/pom.xml
+++ b/shenyu-bootstrap/pom.xml
@@ -316,6 +316,13 @@
             <version>${project.version}</version>
         </dependency>
 
+        <!--shenyu data sync start use nacos-->
+        <dependency>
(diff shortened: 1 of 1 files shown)


yu199195:
we cannot  add nacos

qicz:
> we cannot add nacos

I see, just placeholder like `zookeeper sync starter`

yu199195:
sync data to nacos is a way, but is not provided by default~   shenyu-boostrap its release package in apache,  so can not add nacos

qicz:
> sync data to nacos is a way, but is not provided by default~ shenyu-boostrap its release package in apache, so can not add nacos

ok

Issue #2128, opened 2021-09-25
[BUG] org.apache.shenyu.loadbalancer.cache.UpstreamCheckTaskTest.testRun test fail


dengliming:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

org.apache.shenyu.loadbalancer.cache.UpstreamCheckTaskTest.testRun test fail sometimes.

see https://github.com/apache/incubator-shenyu/runs/3706238594

### Expected Behavior

No Exception

### Environment

```markdown
ShenYu version(s):
```

prospect167:
pls assign to me

PR #2146, opened 2021-09-26
[ISSUE #2128] Fix unit tests occasionally fail to execute.

kevinBobo:
…un test fail"(#2128)

Make sure that:

### DIFF
--- a/shenyu-loadbalancer/src/test/java/org/apache/shenyu/loadbalancer/cache/UpstreamCheckTaskTest.java
+++ b/shenyu-loadbalancer/src/test/java/org/apache/shenyu/loadbalancer/cache/UpstreamCheckTaskTest.java
@@ -75,7 +75,7 @@ public class UpstreamCheckTaskTest {
         healthCheckTask.schedule();
         // Wait for the upstream-health-check thread to start.
         Awaitility.await().pollDelay(3, TimeUnit.SECONDS).untilAsserted(() -> assertFalse(healthCheckTask.getCheckStarted().get()));
-        assertTrue(healthCheckTask.getHealthyUpstream().containsKey(selectorId1));
+        assertTrue(healthCheckTask.getUnhealthyUpstream().get(selectorId1).size() > 0);
         // Let it coverage line 151~163
         when(upstream.isHealthy()).thenReturn(false).thenReturn(true);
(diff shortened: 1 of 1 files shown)


shenyu-loadbalancer/src/test/java/org/apache/shenyu/loadbalancer/cache/UpstreamCheckTaskTest.java
```diff
@@ -75,15 +75,15 @@ public void testRun() {
         healthCheckTask.schedule();
         // Wait for the upstream-health-check thread to start.
         Awaitility.await().pollDelay(3, TimeUnit.SECONDS).untilAsserted(() -> assertFalse(healthCheckTask.getCheckStarted().get()));
-        assertTrue(healthCheckTask.getHealthyUpstream().containsKey(selectorId1));
+        assertTrue(healthCheckTask.getUnhealthyUpstream().get(selectorId1).size() > 0);
         // Let it coverage line 151~163
         when(upstream.isHealthy()).thenReturn(false).thenReturn(true);
```
dengliming, line 80: `when(upstream.isHealthy()).thenReturn(false).thenReturn(true);` what does this mean?

dengliming (changes requested):

shenyu-loadbalancer/src/test/java/org/apache/shenyu/loadbalancer/cache/UpstreamCheckTaskTest.java
yu199195: maybe is right

kevinBobo: I think the original author wants to change to return true when returning false。

yu199195 (approved):

dengliming (approved):

shenyu-loadbalancer/src/test/java/org/apache/shenyu/loadbalancer/cache/UpstreamCheckTaskTest.java
dengliming: > I think the original author wants to change to return true when returning false。

We can just keep the last return.

Issue #2285, opened 2021-10-30
[BUG] `encryptKey` missed in rule handler of CryptorResponse plugin


JooKS-me:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

`encryptKey` missed in rule handler of CryptorResponse plugin

![image](https://user-images.githubusercontent.com/62384022/139518342-c88ddd4d-bef4-4ec2-a1e4-edcf34ec4688.png)

### Expected Behavior

Have `encryptKey`

### Environment

```markdown
ShenYu version(s): latest in master
```

### Anything else?

The `id` of `plugin_handle` table is conflicting.

PR #2286, opened 2021-10-30
[type: refactor] Refactor sql init script

JooKS-me:
Try to avoid problem like #2285...

Re-define the id of `plugin_handle` table as `<plugin_id + handle_id>` : 
1. The `plugin_id` is the id defined in `plugin ` table.
2. The `handle_id` is a growing number from 0 to 99 .


qicz:
@JooKS-me the `upgrade` SQL works for MySQL and pg at the same time. but the "`" can not work in pg must remove it.

JooKS-me:
> @JooKS-me the `upgrade` SQL works for MySQL and pg at the same time. but the "`" can not work in pg must remove it.

Hi, thanks for reminding. I'll try to make it compatible for PostgreSQL later.

qicz:
> Hi, thanks for reminding. I'll try to make it compatible for PostgreSQL later.

OK，the `upgrade` SQL was provided by me. In the previous test, I found that ”`“ syntax was not compatible with MySQL and PostgreSQL at the same time. I just added an upgrade guide. #2287

PR #2379, opened 2021-11-17
feature(dist): make shenyu-bootstrap build 2 images

lxl910128:
Make shenyu-bootstrap build 2 images when user docker profile.
It's same to shenyu-admin


midnight2104:
pls check ci

yu199195 (approved):

PR #2459, opened 2021-11-25
[type: feature]Add RpcContextPlugin to transmit header to rpc context.

KevinClair:
Make sure that:


yu199195 (changes requested):
wait me review

shenyu-plugin/shenyu-plugin-rpc-context/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>shenyu-plugin</artifactId>
+        <groupId>org.apache.shenyu</groupId>
+        <version>2.4.2-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>shenyu-plugin-rpc-context</artifactId>
```
tydhot, line 29: The name rpccontext is the same as Dubbo's rpccontext, which is easy to be misunderstood

KevinClair: Yes, it was originally to pass Dubbo's rpccontext, so I named the plug with rpccontext. Or you can suggest a better name to explain the meaning of the plug-in.

tydhot: Why not make a general context? The client plugin obtains and injects its own implementation from this context, and the server plugin obtains and writes the implementation of this context from its own request.After all, besides Dubbo, grpc also has its own context to pass

KevinClair: Sounds great, how about rename to `shenyu-plugin-general-context`?Or maybe just `shenyu-plugin-context`?

KevinClair:
Done, dashboard see [this](https://github.com/apache/incubator-shenyu-dashboard/pull/165)

yu199195 (approved):

PR #2460, opened 2021-11-25
[type: refactor] fix boostrap config bug

yu199195:

Make sure that:

### DIFF
--- a/shenyu-bootstrap/src/main/java/org/apache/shenyu/bootstrap/configuration/NettyTcpConfig.java
+++ b/shenyu-bootstrap/src/main/java/org/apache/shenyu/bootstrap/configuration/NettyTcpConfig.java
@@ -38,9 +38,9 @@ public class NettyTcpConfig {
 
     private boolean tcpNodelay = true;
 
-    private boolean soKeepalive = false;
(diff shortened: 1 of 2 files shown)


AhahaGe (commented):
I appreciate you can point out the bug you fixed, so that I can check in my local.
Please add comment on variable, and it's better to show why the default value should be, and when need to optimize or optimize suggestion.
How about use function coding pattern to replace private static class. For example:
`webServerFactory.addServerCustomizers(
            builder -> builder.tcpConfiguration(
                tcpServer -> tcpServer.runOn(LoopResources.create("shenyu-netty",
                        nettyTcpConfig.getSelectCount(), nettyTcpConfig.getWorkerCount(), true))
                    .selectorOption(ChannelOption.SO_BACKLOG, nettyTcpConfig.getSoBacklog())
                    .selectorOption(ChannelOption.SO_REUSEADDR, nettyTcpConfig.isSoReuseaddr())
                    .option(ChannelOption.CONNECT_TIMEOUT_MILLIS, nettyTcpConfig.getConnectTimeoutMillis())`

SaberSola (approved):

Issue #2603, opened 2021-12-16
[ContainsPredicateJudge] <contains not work for multiple config>


MiniSC:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

when i use contains for rule , It can only be configured with one
for example 
ip  matchs 127.0.0.1,127.0.0.2
it not work
![image](https://user-images.githubusercontent.com/32318597/146377959-af9af3b0-1310-452f-ba25-f8cdcc742b43.png)
there should be conditionData.getParamValue().trim()  to contains realData

### Expected Behavior

for example 
ip  matchs 127.0.0.1,127.0.0.2
can match 127.0.0.1 or 127.0.0.2

### Steps To Reproduce

not work

### Environment

```markdown
ShenYu version(s):v2.4.1
```

### Debug logs

![image](https://user-images.githubusercontent.com/32318597/146378336-cec3baf9-0bc1-40c7-bb66-f621b6139268.png)
![image](https://user-images.githubusercontent.com/32318597/146378345-d5c1a2b2-275f-4733-8ee3-55a3dbc0d64c.png)

PR #2608, opened 2021-12-17
[ISSUE #2603]  ContainsPredicateJudge not work for multiple config

MiniSC:

Make sure that:

### DIFF
--- a/shenyu-plugin/shenyu-plugin-base/src/main/java/org/apache/shenyu/plugin/base/condition/judge/ContainsPredicateJudge.java
+++ b/shenyu-plugin/shenyu-plugin-base/src/main/java/org/apache/shenyu/plugin/base/condition/judge/ContainsPredicateJudge.java
@@ -28,6 +28,6 @@ public class ContainsPredicateJudge implements PredicateJudge {
 
     @Override
     public Boolean judge(final ConditionData conditionData, final String realData) {
-        return realData.contains(conditionData.getParamValue().trim());
+        return conditionData.getParamValue().trim().contains(realData);
     }
 }
--- a/shenyu-plugin/shenyu-plugin-base/src/test/java/org/apache/shenyu/plugin/base/condition/judge/ContainsPredicateJudgeTest.java
+++ b/shenyu-plugin/shenyu-plugin-base/src/test/java/org/apache/shenyu/plugin/base/condition/judge/ContainsPredicateJudgeTest.java
@@ -1,0 +1,46 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License,  Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,  software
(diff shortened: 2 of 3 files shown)


shenyu-plugin/shenyu-plugin-base/src/main/java/org/apache/shenyu/plugin/base/condition/data/URIParameterData.java
```diff
@@ -17,17 +17,31 @@
...
-    
+
     @Override
     public String builder(final String paramName, final ServerWebExchange exchange) {
         return exchange.getRequest().getURI().getPath();
     }
+
+    @Override
+    public Boolean containsJudge(final ConditionData conditionData, final String realData) {
+        String regex = ",";
```
yu199195, line 42: URI have contains？ why not used match？

MiniSC: Mainly to fix the IP contains problem, this upgrading for uri contains just fit the old test cast in PredicateJudgeFactoryTest and let uri can be contains, but i think this separator [,] will not be used

MiniSC:
I remove my upgrading for uri , restore the old implementation for pass the PredicateJudgeFactoryTest

yu199195:
@MiniSC  hi， uri used match method, whether it overrides the contains？

MiniSC:
> @MiniSC hi， uri used match method, whether it overrides the contains？

because old contains let all param type use realData to contains condition's Data ,  old PredicateJudgeFactoryTest want let condition's uri like /http/** can be matched with /http/**/xxx in realData, but my upgrading will use condition's Data to contains realData, can not pass the old Test,  and i think in the problem of uri contains ,you are right , so i restore the old implements with uri, and it's reason of ovverride the contains

yu199195:
@MiniSC  hi, i mean uri used match, Is it no longer necessary to use the contains
>  because old contains method impelement let all param type use realData to contains condition's Data , old PredicateJudgeFactoryTest want let condition's uri like /http/** can be matched with /http/**/xxx in realData, but my upgrading will use condition's Data to contains realData, can not pass the old Test, and i think in the problem of uri contains ,you are right ,we will not use contains in uri , so i restore the old implements with uri, and it's reason of ovverride the contai

MiniSC:
> @MiniSC hi, i mean uri used match, Is it no longer necessary to use the contains
> 
> > because old contains method impelement let all param type use realData to contains condition's Data , old PredicateJudgeFactoryTest want let condition's uri like /http/** can be matched with /http/**/xxx in realData, but my upgrading will use condition's Data to contains realData, can not pass the old Test, and i think in the problem of uri contains ,you are right ,we will not use contains in uri , so i restore the old implements with uri, and it's reason of ovverride the contai

so you means i should remove the override and the test case?

shenyu-plugin/shenyu-plugin-base/src/main/java/org/apache/shenyu/plugin/base/condition/data/ParameterData.java
```diff
@@ -25,15 +26,28 @@
...
+
+    /**
+     * Judge type for contains,
+     * different Param Type have own comparison method.
+     *
+     * @param conditionData the rule conditionData
+     * @param realData      the request's realData
+     * @return {@link Boolean}
+     */
+    default Boolean containsJudge(final ConditionData conditionData, final String realData) {
```
yu199195, line 50: remove this

yu199195 (approved):

yu199195:
@MiniSC  you can add my wechat xixy199195， i will join you to contributor group

PR #2734, opened 2022-01-08
[type:test] add pathUtils test

plutokaito:
add unit test for pathUtils

Make sure that:

### DIFF
--- a/shenyu-admin/src/test/java/org/apache/shenyu/admin/utils/PathUtilsTest.java
+++ b/shenyu-admin/src/test/java/org/apache/shenyu/admin/utils/PathUtilsTest.java
@@ -17,6 +17,7 @@
 
 package org.apache.shenyu.admin.utils;
 
+import org.apache.shenyu.common.constant.AdminConstants;
 import org.junit.Test;
 
 import static org.hamcrest.Matchers.is;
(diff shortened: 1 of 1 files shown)


shenyu-admin/src/test/java/org/apache/shenyu/admin/utils/PathUtilsTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+public class PathUtilsTest {
+    private static final String URI_WRAPPER = "springCloud/test/**";
+
+    private static final String URI = "springCloud/test";
+
+    @Test
+    public void testDecoratorPath() {
+        String uri = PathUtils.decoratorPath(URI);
+        assertThat(uri, is(URI_WRAPPER));
```
tydhot, line 33: The original method contains the static variable adminconstants URI_ I think the use of suffix should also be reflected in the test

plutokaito: like `assertThat(uri, is(URI + AdminConstants.URI_SUFFIX));`?

tydhot: i think it would be better

tydhot (approved):

PR #2738, opened 2022-01-09
[type: refactor] optimize & fix uri with invalid char.

qicz:
fix `status=[abc%2Cedf]`query  `Invalid character '[' for QUERY_PARAM in status=[abc%2Cedf]` bug

### DIFF
--- a/shenyu-plugin/shenyu-plugin-uri/src/main/java/org/apache/shenyu/plugin/uri/URIPlugin.java
+++ b/shenyu-plugin/shenyu-plugin-uri/src/main/java/org/apache/shenyu/plugin/uri/URIPlugin.java
@@ -43,7 +43,6 @@ public class URIPlugin implements ShenyuPlugin {
             return chain.execute(exchange);
         }
         String rewriteUri = (String) exchange.getAttributes().get(Constants.REWRITE_URI);
-        URI uri = exchange.getRequest().getURI();
         if (StringUtils.isNoneBlank(rewriteUri)) {
             path = path + rewriteUri;
         } else {
@@ -52,8 +51,10 @@ public class URIPlugin implements ShenyuPlugin {
                 path = path + realUrl;
             }
         }
-        URI realUri = URI.create(path + "?" + RequestQueryCodecUtil.getCodecQuery(exchange));
-        exchange.getAttributes().put(Constants.HTTP_URI, realUri);
+        if (StringUtils.isNoneBlank(exchange.getRequest().getURI().getQuery())) {
+            path = String.join("?", path, RequestQueryCodecUtil.getCodecQuery(exchange));
+        }
+        exchange.getAttributes().put(Constants.HTTP_URI, URI.create(path));
         return chain.execute(exchange);
     }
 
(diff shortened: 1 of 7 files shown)


shenyu-plugin/shenyu-plugin-uri/src/main/java/org/apache/shenyu/plugin/uri/URIPlugin.java
```diff
@@ -42,27 +42,18 @@
...
-            path = String.join("?", path, uri.getRawQuery());
-            realURI = UriComponentsBuilder.fromHttpUrl(path).build(true).toUri();
-        } else {
-            if (StringUtils.isNotEmpty(uri.getQuery())) {
-                path = String.join("?", path, uri.getQuery());
-            }
-            realURI = UriComponentsBuilder.fromHttpUrl(path).build(false).toUri();
-        }
-        exchange.getAttributes().put(Constants.HTTP_URI, realURI);
+        URI realUri = URI.create(path + "?" + RequestQueryCodecUtil.getCodecQuery(exchange));
```
yu199195, line 55: why modify this， only used  UriUtils.encode(item, StandardCharsets.UTF_8))？

yu199195 (changes requested):
not merge

shenyu-plugin/shenyu-plugin-uri/src/main/java/org/apache/shenyu/plugin/uri/URIPlugin.java
qicz: some query like `status=[abc%2Cedf]` will catch the `Invalid character '[' for QUERY_PARAM in status=[abc%2Cedf]`

qicz: I will remove the `encode` transfer of the original data to upstream.

qicz:
> not merge

it is a bug.

some query like status=[abc%2Cedf] will catch the Invalid character '[' for QUERY_PARAM in status=[abc%2Cedf]

yu199195:
If modified this way, all of them are used url encoder. But for some normal ones, it is not necessary.

> some query like status=[abc%2Cedf] will catch the Invalid character '[' for QUERY_PARAM in status=[abc%2Cedf]

qicz:
> If modified this way, all of them are used url encoder. But for some normal ones, it is not necessary.
> 
> > some query like status=[abc%2Cedf] will catch the Invalid character '[' for QUERY_PARAM in status=[abc%2Cedf]

A framework like `spring` will decode automatically.

qicz:
@yu199195 I have refactored this and transferred the original data to upstream, just encode ' ' (the empty string).

qicz:
@yu199195 I has been tested, support special characters, support partial encoding, transparent transmission of business data, except for Spaces, other is unencoded raw data

dengliming (commented):
We need integration testing for these changes.

qicz:
> We need integration testing for these changes.

I am trying to test but the integration uses the ok HTTP client, and it is different with WebClient.

there are many example HTTP run results
image
image
image
image
image

yu199195 (approved):

Issue #2706, opened 2022-01-05
[Task] shenyu-admin optimize


qicz:
### Task List

- [x] refactor `BeanUtil` copy properties to `Transfer`
- [ ] refactor `ShenyuResultMessage`

![image](https://user-images.githubusercontent.com/2174082/148152154-bdf2f6d5-9389-43cc-bc09-9595b899794c.png)

zhouyadong:
please assign it  to me

PR #2750, opened 2022-01-11
[ISSUE #2706] Transfer  to copy properties, and optimize message.

zhouyadong:
Transfer  to copy properties, and optimize message.


shenyu-admin/src/main/java/org/apache/shenyu/admin/transfer/DashboardUserTransfer.java
```diff
@@ -0,0 +1,94 @@
...
+     * The constant INSTANCE.
+     */
+    INSTANCE;
+
+    /**
+     * conversion data to VO
+     * @param dashboardUserDTO original data
+     * @return {@linkplain DashboardUserVO}
+     */
+    public DashboardUserVO transferDTO2VO(final DashboardUserDTO dashboardUserDTO) {
```
AhahaGe, line 42: how about use transferDTOToVO and the rest method

qicz (approved):

PR #2784, opened 2022-01-17
[type:feature] Make Maven CI Friendly Versions

jiajiangnan:

Currently version segment is defined in every pom.xml.
Every version segment should be synchronized if there is any change.

This PR has resolved it, by defined a properties '' in the pom.xml of parent module 'shenyu'. like following image:

image

Make sure that:


jiajiangnan:
find has already some other way to update version plugin ‘versions-maven-plugin’，so close this PR

Issue #2810, opened 2022-01-26
[BUG] NPE when login with a unexist user


dragon-zhang:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

got a NPE when I login with a unexist user

### Expected Behavior

no NPE

### Steps To Reproduce

1. run `org.apache.shenyu.admin.ShenyuAdminBootstrap`
2. goto (http://0.0.0.0:9095/#/user/login)
3. login with a unexist user

### Environment

```markdown
ShenYu version(s): 2.4.2
```

### Debug logs

java.lang.NullPointerException: null
	at java.util.Optional.orElseGet(Optional.java:267)
	at org.apache.shenyu.admin.transfer.DashboardUserTransfer.transferVO2LoginVO(DashboardUserTransfer.java:72)
	at org.apache.shenyu.admin.model.vo.LoginDashboardUserVO.buildLoginDashboardUserVO(LoginDashboardUserVO.java:64)
	at org.apache.shenyu.admin.service.impl.DashboardUserServiceImpl.login(DashboardUserServiceImpl.java:242)
	at org.apache.shenyu.admin.service.impl.DashboardUserServiceImpl$$FastClassBySpringCGLIB$$e3d6f4e.invoke(<generated>)

### Anything else?

nop

PR #2811, opened 2022-01-26
[ISSUE #2810] fix NPE

dragon-zhang:
fix https://github.com/apache/incubator-shenyu/issues/2810

### DIFF
--- a/shenyu-admin/src/main/java/org/apache/shenyu/admin/transfer/DashboardUserTransfer.java
+++ b/shenyu-admin/src/main/java/org/apache/shenyu/admin/transfer/DashboardUserTransfer.java
@@ -49,7 +49,7 @@ public enum DashboardUserTransfer {
                     dashboardVO.setEnabled(data.getEnabled());
                     return dashboardVO;
                 })
-                .orElseGet(null);
+                .orElse(null);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


shenyu-admin/src/main/java/org/apache/shenyu/admin/transfer/DashboardUserTransfer.java
```diff
@@ -88,7 +88,7 @@ public DashboardUserEditVO transfer2EditVO(final DashboardUserVO dashboardUserVO
                     vo.setDateCreated(data.getDateCreated());
                     vo.setDateUpdated(data.getDateUpdated());
                     return vo;
-                }).orElseGet(null);
+                }).orElse(null);
```
yu199195, line 91: why not modify orElseGet to orElse

dragon-zhang: when I use `orElseGet(()->null)`, IDEA tells me to simplify to `orElse(null)`

dragon-zhang: I think `lazy evaluation` is not needed, not create any instances here.

yu199195 (approved):

Issue #2812, opened 2022-01-26
[Question] <should synchronized `instance` rather then `class`>


dragon-zhang:
### Question

In `org.apache.shenyu.admin.disruptor.subscriber.MetadataExecutorSubscriber#executor`
```java
    @Override
    public void executor(final Collection<MetaDataRegisterDTO> metaDataRegisterDTOList) {
        for (MetaDataRegisterDTO metaDataRegisterDTO : metaDataRegisterDTOList) {
            ShenyuClientRegisterService shenyuClientRegisterService = this.shenyuClientRegisterService.get(metaDataRegisterDTO.getRpcType());
            Objects.requireNonNull(shenyuClientRegisterService);
            //here, should be replace to "synchronized (shenyuClientRegisterService)"
            synchronized (ShenyuClientRegisterService.class) {
                shenyuClientRegisterService.register(metaDataRegisterDTO);
            }
        }
    }
```
`Shenyuclientregisterservice.Class` should not be `synchronized`.

PR #2813, opened 2022-01-26
[ISSUE #2812] synchronized instance rather then class

dragon-zhang:
fix https://github.com/apache/incubator-shenyu/issues/2812


shenyu-admin/src/main/java/org/apache/shenyu/admin/disruptor/subscriber/MetadataExecutorSubscriber.java
```diff
@@ -47,7 +47,7 @@ public void executor(final Collection<MetaDataRegisterDTO> metaDataRegisterDTOLi
         for (MetaDataRegisterDTO metaDataRegisterDTO : metaDataRegisterDTOList) {
             ShenyuClientRegisterService shenyuClientRegisterService = this.shenyuClientRegisterService.get(metaDataRegisterDTO.getRpcType());
             Objects.requireNonNull(shenyuClientRegisterService);
-            synchronized (ShenyuClientRegisterService.class) {
+            synchronized (shenyuClientRegisterService) {
```
yu199195, line 50: why ? can you express the meaning of such a change?

yu199195 (approved):

Issue #2679, opened 2021-12-31
[Task] Add integration test for more http api


JooKS-me:
### Description

Hi community,

In `DividePluginTest`, there is only `/order/save` is covered. But we should test all http api in `shenyu-examples-http`. So we need to do the following task.

If you want to participate, please leave a message under issue.

### Task List

- [x] add integration test for http api in `HttpTestController` @totalo 
- [x] add integration test for http api in `OrderController` @totalo 
- [x] add integration test for http api in `RequestController` @gaochuang98 
- [x] add integration test for http api in `ShenyuClientPathController` @gaochuang98 
- [x] add integration test for http api in `SpringMvcMappingPathController` @QinHaiqi 
- [x] Add an controller related to file upload and download in `shenyu-examples-http`, and add integration test @QinHaiqi 

The integration test is expected to add in `shenyu-integrated-test-http`, and you can new a test class in `org.apache.shenyu.integrated.test.http` while doing integrated test.

QinHaiqi:
give me a chance

QinHaiqi:
give me a chance SpringMvcMappingPathController and shenyu-examples-http

JooKS-me:
> give me a chance SpringMvcMappingPathController and shenyu-examples-http

good luck~

QinHaiqi:
let me try **SpringMvcMappingPathController** and  **shenyu-examples-http**

totalo:
I want to try `HttpTestController` and `OrderController`

gaochuang98:
give me a chance ShenyuClientPathController and RequestController

JooKS-me:
@totalo @gaochuang98 Thanks for help

AhahaGe:
please assgin me dubbo example project

PR #2820, opened 2022-01-29
[ISSUE #2679] Add an controller related to file upload and download

QinHaiqi:

Make sure that:


shenyu-examples/shenyu-examples-http/pom.xml
```diff
@@ -35,6 +35,11 @@
     </properties>
 
     <dependencies>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
```
yu199195, line 39: why add this

yu199195:
pls check integration http test

shenyu-examples/shenyu-examples-http/pom.xml
QinHaiqi: This dependency is required for file upload and download in this way. But this approach will conflict with the existing code, so I'm going to try to write it with weblux

PR #2823, opened 2022-02-01
[ISSUE 2754] refactor admin buildHandle about register uri.

lianjunwei:
This bug has been fixed in version 2.4.2. But the code can still be optimized, because when the handler is [], there is no need to iterate and compare again to get diffList.


yu199195:
hi， pls check ci

lianjunwei:
> hi， pls check ci

hi, Unused import code has been removed .

yu199195 (approved):

PR #2839, opened 2022-02-08
[ISSUE #2831] fix double log

### LINKED ISSUE #2831: [BUG] <Double the amount of logs output>

### Current Behavior

We currently have around 500,000 DAU and around 2 million MAU.
The output out file is not packed and compressed for logs, which can easily cause the server storage memory to be full.

image

so.
image

I think it's best not to output logs here.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
ShenYu version(s):
```

### Debug logs

_No response_

### Anything else?

_No response_

impactCn:
Fixes #2831

Make sure that:

### DIFF
--- a/shenyu-admin/src/main/resources/logback.xml
+++ b/shenyu-admin/src/main/resources/logback.xml
@@ -30,19 +30,6 @@
 
     <property name="LOG_PATTERN" value="%d{yyyy-MM-dd HH:mm:ss} [%thread] %-5level %logger - %msg%n"/>
 
-    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
-        <encoder>
-            <pattern>${LOG_PATTERN}</pattern>
-        </encoder>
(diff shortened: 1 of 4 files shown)


tuohai666 (changes requested):
The double logs come from the logback that output to both file and console.
The logs output to console are redirected to LOG_FILES, and the logs output to file which is LOG_FILES. That makes the LOG_FILES contains double logs.
It's better to solve this problem by depress the logback logs output to console. 

While it's necessary to retain `>> ${LOG_FILES}` in order to record the logs when VM fail or other errors.
Another reason why we have to retain `>> ${LOG_FILES}` is that when the process run in a docker container the entrypoint.sh will tail LOG_FILES immediately before the logback create that log file.

yu199195 (approved):

tuohai666 (approved):

Issue #2860, opened 2022-02-10
[Question] better for spring cloud client to set port automatically?


dragon-zhang:
### Question

In `org.apache.shenyu.client.springcloud.init.ContextRegisterListener#onApplicationEvent`, if `org.springframework.context.event.ContextRefreshedEvent` change to `org.springframework.boot.web.context.WebServerInitializedEvent`, we can use `org.springframework.boot.web.server.WebServer#getPort` to set port automatically, so change it ?
```java
    @Override
    public void onApplicationEvent(@NonNull final WebServerInitializedEvent webServerInitializedEvent) {
        if (!registered.compareAndSet(false, true)) {
            return;
        }
        if (Boolean.TRUE.equals(isFull)) {
            publisher.publishEvent(buildMetaDataDTO());
        }
        //Note: In this way, no matter what container is actually used,
        //you can get the port that is actually started in the end.
        int mergedPort = port <= 0 ? webServerInitializedEvent.getWebServer().getPort() : port;
        publisher.publishEvent(buildUriRegisterDTO(mergedPort));
    }
```

yu199195:
can support spring 4.x ？ WebServer is spring 5.0

dragon-zhang:
I also adapted spring MVC cleint. I will try the spring 4.x version

dragon-zhang:
Do I need to submit code to the 'shenyu-examples' module?

yu199195:
not used WebServerInitializedEvent

yu199195:
now， can not support spring boot 1.x？

dragon-zhang:
> not used WebServerInitializedEvent

I've already done so. In addition, I'm sure `shenyu-client` has compatibility problems with lower versions of springboot. Take a look at the warehouse I created: https://github.com/dragon-zhang/shenyu-examples-spring4x

dragon-zhang:
> now， can not support spring boot 1.x？

I'm trying.

dragon-zhang:
see https://github.com/apache/incubator-shenyu/pull/2861

PR #2861, opened 2022-02-10
[ISSUE #2860] spring cloud client auto set port

dragon-zhang:
you can run `org.apache.shenyu.examples.springcloud.ShenyuTestSpringCloudApplication` to see result directly.

### DIFF
--- a/shenyu-client/shenyu-client-http/shenyu-client-springcloud/src/main/java/org/apache/shenyu/client/springcloud/init/ContextRegisterListener.java
+++ b/shenyu-client/shenyu-client-http/shenyu-client-springcloud/src/main/java/org/apache/shenyu/client/springcloud/init/ContextRegisterListener.java
@@ -88,6 +88,8 @@ public class ContextRegisterListener implements ApplicationListener<WebServerIni
         if (Boolean.TRUE.equals(isFull)) {
             publisher.publishEvent(buildMetaDataDTO());
         }
+        //Note: In this way, no matter what container is actually used,
+        //you can get the port that is actually started in the end.
         int mergedPort = port <= 0 ? webServerInitializedEvent.getWebServer().getPort() : port;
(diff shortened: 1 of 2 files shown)


shenyu-client/shenyu-client-http/shenyu-client-springcloud/src/main/java/org/apache/shenyu/client/springcloud/init/ContextRegisterListener.java
```diff
@@ -28,35 +28,36 @@
...
+import java.util.Optional;
 import java.util.Properties;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * The type Context register listener.
  */
-public class ContextRegisterListener implements ApplicationListener<ContextRefreshedEvent> {
-    
+public class ContextRegisterListener implements ApplicationListener<WebServerInitializedEvent> {
```
yu199195, line 43: WebServerInitializedEvent  it is spring version？， if user is spring 4.x，Does that not work?

dragon-zhang: when I use springboot 1.5.22.RELEASE and shenyu-spring-boot-starter-client-springcloud 2.4.2, springboot can not start.

dragon-zhang: I guess `WebServerInitializedEvent` can not work.

dragon-zhang: Think it can be implemented in combination with @ConditionalOnClass, I will reimplement it later.

Issue #2847, opened 2022-02-09
[BUG] <the retryCount of rule in divide plugin does not work>


lzqdename:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior
![1644379947(1)](https://user-images.githubusercontent.com/7373911/153120614-10d3cc04-f217-40ba-b426-785e66521dbd.png)

under this settings, the retryCount does not work rightly
---
analyze from log,the webclient just invoke 1 times, but 3 times of roundbin strategy expected!

### Expected Behavior

3 times of roundbin strategy expected!

### Environment

```markdown
ShenYu version(s):2.4.2
```

dragon-zhang:
I'll take care of it.

PR #2873, opened 2022-02-11
[ISSUE #2847] fix retryCount not work bug

dragon-zhang:
1.  run `org.apache.shenyu.admin.ShenyuAdminBootstrap`;
2.  run `org.apache.shenyu.bootstrap.ShenyuBootstrapApplication`;
3.  run `org.apache.shenyu.examples.http.ShenyuTestHttpApplication`;
4.  then send request:
```text
POST http://localhost:9195/http/shenyu/client/timeout
Accept: application/json
Content-Type: application/json
```
watch `org.apache.shenyu.examples.http.controller.ShenyuClientPathController#timeout`, you can see the result is correct.

### DIFF
--- a/shenyu-plugin/shenyu-plugin-httpclient/src/main/java/org/apache/shenyu/plugin/httpclient/WebClientPlugin.java
+++ b/shenyu-plugin/shenyu-plugin-httpclient/src/main/java/org/apache/shenyu/plugin/httpclient/WebClientPlugin.java
@@ -18,6 +18,7 @@
 package org.apache.shenyu.plugin.httpclient;
 
 import io.netty.channel.ConnectTimeoutException;
+import io.netty.handler.timeout.ReadTimeoutException;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.shenyu.common.constant.Constants;
 import org.apache.shenyu.common.enums.PluginEnum;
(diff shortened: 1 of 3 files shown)


shenyu-plugin/shenyu-plugin-httpclient/src/main/java/org/apache/shenyu/plugin/httpclient/WebClientPlugin.java
```diff
@@ -118,7 +120,7 @@ public boolean skip(final ServerWebExchange exchange) {
                 .exchange()
                 .doOnError(e -> LOG.error(e.getMessage(), e))
                 .timeout(Duration.ofMillis(timeout))
-                .retryWhen(Retry.onlyIf(x -> x.exception() instanceof ConnectTimeoutException)
+                .retryWhen(Retry.anyOf(TimeoutException.class, ConnectTimeoutException.class, ReadTimeoutException.class)
```
yu199195, line 123: can add exception.class?

dragon-zhang: > can add exception.class?

How about `Throwable`? If ok, I'll change to `reactor.retry.Retry#any`.

yu199195 (approved):

Issue #2916, opened 2022-02-21
[Question] How about change ribbon `ServerListRefreshInterval` from 30s to 10s ?


dragon-zhang:
### Question

If the data of the registry (such as zookeeper) has been updated, `SpringCloudPlugin` will still select `unavailable` instances when using `Ribbon` for load balancing, change `ServerListRefreshInterval` from `30s` to `10s` can't completely solve this problem, but it can be avoided to a great extent.

dragon-zhang:
Oh, maybe it's better to refactor like DividePlugin ?



发自我的iPhone


------------------ Original ------------------
From: dragon-zhang ***@***.***&gt;
Date: Mon,Feb 21,2022 5:45 PM
To: apache/incubator-shenyu ***@***.***&gt;
Cc: dragon-zhang ***@***.***&gt;, Your activity ***@***.***&gt;
Subject: Re: [apache/incubator-shenyu] [Question] How about change ribbon `ServerListRefreshInterval` from 30s to 10s ? (Issue #2916)





 
Question
 
If the data of the registry (such as zookeeper) has been updated, SpringCloudPlugin will still select unavailable instances when using Ribbon for load balancing, change ServerListRefreshInterval from 30s to 10s can't completely solve this problem, but it can be avoided to a great extent.
 
—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

yu199195:
this props，maybe can dynamic config~

dragon-zhang:
In `ShenyuClientRegisterSpringCloudServiceImpl#buildHandle`, I found that `URLs` has been set in the `handle` field of the `selector`, which means that we can do this directly in 'SpringCloudPlugin#doExecute':
```java
    private List<Upstream> convertUpstreamList(final List<DivideUpstream> upstreamList) {
        return upstreamList.stream().map(u -> Upstream.builder()
                .protocol(u.getProtocol())
                .url(u.getUpstreamUrl())
                .weight(u.getWeight())
                .status(u.isStatus())
                .timestamp(u.getTimestamp())
                .warmup(u.getWarmup())
                .build()).collect(Collectors.toList());
    }

    @Override
    protected Mono<Void> doExecute(final ServerWebExchange exchange, final ShenyuPluginChain chain, final SelectorData selector,
                                   final RuleData rule) {
        if (Objects.isNull(rule)) {
            return Mono.empty();
        }
[...]

PR #2922, opened 2022-02-22
[type:refactor][ISSUE #2916]change ribbon default ServerListRefreshInterval from 30s to 10s

dragon-zhang:
see https://github.com/apache/incubator-shenyu/issues/2916.

You can follow these steps to verify whether the results are correct :
1. run `ShenyuAdminBootstrap`;
2. turn on `SpringCloudPlugin` in `shenyu-admin`;
3. run `EurekaServerApplication`;
4. run `ShenyuTestSpringCloudApplication`;
5. enable `Ribbon` then run `ShenyuBootstrapApplication`;
6. run `spring-cloud-test-api.http`.

### DIFF
--- a/shenyu-common/src/main/java/org/apache/shenyu/common/config/ShenyuConfig.java
+++ b/shenyu-common/src/main/java/org/apache/shenyu/common/config/ShenyuConfig.java
@@ -33,23 +33,25 @@ import java.util.stream.Stream;
 public class ShenyuConfig {
 
     private SwitchConfig switchConfig = new SwitchConfig();
-    
+
     private FileConfig file = new FileConfig();
-    
+
     private ExcludePath exclude = new ExcludePath();
 
     private FallbackPath fallback = new FallbackPath();
-    
+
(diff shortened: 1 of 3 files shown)


shenyu-common/src/main/java/org/apache/shenyu/common/config/ShenyuConfig.java
```diff
@@ -880,40 +900,40 @@ public void setAllowCredentials(final boolean allowCredentials) {
...
-    
+
         /**
          * Instantiates a new Instance config.
          *
          * @param registerType the register type
-         * @param serverLists the server lists
-         * @param props the props
+         * @param serverLists  the server lists
+         * @param props        the props
```
yu199195, line 929: can not format code

dragon-zhang: ok

yu199195 (approved):

yu199195 (approved):

Issue #1888, opened 2021-08-06
Optimize the password encryption and password change logic for the admin user


dengliming:
## Task

Currently, We use the AES algorithm to encrypt the password and provide the password back to the front end. I think It's not safe. 

![image](https://user-images.githubusercontent.com/7796156/128477713-2de63807-7a51-4e3f-9b90-e24f6cfc81bc.png)

I suggest the following optimization:

1. The API does not return the password field. See `org.apache.shenyu.admin.controller.DashboardUserController#detailDashboardUser`
2. Consider Using sha512 algorithm to encrypt the password (**Need to discuss**)

Please read the [Contribution Guideline](https://shenyu.apache.org/community/contributor/) before submitting the PR

rom1c77:
i think the password field value need clear.

qurenneng:
I'll take care of this

dengliming:
@qurenneng Hi~ any progress?

qurenneng:
Solved in the last two days



At 2021-08-21 10:21:31, "dengliming" ***@***.***> wrote:

@qurenneng Hi~ any progress?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

qurenneng:
Solved in the last two days

dengliming:
which pr?

qurenneng:
the password (Need to discuss)  This is what we need to do here

erdengk:
Hi， @dengliming 
It seems that no one has contributed to this issue yet. I have done some research and have ideas for improvement. I would like to ask you to review it.

1. Regarding the first point you mentioned

>The API does not return the password field

I looked at your proposed method and detailDashboardUser doesn't return a password field. (Maybe I read the code wrong, please help me)

2. Consider Using sha512 algorithm to encrypt the password

The password encryption algorithm used in shenyu (v2.4.2) is the AES algorithm, which is an algorithm that supports decryption. If the user's database is leaked, the user name and password may be cracked.

The SHA-512 algorithm is an irreversible algorithm. Compared with the MD5 algorithm, the SHA-512 algorithm achieves that two different data have different hash values. Therefore, the SHA-512 algorithm has a higher level of security.

I noticed that when shenyu changes the password, it does not verify the previous password, and directly overwrites it, so the encrypted password stored in the database does not need to be decrypted.

[...]

PR #2928, opened 2022-02-23
[ISSUE #1888] Optimize the password encryption 

erdengk:
// Describe your PR here; eg. 

Fixes  https://github.com/apache/incubator-shenyu/issues/1888

The purpose of this pr is to modify the password encryption method.

There are the following changes
1. Modify the encrypted password in the initialization sql
2. Add the SHA encryption tool class, and add the test code of the encryption tool. Change the encryption method of the login function,
3. Delete the previous code about encryption.

Make sure that:


shenyu-admin/src/main/java/org/apache/shenyu/admin/utils/ShaUtils.java
```diff
@@ -0,0 +1,56 @@
...
+        return Optional.ofNullable(src).map(item -> {
+            if ("".equals(src)) {
+                return null;
+            }
+            try {
+                MessageDigest messageDigest = MessageDigest.getInstance("SHA-512");
+                messageDigest.update(item.getBytes());
+                byte[] byteBuffer = messageDigest.digest();
+                StringBuffer strHexString = new StringBuffer();
+                for (byte b:byteBuffer) {
```
yu199195, line 46: for （） replace to lambda?

shenyu-admin/src/main/java/org/apache/shenyu/admin/utils/ShaUtils.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * sh512 Encryption string.
+     *
+     * @param src the src
+     * @return the string
+     */
+    public static String shaEncryption(final String src) {
+        return Optional.ofNullable(src).map(item -> {
+            if ("".equals(src)) {
```
yu199195, line 38: used StringUtils.isEmty() , then  return

erdengk: ok

yu199195 (approved):

Issue #2055, opened 2021-09-12
[Question] <Resilience4JPlugin's conf limitForPeriod invalid>


rockpig:
### Question

I test resilience4J plugin, and set limitForPeriod 2 for RuleData, but it was changed 50 in 
 `org.apache.shenyu.common.dto.convert.Resilience4JHandle#checkData() : 393` : 

> resilience4JHandle.setLimitForPeriod(Math.max(resilience4JHandle.getLimitForPeriod(), Constants.LIMIT_FOR_PERIOD));

If I went to limit 2 request per second, it will failed.

I think it should be fix to:

> resilience4JHandle.setLimitForPeriod(Math.min(resilience4JHandle.getLimitForPeriod(), Constants.LIMIT_FOR_PERIOD));

So It is a bug or the way I used has problem?

SaberSola:
Hi, Perhaps this method named 'checkData' is not reasonable,you are right,can you provide a pr fix this bug or remove the 'checkData' method directly

rockpig:
Of course, I will fix it.

PR #2929, opened 2022-02-23
set default value if the customer's config is invalid for resilience4…

### LINKED ISSUE #2927: [Question] For Resilience4J plugin, why using max value between user configuration and preset, instead of making the preset value as default?
### Question

You can see below codes in `Resilience4JHandle#checkData`. The question is as mentioned in title. Is it by design?

image

BTW, if it's purpose is to use preset value as default, I can help to fix it.

Similar question, #2055 .

hgaol:
…j plugin

Fix #2927 and #2055 . For resilience4j plugin.
- Rename `CIRCUIT_ENABLE` to `CIRCUIT_DISABLE` to avoid misunderstanding. Disable is the default value.
- Verify the number and set to default if not valid.

Make sure that:

### DIFF
--- a/shenyu-common/src/main/java/org/apache/shenyu/common/constant/Constants.java
+++ b/shenyu-common/src/main/java/org/apache/shenyu/common/constant/Constants.java
@@ -280,7 +280,12 @@ public interface Constants {
     /**
      * circuitBreaker circuitEnable.
      */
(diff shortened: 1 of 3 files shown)


hgaol:
There is error for integration test 
image

yu199195 (approved):

Issue #2627, opened 2021-12-20
can dividePlugin support to retry  another upstream but not retry only on one upstream  ,when is fail .


yuzhijia88:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Environment

```markdown
ShenYu version(s): 2.4.1
```

yu199195:
yes， support all upstream

dragon-zhang:
For example, there are 3 upstreams like `http://localhost:1111`、`http://localhost:1112` and `http://localhost:1113`, when call `http://localhost:1111` fails, I think he means to call `http://localhost:1112` or `http://localhost:1113` for retry, not just call `http://localhost:1111` for retry, don't you ? @yuzhijia88

yuzhijia88:
> For example, there are 3 upstreams like `http://localhost:1111`、`http://localhost:1112` and `http://localhost:1113`, when call `http://localhost:1111` fails, I think he means to call `http://localhost:1112` or `http://localhost:1113` for retry, not just call `http://localhost:1111` for retry, don't you ? @yuzhijia88

yes

dragon-zhang:
> > For example, there are 3 upstreams like `http://localhost:1111`、`http://localhost:1112` and `http://localhost:1113`, when call `http://localhost:1111` fails, I think he means to call `http://localhost:1112` or `http://localhost:1113` for retry, not just call `http://localhost:1111` for retry, don't you ? @yuzhijia88
> 
> yes

It can be done and if it done, you need to add some config for `DividePlugin`(in order to be compatible with the old version of `DividePlugin`), can you accept it ?

yuzhijia88:
> > > For example, there are 3 upstreams like `http://localhost:1111`、`http://localhost:1112` and `http://localhost:1113`, when call `http://localhost:1111` fails, I think he means to call `http://localhost:1112` or `http://localhost:1113` for retry, not just call `http://localhost:1111` for retry, don't you ? @yuzhijia88
> > 
> > 
> > yes
> 
> It can be done and if it done, you need to add some config for `DividePlugin`(in order to be compatible with the old version of `DividePlugin`), can you accept it ?

yes

dragon-zhang:
[...]

PR #2939, opened 2022-02-26
[ISSUE #2627] support failover RetryStrategy

dragon-zhang:
see https://github.com/apache/incubator-shenyu/issues/2627, and I need some help:

First, I don't know how to add `backoff` (you can see it after `todo`);

Second, my implementation is `not elegant enough`, but it is not easy to reuse the code in other plugins such as `DividePlugin` and `URIPlugin`;

### DIFF
--- a/shenyu-plugin/shenyu-plugin-httpclient/src/main/java/org/apache/shenyu/plugin/httpclient/WebClientPlugin.java
+++ b/shenyu-plugin/shenyu-plugin-httpclient/src/main/java/org/apache/shenyu/plugin/httpclient/WebClientPlugin.java
@@ -17,6 +17,7 @@
 
 package org.apache.shenyu.plugin.httpclient;
 
+import com.google.common.collect.Sets;
 import io.netty.channel.ConnectTimeoutException;
 import io.netty.handler.timeout.ReadTimeoutException;
 import org.apache.commons.collections4.CollectionUtils;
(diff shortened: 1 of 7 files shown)


shenyu-plugin/shenyu-plugin-httpclient/src/main/java/org/apache/shenyu/plugin/httpclient/WebClientPlugin.java
```diff
@@ -100,12 +121,86 @@ public boolean skip(final ServerWebExchange exchange) {
...
+                return Mono.error(new ShenyuException(ShenyuResultEnum.CANNOT_FIND_HEALTHY_UPSTREAM_URL.getMsg()));
+            }
+            final URI newUri = buildUri(upstream, exchange, shenyuContext);
+            // in order not to affect the next retry call, this uri needs to be excluded
+            exclude.add(newUri);
+            return handleRequest(method, newUri, exchange, timeout);
+        });
+    }
+
+    private URI buildUri(final Upstream upstream,
```
yu199195, line 176: can refactor common method， because in URI plugin？

dragon-zhang: > can refactor common method， because in URI plugin？

Yes, most of in `URIPlugin`, I'll try for it.

dragon-zhang:
Because of the large amount of changes, I will split it into multiple PRs and close this PR.

PR #2965, opened 2022-03-02
Refactor code

erdengk:
There are two changes in this pr
1. The first is to use String instead of Map. I don't think it's necessary to use a map just for two string variables, and using a map comes at the cost of traversing the map.

   Before:

   ```java
           Map<String, Object> loginMap = new HashMap<>(2);
           loginMap.put(Constants.LOGIN_NAME, username);
           loginMap.put(Constants.PASS_WORD, password);
           String result = OkHttpTools.getInstance().get(url, loginMap);
   ```

   After:

   ```java
    String result = OkHttpTools.getInstance().get(url, userName, passWord);
   ```

2. The second is to delete unit test statements that always result in true. As shown below, the current statement is always true.
  ```java
  ShenyuException shenyuException = new ShenyuException(throwable);
  // Condition 'serializerException instanceof SerializerException' is always 'true' 
  assertTrue(shenyuException instanceof ShenyuException); 
  ```

Make sure that:

### DIFF
--- a/shenyu-register-center/shenyu-register-client/shenyu-register-client-http/src/main/java/org/apache/shenyu/register/client/http/utils/OkHttpTools.java
+++ b/shenyu-register-center/shenyu-register-client/shenyu-register-client-http/src/main/java/org/apache/shenyu/register/client/http/utils/OkHttpTools.java
@@ -23,9 +23,9 @@ import okhttp3.Request;
 import okhttp3.RequestBody;
 import okhttp3.Headers;
 import okhttp3.HttpUrl;
+import org.apache.shenyu.common.constant.Constants;
 
 import java.io.IOException;
-import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
 /**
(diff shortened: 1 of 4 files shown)


yu199195:
pls check ci

shenyu-register-center/shenyu-register-client/shenyu-register-client-http/src/main/java/org/apache/shenyu/register/client/http/utils/OkHttpTools.java
```diff
@@ -99,16 +99,19 @@ public String post(final String url, final String json, final Headers headers) t
...
      *
      * @param url   the url
-     * @param query the query
+     * @param userName the userName
+     * @param passWord the passWord
      * @return the http result
      * @throws IOException the io exception
      */
-    public String get(final String url, final Map<String, Object> query) throws IOException {
+    public String get(final String url, final String userName, final String passWord) throws IOException {
```
yu199195, line 107: Don't change it like that

yu199195: can add  another method  .   userName  ，passWord  maybe in Map

erdengk: Hi, @yu199195 
I modified this because only one method uses the get function. And this method is only used here and will not affect other methods. So I refactored the code like this.

erdengk: Hi ,@yu199195 
I modified this because only one method uses the get function. And this method is only used here and will not affect other methods.So I refactored the code like this.

yu199195: This method may later be called with multiple parameters

yu199195 (approved):

Issue #2931, opened 2022-02-24
hystrix plugin rule configuration, must restart the gateway to take effect


zouchangfu:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Environment

```markdown
ShenYu version(s):v2.4.1
```

hutaishi:
Who is doing this? If not, can you assign it to me and let me exercise

lxl910128:
hello , you can use this feature synchronous config.

<img width="1342" alt="image" src="https://user-images.githubusercontent.com/8736745/155937633-925d96da-00db-4213-bf27-902580ad4213.png">

erdengk:
@lxl910128  
I tried sync but it still doesn't work.
I think this is also a bug. I created a new issue #2992

PR #2985, opened 2022-03-05
[ISSUE #2909] Add integration test for Hystrix plugin

### LINKED ISSUE #2909: [Volunteer] Add integration test for Hystrix plugin
### Volunteer help

- [ ] Add integration test for Hystrix plugin

We should test as many parameter configurations as possible.

erdengk:

Fixes #2909 

- [x] Your local test passed mvn clean install -Dmaven.javadoc.skip=true.
- [x] Add integration test for Hystrix plugin
- [ ] write correct tests

### DIFF
--- a/shenyu-integrated-test/shenyu-integrated-test-http/src/test/java/org/apache/shenyu/integrated/test/http/combination/HystrixPluginTest.java
+++ b/shenyu-integrated-test/shenyu-integrated-test-http/src/test/java/org/apache/shenyu/integrated/test/http/combination/HystrixPluginTest.java
@@ -1,0 +1,118 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.shenyu.integrated.test.http.combination;
+
+import org.apache.shenyu.common.constant.Constants;
+import org.apache.shenyu.common.dto.ConditionData;
+import org.apache.shenyu.common.dto.convert.rule.HystrixHandle;
+import org.apache.shenyu.common.enums.HystrixIsolationModeEnum;
+import org.apache.shenyu.common.enums.OperatorEnum;
+import org.apache.shenyu.common.enums.ParamTypeEnum;
+import org.apache.shenyu.common.enums.PluginEnum;
+import org.apache.shenyu.common.utils.JsonUtils;
+import org.apache.shenyu.integratedtest.common.AbstractPluginDataInit;
+import org.apache.shenyu.integratedtest.common.helper.HttpHelper;
+import org.apache.shenyu.integratedtest.common.result.ResultBean;
+import org.apache.shenyu.web.controller.LocalPluginController.RuleLocalData;
+import org.hamcrest.CoreMatchers;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public final class HystrixPluginTest extends AbstractPluginDataInit {
+
+    private static final String TEST_HYSTRIX_PATH = "/http/test/success";
+
+    private static final String TEST_HYSTRIX_BAD_REQUEST_PATH = "/http/test/request/badrequest";
+
+    @BeforeEach
+    public void setup() throws IOException {
+        String pluginResult = initPlugin(PluginEnum.HYSTRIX.getName(), "");
+        assertThat(pluginResult, is("success"));
+    }
+
+    @Test
+    public void testPass() throws IOException, ExecutionException, InterruptedException {
+        String selectorAndRulesResult =
+                initSelectorAndRules(PluginEnum.HYSTRIX.getName(), "", buildSelectorConditionList(), buildRuleLocalDataList(null));
+        assertThat(selectorAndRulesResult, is("success"));
+        Future<ResultBean> resp = this.getService().submit(() -> HttpHelper.INSTANCE.getFromGateway(TEST_HYSTRIX_PATH, ResultBean.class));
+        assertEquals(200, resp.get().getCode());
+
+    }
+
+    @Test
+    public void testFallbackUri() throws IOException, ExecutionException, InterruptedException {
+        String selectorAndRulesResult =
+                initSelectorAndRules(PluginEnum.HYSTRIX.getName(), "", buildSelectorConditionList(), buildRuleLocalDataList(null));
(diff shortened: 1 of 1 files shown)


erdengk:
@JooKS-me 
This is my first time writing integration tests for a plugin. I learned test code for other plugins.
But I ran into some problems. I can't implement circuit breaker by configuring HystrixPlugin.
Currently its configuration code is as follows

```
   HystrixHandle hystrixHandle = new HystrixHandle();
        hystrixHandle.setGroupKey("contextPath");
        hystrixHandle.setCommandKey("commandKey");
        hystrixHandle.setMaxConcurrentRequests(0);
        hystrixHandle.setErrorThresholdPercentage(0);
        hystrixHandle.setRequestVolumeThreshold(0);
        hystrixHandle.setSleepWindowInMilliseconds(1);
        hystrixHandle.setTimeout(Constants.TIME_OUT);
        hystrixHandle.setCallBackUri(TEST_HYSTRIX_BAD_REQUEST_PATH);
        HystrixHandle.HystrixThreadPoolConfig hystrixThreadPoolConfig = new HystrixHandle.HystrixThreadPoolConfig();
        hystrixThreadPoolConfig.setCoreSize(0);
        hystrixThreadPoolConfig.setMaximumSize(0);
        hystrixThreadPoolConfig.setMaxQueueSize(0);
        hystrixHandle.setHystrixThreadPoolConfig(hystrixThreadPoolConfig);
```

I learned how Hystrix works. In my understanding, if the `MaxConcurrentRequests` parameter is set to 1 or 0 , the circuit breaker will be triggered, but it is not actually triggered. 
Maybe there is something wrong with my settings? 
I've tried for a long time and can't solve this problem, please help me.🙏

JooKS-me:
> I learned how Hystrix works. In my understanding, if the `MaxConcurrentRequests` parameter is set to 1 or 0 , the circuit breaker will be triggered, but it is not actually triggered. Maybe there is something wrong with my settings? I've tried for a long time and can't solve this problem, please help me.🙏

I think there may be some bug in Hystrix plugin.

erdengk:
Seems like yes, I'm writing the description of the issue.

JooKS-me:
You can add details under #2931, which I think is related to this issue.

erdengk:
Integration testing for hystrix is complete.
There are two test cases in this integration test. One is that the test request passes normally, and the other is to implement the Hystrix fuse mechanism by setting the semaphore.

shenyu-examples/shenyu-examples-http/src/main/java/org/apache/shenyu/examples/http/controller/HttpTestController.java
```diff
@@ -367,4 +367,30 @@ private UserDTO buildUser(final String id, final String name) {
...
         return userDTO;
     }
+
+    /**
+     * the waf pass.
+     *
+     * @return response. result bean
+     */
+    @GetMapping ("/hystrix/pass")
+    public ResultBean hystrixPass() {
```
JooKS-me, line 377: IMO, we can reuse existing endpoint, no need to add new ones, WDYT?

shenyu-integrated-test/shenyu-integrated-test-http/src/test/java/org/apache/shenyu/integrated/test/http/combination/HystrixPluginTest.java
```diff
@@ -0,0 +1,128 @@
...
+    public void testPass() throws IOException, ExecutionException, InterruptedException {
+        String selectorAndRulesResult =
+                initSelectorAndRules(PluginEnum.HYSTRIX.getName(), "", buildSelectorConditionList(), buildRuleLocalDataList(TEST_HYSTRIX_PATH));
+        assertThat(selectorAndRulesResult, is("success"));
+        Future<ResultBean> resp = this.getService().submit(() -> HttpHelper.INSTANCE.getFromGateway(TEST_HYSTRIX_PATH, ResultBean.class));
+        assertEquals(200, resp.get().getCode());
+    }
+
+    @Test
+    public void testFallbackBySmeaphore() throws IOException, ExecutionException, InterruptedException {
```
JooKS-me, line 76: typo(`Smeaphore`) :)

shenyu-integrated-test/shenyu-integrated-test-http/src/test/java/org/apache/shenyu/integrated/test/http/combination/HystrixPluginTest.java
```diff
@@ -0,0 +1,128 @@
...
+        String selectorAndRulesResult =
+                initSelectorAndRules(PluginEnum.HYSTRIX.getName(), "", buildSelectorConditionList(), buildRuleLocalDataList(TEST_HYSTRIX_BAD_REQUEST_PATH));
+        assertThat(selectorAndRulesResult, CoreMatchers.is("success"));
+        Set<String> resultSet = new HashSet<>();
+        Type returnType = new TypeToken<Map<String, Object>>() {
+        }.getType();
+        Future<Map<String, Object>> resp = this.getService().submit(() -> HttpHelper.INSTANCE.getFromGateway(TEST_HYSTRIX_BAD_REQUEST_PATH, returnType));
```
JooKS-me, line 83: It would be better to change the variable name to `resp0`~

shenyu-web/src/main/java/org/apache/shenyu/web/fallback/DefaultFallbackController.java
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+/**
+ * The default fallback for hystrix plugin and resilience4j plugin.
+ */
+@RestController
+@RequestMapping("/fallback")
+public class DefaultFallbackController {
```
JooKS-me, line 33: Is this a duplicate of #3028?

JooKS-me (changes requested):

shenyu-examples/shenyu-examples-http/src/main/java/org/apache/shenyu/examples/http/controller/HttpTestController.java
erdengk: https://github.com/apache/incubator-shenyu/blob/e3456e91a093057455fbf09916cca09a3982e9ca/shenyu-examples/shenyu-examples-http/src/main/java/org/apache/shenyu/examples/http/controller/HttpTestController.java#L159-L164


https://github.com/apache/incubator-shenyu/blob/e3456e91a093057455fbf09916cca09a3982e9ca/shenyu-examples/shenyu-examples-http/src/main/java/org/apache/shenyu/examples/http/controller/HttpTestController.java#L204-L212

I noticed that the previous test also created a corresponding method, so I also created a new method belonging to the Hystrix plugin. I think doing this reduces the coupling between methods.

shenyu-integrated-test/shenyu-integrated-test-http/src/test/java/org/apache/shenyu/integrated/test/http/combination/HystrixPluginTest.java
erdengk: ok

shenyu-examples/shenyu-examples-http/src/main/java/org/apache/shenyu/examples/http/controller/HttpTestController.java
JooKS-me: We should probably unify and simplify these endpoints instead of making them more complex and repetitive. . . Of course, you can also do this in subsequent prs.

shenyu-web/src/main/java/org/apache/shenyu/web/fallback/DefaultFallbackController.java
erdengk: This is my fault. I committed the correct changes

shenyu-examples/shenyu-examples-http/src/main/java/org/apache/shenyu/examples/http/controller/HttpTestController.java
erdengk: I think there should be separate controllers for each type of test.
When integrated, a single change affects all tests.
I think the unified approach is suitable for situations that don't change, like ResultBean, only code, message, data .

JooKS-me (approved):
Good job!

PR #3005, opened 2022-03-08
Fixed Possible Json Ordering Permutations in Tests

kaiyaok2:
**Description**
2 flaky(possible to fail under different OS's / future Java versions) tests are found using Nondex when running commands ```mvn edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest={test}``` , where the 2 tests are `common.utils.GsonUtilsTest.testToJson` and `common.utils.JsonUtilsTest.toJson`. The test flakiness is due to direct comparisons between Json Strings and outputs from `JsonUtils.toJson()` and `GsonUtils.getInstance().toJson()`. However, JsonObject-likes do not guarantee entry orders (set-like construction), and internal permutations may occur in the output from `toJson()`. 
**Proposed Fixes**
Using `JsonParser.parse()` to convert string-formed Json's into rigid `JsonElement` before direct comparison.

### DIFF
--- a/shenyu-common/src/test/java/org/apache/shenyu/common/utils/GsonUtilsTest.java
+++ b/shenyu-common/src/test/java/org/apache/shenyu/common/utils/GsonUtilsTest.java
@@ -66,8 +66,11 @@ public class GsonUtilsTest {
     @Test
     public void testToJson() {
         TestObject testObject = generateTestObject();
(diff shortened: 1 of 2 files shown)


yu199195:
@kaiyaok2  good idea， you can add my wechat ：xixy199195

yu199195 (approved):

Issue #3025, opened 2022-03-11
[Question] How about put http request headers into generalContext ?


dragon-zhang:
### Question

In `org.apache.shenyu.plugin.general.context.GeneralContextPlugin#doExecute`, just like below:
```java
    protected Mono<Void> doExecute(final ServerWebExchange exchange, final ShenyuPluginChain chain, final SelectorData selector, final RuleData rule) {
        Map<String, List<GeneralContextHandle>> generalContextHandleMap = GeneralContextPluginDataHandler.CACHED_HANDLE.get().obtainHandle(CacheKeyUtils.INST.getKey(rule));
        Map<String, Map<String, String>> generalContextMap = new HashMap<>();
        HttpHeaders headers = exchange.getRequest().getHeaders();
        generalContextHandleMap.forEach((rpcType, v) -> {
            if (CollectionUtils.isEmpty(v)) {
                return;
            }
            Map<String, String> generalContextMapWithRpcType = new HashMap<>();
            v.forEach(each -> {
                if (StringUtils.isBlank(each.getGeneralContextType()) || StringUtils.isBlank(each.getGeneralContextKey())) {
                    return;
                }
                switch (each.getGeneralContextType()) {
                    case Constants.ADD_GENERAL_CONTEXT_TYPE:
                        generalContextMapWithRpcType.put(each.getGeneralContextKey(), each.getGeneralContextValue());
                        break;
                    case Constants.TRANSMIT_HEADER_TO_GENERAL_CONTEXT_TYPE:
                        generalContextMapWithRpcType.put(StringUtils.isBlank(each.getGeneralContextValue()) ? each.getGeneralContextKey() : each.getGeneralContextValue(),
                                headers.getFirst(each.getGeneralContextKey()));
                        break;
                    default:
                        break;
                }
            });
            // put the request header from the front end into the `generalContextMapWithRpcType`
            exchange.getRequest().getHeaders().forEach((k, val) -> generalContextMapWithRpcType.put(k, String.join(", ", val)));
            generalContextMap.put(rpcType, generalContextMapWithRpcType);
        });
        exchange.getAttributes().put(Constants.GENERAL_CONTEXT, generalContextMap);
        return chain.execute(exchange);
    }
```

yu199195:
it maybe Optimization

KevinClair:
It has been supported, add config in plugin's selector with `transmitHeaderToGeneralContext`;

dragon-zhang:
[...]

PR #3029, opened 2022-03-11
[ISSUE #3025] fix http headers lose bug

dragon-zhang:
see #3025


shenyu-plugin/shenyu-plugin-general-context/src/main/java/org/apache/shenyu/plugin/general/context/GeneralContextPlugin.java
```diff
@@ -60,8 +60,9 @@
                         generalContextMapWithRpcType.put(each.getGeneralContextKey(), each.getGeneralContextValue());
                         break;
                     case Constants.TRANSMIT_HEADER_TO_GENERAL_CONTEXT_TYPE:
+                        final List<String> header = headers.get(each.getGeneralContextKey());
```
KevinClair, line 63: Why this varable is final?

shenyu-plugin/shenyu-plugin-general-context/src/main/java/org/apache/shenyu/plugin/general/context/GeneralContextPlugin.java
```diff
@@ -60,8 +60,9 @@
...
                         generalContextMapWithRpcType.put(StringUtils.isBlank(each.getGeneralContextValue()) ? each.getGeneralContextKey() : each.getGeneralContextValue(),
-                                headers.getFirst(each.getGeneralContextKey()));
+                                CollectionUtils.isEmpty(header) ? null : String.join(", ", header));
```
KevinClair, line 65: `String.join(",", header);`

shenyu-plugin/shenyu-plugin-general-context/src/main/java/org/apache/shenyu/plugin/general/context/GeneralContextPlugin.java
dragon-zhang: > varable

If a parameter can be either mutable or immutable, it's better to make it immutable.

yu199195 (approved):

PR #3035, opened 2022-03-13
When the service shutdown, remov it from the selector

### LINKED ISSUE #3017: [BUG] When the service shutdown, it is not removed from the selector (Divide)

- [ ] #3018

### Current Behavior

With DevidePlugin, there are still requests for those client service after you close them one or more.

### Expected Behavior

After a service is down, the client service is no longer requested until it is restarted.

### Steps To Reproduce

1. Start client services.
2. Stop one or more of them.
3. Send multiple requests to the ShenYu.

### Environment

```markdown
ShenYu version(s): 2.4.2
```

### Debug logs

No error logs.

in AbstractShenyuClientRegisterServiceImpl.doRegisterURI:
if (CollectionUtils.isEmpty(uriList)) return "";

in ShenyuClientRegisterDivideServiceImpl.buildHandle
List<DivideUpstream> diffList = addList.stream().filter(divideUpstream -> !existList.contains(divideUpstream)).collect(Collectors.toList());  //returns diffList size of 0, because addList is the remaining list available, all exist in existList.

### Anything else?

_No response_

jerbo99:
Fixes #3017


shenyu-admin/src/main/java/org/apache/shenyu/admin/disruptor/executor/RegisterServerConsumerExecutor.java
```diff
@@ -64,8 +58,7 @@ public void run() {
     private boolean isValidData(final Object data) {
         if (data instanceof URIRegisterDTO) {
             URIRegisterDTO uriRegisterDTO = (URIRegisterDTO) data;
-            return Objects.nonNull(uriRegisterDTO.getPort())
-                    && StringUtils.isNoneBlank(uriRegisterDTO.getAppName(), uriRegisterDTO.getHost());
+            return StringUtils.isNotBlank(uriRegisterDTO.getContextPath());
```
yu199195, line 61: why this Judgment？

jerbo99: When there is no upstream service, there are only the contextPath and rpcType attributes in the URIRegisterDTO. rpcType is added this time, URIRegisterExecutorSubscriber.findService needs to use.

jerbo99: The status of the upstream service is set in CommonUpstreamUtils.

yu199195:
@jerbo99  you can add my wechat ：xixy199195，i will join you to contributor group

yu199195 (changes requested):
not merged

Issue #3026, opened 2022-03-11
[Question] How about RPC and HTTP using shared thread pool ?


dragon-zhang:
### Question

When a request comes, it can be called either to `webClient` or RPC (Dubbo, sofa, grpc, etc.). Using shared thread pool can greatly improve resource utilization.

yu199195:
@dragon-zhang  can you provide this new featrue？

dragon-zhang:
> 

sure, I can do it.

dragon-zhang:
> @dragon-zhang can you provide this new featrue？

sure, I can do it.

yu199195:
@dragon-zhang Looking forward to your contributor，thanks

PR #3043, opened 2022-03-15
[ISSUE #3026] part1: support MemoryLimitedLinkedBlockingQueue

dragon-zhang:
see #3026 , in `org.apache.shenyu.common.concurrent.MemoryLimitedLinkedBlockingQueueTest`, I write a simple example to tell what it can do. In short, for example, it can limit the maximum memory of the queue used by the `ThreadPoolExecutor` to `100MB`, which is more reliable than the traditional queue(like `LinkedBlockingQueue`) to control the memory through the number of queues.


yu199195:
What scenario is it used in?

dragon-zhang:
> What scenario is it used in?

Blocking queue of shared thread pool.

yu199195 (approved):

Issue #3056, opened 2022-03-18
[BUG] when Dubbo enables multi protocol and `shenyu.client.dubbo.props.port` is configured, `ApacheDubboServiceBeanListener` and `AlibabaDubboServiceBeanListener` gets the wrong port number


dragon-zhang:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

`ApacheDubboServiceBeanListener` gets the wrong port number.

### Expected Behavior

`ApacheDubboServiceBeanListener` gets the correct port number.

### Steps To Reproduce

1.config muti protocol like below:
```yml
dubbo:
  application:
    name: test-dubbo-service
  registry:
    address: zookeeper://127.0.0.1:2181
  protocols:
    dubbo:
      name: dubbo
      port: 20888
    tri:
      name: tri
      port: 20889
  scan:
    base-packages: org.apache.shenyu.examples.apache.dubbo.service.annotation.impl
```
2.using multiple protocols in the same application like below:
```java
@Service(protocol = {"dubbo"})
public class DubboMultiParamServiceImpl implements DubboMultiParamService {
    //......
}
```
```java
@Service(protocol = {"tri"})
public class DubboTestServiceImpl implements DubboTestService {
    //......
}
```
3. run dubbo application with `shenyu-client-apache-dubbo` or `shenyu-client-alibaba-dubbo`

### Environment

```markdown
ShenYu version(s): 2.4.2
```

### Debug logs

nop

### Anything else?

nop

dragon-zhang:
see https://github.com/apache/incubator-shenyu-website/pull/490

PR #3057, opened 2022-03-18
[ISSUE #3056] fix wrong port when dubbo use multi protocols

dragon-zhang:
see #3056


dragon-zhang:
The repair document will be used instead.

PR #3059, opened 2022-03-18
[ISSUE #3056] change to dubbo multi protocols example

dragon-zhang:
change a dubbo example to multi protocols


dragon-zhang:
see https://github.com/apache/incubator-shenyu-website/pull/490

PR #3078, opened 2022-03-21
[type:refactor] format SQL

midnight2104:
Hi Community:
   This PR is to format the current SQL, following the principle of left-aligning keywords (Uppercase) and right-aligning fields. I would like to make this task a `good first issue` so that new comers can get involved in handling the rest of the SQL file. Do you think it's a good idea?


yu199195:
> maybe you should create an issue

good idea

midnight2104:
yes, here is  issue https://github.com/apache/incubator-shenyu/issues/3088 . pls review .

yu199195 (approved):

PR #3096, opened 2022-03-23
[type:refactor] optimize webclient message writer to cancel business response override.

qicz:
optimize webclient message writer to cancel business response override. like `NettyClientMessageWriter`

Make sure that:

### DIFF
--- a/shenyu-plugin/shenyu-plugin-response/src/main/java/org/apache/shenyu/plugin/response/strategy/WebClientMessageWriter.java
+++ b/shenyu-plugin/shenyu-plugin-response/src/main/java/org/apache/shenyu/plugin/response/strategy/WebClientMessageWriter.java
@@ -50,14 +50,8 @@ public class WebClientMessageWriter implements MessageWriter {
         return chain.execute(exchange).then(Mono.defer(() -> {
             ServerHttpResponse response = exchange.getResponse();
             ClientResponse clientResponse = exchange.getAttribute(Constants.CLIENT_RESPONSE_ATTR);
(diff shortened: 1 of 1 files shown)


shenyu-plugin/shenyu-plugin-response/src/main/java/org/apache/shenyu/plugin/response/strategy/WebClientMessageWriter.java
```diff
@@ -50,16 +49,10 @@
         return chain.execute(exchange).then(Mono.defer(() -> {
             ServerHttpResponse response = exchange.getResponse();
             ClientResponse clientResponse = exchange.getAttribute(Constants.CLIENT_RESPONSE_ATTR);
-            if (Objects.isNull(clientResponse)
```
yu199195, line 53: why remove this

qicz: image
this is the current implementation for `WebClientPlugin` that overrides the business response status code. and `NettyClientMessageWriter` does not cover the business response status and status code. so I change this code with the same logic.

qicz: if the business responds to the 500 or 502 or 504,  the business data will be covered by Shenyu.

yu199195 (approved):

PR #3150, opened 2022-03-30
HTTP registers the client plug-in port occupancy detection

### LINKED ISSUE #3144: [BUG] http registration optimization

### Current Behavior

The same IP port service starts multiple times, and when the problem service is closed, the normal service is taken offline.

### Expected Behavior

The same IP port service is started multiple times, and the first service started can be used normally

### Steps To Reproduce

Register with http.
1. Start the service
2. The same IP port tries to start the service again
3. The service in step 2 stops, the service of this IP port goes offline
4. The service in step 1 is no longer invoked

### Environment

```markdown
ShenYu version(s): 2.4.2
```

### Debug logs

_No response_

### Anything else?

_No response_

jerbo99:
Fixes #3144


yu199195:
pls check ci and modify pr title

yu199195 (approved):

PR #3152, opened 2022-03-30
[Type:fature] add new pagination scheme 

li-keguo:
I have new pagination scheme. it simple,little code,standardized

work list:

- [x] add new pagination scheme .
- [x] used PageHelper paged.
- [x] add app auth paged list API.


yu199195 (changes requested):
next version

li-keguo:
ok, What should I do, need to turn off pr first?

yu199195:
> ok, What should I do, need to turn off pr first?

not close,

Issue #3171, opened 2022-04-01
[Question] performance problem in `MotanProxyService`


dragon-zhang:
### Question

In `org.apache.shenyu.plugin.motan.proxy.MotanProxyService#genericInvoker`, we have a performance problem :
```java
    public Mono<Object> genericInvoker(final String body, final MetaData metaData, final ServerWebExchange exchange) throws ShenyuException {
        Map<String, Map<String, String>> rpcContext = exchange.getAttribute(Constants.GENERAL_CONTEXT);
        Optional.ofNullable(rpcContext).map(context -> context.get(PluginEnum.MOTAN.getName())).ifPresent(context -> {
            context.forEach((k, v) -> RpcContext.getContext().setRpcAttachment(k, v));
        });
        RefererConfig<CommonHandler> reference = ApplicationConfigCache.getInstance().get(metaData.getPath());
        if (Objects.isNull(reference) || StringUtils.isEmpty(reference.getServiceInterface())) {
            ApplicationConfigCache.getInstance().invalidate(metaData.getPath());
            reference = ApplicationConfigCache.getInstance().initRef(metaData);
        }
        CommonHandler commonHandler = reference.getRef();
        ApplicationConfigCache.MotanParamInfo motanParamInfo = ApplicationConfigCache.PARAM_MAP.get(metaData.getMethodName());
        Object[] params;
        if (Objects.isNull(motanParamInfo)) {
            params = new Object[0];
        } else {
            int num = motanParamInfo.getParamTypes().length;
            params = new Object[num];
            Map<String, Object> bodyMap = GsonUtils.getInstance().convertToMap(body);
            for (int i = 0; i < num; i++) {
                params[i] = bodyMap.get(motanParamInfo.getParamNames()[i]).toString();
            }
        }
        ResponseFuture responseFuture;
        //CHECKSTYLE:OFF IllegalCatch
        try {
            responseFuture = (ResponseFuture) commonHandler.asyncCall(metaData.getMethodName(), params, Object.class);
        } catch (Throwable e) {
            LOG.error("Exception caught in MotanProxyService#genericInvoker.", e);
            return null;
        }
        //CHECKSTYLE:ON IllegalCatch
        ResponseFuture finalResponseFuture = responseFuture;
//fixme If there are many time-consuming tasks, the ForkJoinPool with only a few threads will become a performance bottleneck
        CompletableFuture<Object> future = CompletableFuture.supplyAsync(finalResponseFuture::getValue);
[...]

PR #3172, opened 2022-04-01
[ISSUE #3171] change threadPool to CachedThreadPool from ForkJoinPool

dragon-zhang:
see #3171


shenyu-plugin/shenyu-plugin-motan/src/main/java/org/apache/shenyu/plugin/motan/proxy/MotanProxyService.java
```diff
@@ -46,6 +49,8 @@
 
     private static final Logger LOG = LoggerFactory.getLogger(MotanProxyService.class);
 
+    private final ExecutorService threadPool = Executors.newCachedThreadPool(ShenyuThreadFactory.create("shenyu-motan", true));
```
AhahaGe, line 52: for your mail, when we have a large number of requests, cached thread pool may create Integer.MAX_VALUE size threads, may cause oom

dragon-zhang: > 

you can see `org.apache.dubbo.common.threadpool.support.cached.CachedThreadPool`, dubbo also use a `CachedThreadPool`, so it shouldn't be a problem.

In the longer term, this thread pool will be replaced by a shared thread pool(you can see https://github.com/apache/incubator-shenyu/issues/3026).

dragon-zhang: > for your mail, when we have a large number of requests, cached thread pool may create Integer.MAX_VALUE size threads, may cause oom

you can see `org.apache.dubbo.common.threadpool.support.cached.CachedThreadPool`, dubbo also use a `CachedThreadPool`, so it shouldn't be a problem.

In the longer term, this thread pool will be replaced by a shared thread pool(you can see https://github.com/apache/incubator-shenyu/issues/3026).

AhahaGe: how about use ShenyuThreadPoolExecutor here?

dragon-zhang: > Dubbo's cached thread pool support config maxThreads and maxQueue, as it can used to protect dubbo process, so I suggest we config maxThreads and maxQueue like ShenyuThreadPoolExecutor to protect bootstrap process from oom.

`ShenyuThreadPoolExecutor` now supports configuring the maximum number of threads. You can see it on `shenyu-bootstrap/src/main/resources/application.yml`:
```yml
#  sharedPool:
#    enable: true
#    prefix: "shenyu-shared"
#    corePoolSize: 200
#    maximumPoolSize: 2000
#    keepAliveTime: 60000
#    # 1GB
#    maxWorkQueueMemory: 1073741824
```

dragon-zhang: > Dubbo's cached thread pool support config maxThreads and maxQueue, as it can used to protect dubbo process, so I suggest we config maxThreads and maxQueue like ShenyuThreadPoolExecutor to protect bootstrap process from oom.

I planed to complete the transformation of the shared thread pool in the following order:
1. use CachedThreadPool as default thread pool(this PR);
2. support config thread pool type, core threads, max threads and max queues;
3. add support for `ShenyuThreadPoolExecutor`;

Because `2.4.3` will be released soon, only one part will be submitted for risk consideration.

dragon-zhang: > Dubbo's cached thread pool support config maxThreads and maxQueue, as it can used to protect dubbo process, so I suggest we config maxThreads and maxQueue like ShenyuThreadPoolExecutor to protect bootstrap process from oom.

I planed to complete the transformation of the shared thread pool in the following order:
1. use `CachedThreadPool` as the default thread pool (this PR);
2. support config thread pool type, core threads, max threads and max queues;
3. add support for `ShenyuThreadPoolExecutor`;

Because `2.4.3` will be released soon, only one part will be submitted for risk consideration.

AhahaGe: Your plan sounds good. Can we use a compromise method which won't cause oom or reduce it and can also improve performance. How about new a CachedThreadPool with a limited maxPoolSize.

dragon-zhang: > Your plan sounds good. Can we use a compromise method which won't cause oom or reduce it and can also improve performance. How about new a CachedThreadPool with a limited maxPoolSize.

Due to the different hardware configuration of the machine, it is impossible to give a constant value. I think it is a better choice to combine `part 1` and `part 2`, how did you feel ?

AhahaGe: Yes, I agree with u.

yu199195:
this feature maybe merged into 2.5.0，so pls refactor sql to 2.4.3-2.5.0

dragon-zhang:
> this feature maybe merged into 2.5.0，so pls refactor sql to 2.4.3-2.5.0

Hi, xiaoyu, I think I've already done so.
图片

yu199195 (approved):

PR #3226, opened 2022-04-11
[BUG] maven package bin lib not right

siegezhang:
when i run command：
mvn clean package -Dmaven.javadoc.skip=true -B -Drat.skip=true -Djacoco.skip=true -DskipITs -DskipTests -Prelease

after install i  find apache-shenyu-incubating-2.4.2-SNAPSHOT-admin-bin/lib some jar is not right ,like below:
WeChat Image_20220412010551

i think maybe is my fault ,but when i download it from shenyu apache website ,it's also has this problem,bootstarp-bin/lib also has this problem.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -15,8 +15,7 @@
   ~ See the License for the specific language governing permissions and
   ~ limitations under the License.
   -->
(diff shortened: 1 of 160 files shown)


yu199195 (changes requested):
not merged

yu199195:
@siegezhang  can you close this pr, because this pr is error

Issue #3228, opened 2022-04-12
[BUG] unexpected result for loadbalance


ywengineer:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

For Loadbalance. 

1. When i change a upstream's status to false or weight to 0, it still can be chosen.
2. When a selector of DevidePlugin has only one upstream, regardless of it's status and weight, it still can be chosen.

### Expected Behavior

It should not be chosen

### Environment

```markdown
ShenYu version(s):2.4.x
```

PR #3229, opened 2022-04-12
[ISSUE #3228] Fix unexpected result of loadbalance select

ywengineer:

Make sure that:

### DIFF
--- a/shenyu-loadbalancer/src/main/java/org/apache/shenyu/loadbalancer/spi/AbstractLoadBalancer.java
+++ b/shenyu-loadbalancer/src/main/java/org/apache/shenyu/loadbalancer/spi/AbstractLoadBalancer.java
@@ -42,7 +42,7 @@ public abstract class AbstractLoadBalancer implements LoadBalancer {
             return null;
         }
         // available upstream list
-        List<Upstream> available = upstreamList.stream().filter(upstream -> getWeight(upstream) > 0).collect(Collectors.toList());
+        List<Upstream> available = upstreamList.stream().filter(Upstream::isStatus).collect(Collectors.toList());
         if (CollectionUtils.isEmpty(available)) {
             return null;
         }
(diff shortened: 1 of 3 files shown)


qicz:
@ywengineer I toggle ci again. wait a moment.

shenyu-loadbalancer/src/main/java/org/apache/shenyu/loadbalancer/spi/AbstractLoadBalancer.java
```diff
@@ -40,10 +41,15 @@ public Upstream select(final List<Upstream> upstreamList, final String ip) {
         if (CollectionUtils.isEmpty(upstreamList)) {
             return null;
         }
-        if (upstreamList.size() == 1) {
-            return upstreamList.get(0);
+        // available upstream list
+        List<Upstream> available = upstreamList.stream().filter(upstream -> getWeight(upstream) > 0).collect(Collectors.toList());
```
KevinClair, line 45: If there are three upstreams in `DividePlugin`, and their weight is 0, maybe it's error here, right?

ywengineer: Yes

yu199195: @ywengineer  this pr need update？

ywengineer: no, return null if there isn't any available upstream, see line 47.

yu199195: if used HashLoadBalancer？ it not need weight~

yu199195 (changes requested):
not merged

shenyu-loadbalancer/src/main/java/org/apache/shenyu/loadbalancer/spi/AbstractLoadBalancer.java
ywengineer: so, do you mean that we should use the status of upstream to determine it is available or not is better than weight?

ywengineer:
Regardless of any LoadBalancer, closed upstream should not be used. is it right? 
Do you mean that we should use the status of upstream to determine it is available or not is better than weight? am i right?

Issue #3233, opened 2022-04-12
[BUG] <When shenyu.register.props.checked is false, divide plugin selector handle is null>


AhahaGe:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

1. follow divide document to start shenyu-admin, shenyu-bootstrap and run shenyu-examples-http
2. run interface in http-test-api.http, failed can't find upstream
3. see selector data in divide plugin, host, ip:port, protocol is null 

### Expected Behavior

1. follow divide document to start shenyu-admin, shenyu-bootstrap and run shenyu-examples-http
2. run interface in http-test-api.http, can success and return data
3. see selector data in divide plugin, host, ip:port, protocol have right value

### Environment

```markdown
ShenYu version(s):
```

Issue #3223, opened 2022-04-11
which condition may delete the divide plugin handle(the upstream config) in the selector table


justlemontree:
### Question

which condition may delete the divide plugin handle in the selector table
the upstreamCheck enabled is false;

yu199195:
What method of  access are you using?  http register？ or  other register center？ or  youself？

justlemontree:
I have configed like that,&nbsp; I find some condition may delete the upstrem from the front page, such as copy the divide selector and edit it, so the divide handle fileld is empty, so the upsteam in the db is removed.



------------------&nbsp;原始邮件&nbsp;------------------
发件人: ***@***.***&gt;; 
发送时间: 2022年4月12日(星期二) 下午3:01
收件人: ***@***.***&gt;; 
抄送: ***@***.***&gt;; ***@***.***&gt;; 
主题: Re: [apache/incubator-shenyu] which condition may delete the divide plugin handle(the upstream config) in the selector table (Issue #3223)





 
if you can config this in shenyu-admin， it not be check upstream：
 shenyu:   register:     props:       checked: false

yu199195:
so，


copy the divide selector and edit it, so the divide handle fileld is empty？  maybe it a bug in shenyu-dashborad

if you not edit，it be removed？

justlemontree:
if&nbsp; not edit，it will not be removed.




------------------&nbsp;原始邮件&nbsp;------------------
发件人: ***@***.***&gt;; 
发送时间: 2022年4月12日(星期二) 下午3:10
收件人: ***@***.***&gt;; 
抄送: ***@***.***&gt;; ***@***.***&gt;; 
主题: Re: [apache/incubator-shenyu] which condition may delete the divide plugin handle(the upstream config) in the selector table (Issue #3223)





 
so，
 
copy the divide selector and edit it, so the divide handle fileld is empty？  maybe it a bug in shenyu-dashborad
 
if you not edit，it be removed？

AhahaGe:
For latest code, use shenyu-examples-http, handle field is null for divide plugin
<img width="1027" alt="image" src="https://user-images.githubusercontent.com/6198217/162918715-0075cebe-350b-4e6c-9ed1-9e042775b2a9.png">

PR #3234, opened 2022-04-12
[ISSUE #3233] fix when shenyu.register.props.checked = false, divide plugin selector handle is null

AhahaGe:
#3233 
#3223  

Make sure that:


shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/UpstreamCheckService.java
```diff
@@ -144,6 +144,9 @@ public void setup() {
             this.fetchUpstreamData();
             executor = new ScheduledThreadPoolExecutor(1, ShenyuThreadFactory.create("scheduled-upstream-task", false));
             scheduledFuture = executor.scheduleWithFixedDelay(this::scheduled, 10, scheduledTime, TimeUnit.SECONDS);
+        } else {
```
yu199195, line 147: check is false why need new ScheduledThreadPoolExecutor ?

yu199195: hi  thanks for you， i will fix it， because have other error

AhahaGe: welcome

Issue #3275, opened 2022-04-18
[Volunteer] Improve unit test coverage of shenyu-common


JooKS-me:
### Volunteer help

As a branch of #3274 .

At present, the test coverage of shenyu-common module is not high. Please try to make the test coverage of each class reach more than 80% (in the case of ensuring the stability of the test).

If you would like to help, please leave a message under this issue. BTW, you can use `Run 'Test in 'xxx'' With Coverage` in IntelliJ IDEA  to see the approximate coverage.
<img width="1007" alt="image" src="https://user-images.githubusercontent.com/62384022/163818963-37ab02fb-cf6b-4d53-9d8f-192a87222178.png">

PS: If the modified code is too much, it can be divided into several pr submissions.

- [x] org.apache.shenyu.common.cache @selectbook 
- [x] org.apache.shenyu.common.concurrent @Albertsirius 
- [x] org.apache.shenyu.common.config @Nick-fengzl
- [x] org.apache.shenyu.common.dto @luoweijr
- [ ] org.apache.shenyu.common.timer @jun0315
- [ ] org.apache.shenyu.common.utils (maybe a large job) @zhangcly

selectbook:
@JooKS-me Hi, I would like to participate in this and pick the package of  org.apache.shenyu.common.cache

Albertsirius:
@JooKS-me Hello, I would like to participate in this task. Can you assign the package of org.apache.shenyu.common.concurrent to me ?

JooKS-me:
> @JooKS-me Hello, I would like to participate in this task. Can you assign the package of org.apache.shenyu.common.concurrent to me ?

Good luck

luoweijr:
@JooKS-me  Hello, I would like to hava a try and pick the package of org.apache.shenyu.common.dto

jun0315:
@JooKS-me Hi, I would like to participate in this and pick the package of org.apache.shenyu.common.timer

fengzhili12:
Hello, I would like to participate in this task. Can you assign the package of org.apache.shenyu.common.config to me ?

fengzhili12:
@JooKS-me Hello, I would like to participate in this task. Can you assign the package of org.apache.shenyu.common.config to me ?

zhangcly:
@JooKS-me can you assigin `org.apache.shenyu.common.utils` to me? thanks

JooKS-me:
@Nick-fengzl @zhangcly OK :)

[...]

PR #3313, opened 2022-04-24
[ISSUE #3275] Add unit tests for org.apache.shenyu.common.config

Nick-fengzl:
test shenyu-common: improve unit test coverage of org.apache.shenyu.common.config

### DIFF
--- a/shenyu-common/src/test/java/org/apache/shenyu/common/config/ShenyuConfigTest.java
+++ b/shenyu-common/src/test/java/org/apache/shenyu/common/config/ShenyuConfigTest.java
@@ -18,6 +18,7 @@
 package org.apache.shenyu.common.config;
 
 import com.google.common.collect.Lists;
+import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.junit.Test;
 import org.springframework.util.Assert;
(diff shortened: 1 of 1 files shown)


shenyu-common/src/test/java/org/apache/shenyu/common/config/ShenyuConfigTest.java
```diff
@@ -0,0 +1,258 @@
...
+        ShenyuConfig.SharedPool sharedPool = config.getSharedPool();
+        ShenyuConfig.WebsocketConfig websocket = config.getWebsocket();
+        ShenyuConfig.UpstreamCheck upstreamCheck = config.getUpstreamCheck();
+
+        assertNotNullValue(cross, switchConfig, exclude, fallback, file, instance,
+                extPlugin, local, ribbon, metrics, scheduler, sharedPool, websocket, upstreamCheck);
+    }
+
+    private void assertNotNullValue(final Object... objects) {
+        Assert.noNullElements(objects, "val not allow null");
```
dragon-zhang, line 64: Does `java.lang.String` need to be considered here ? Such as `""`. By the way, there may be some negative numbers(such as `-1` for `int` or `long`).

Nick-fengzl: ok, I'll fix it

dragon-zhang: Looking forward to your contribution :)

yu199195 (approved):

Issue #3362, opened 2022-04-29
[Task] Replace zookeeper client to Curator and save data in zk without magic header


hgaol:
### Description

Replace current zk client with [Curator](https://curator.apache.org/), and save data in zk without magic header. The second feature will be incompatible with earlier version, including `register` and `syncing` in zookeeper. More details can refer to #3360 .

### Task List

1. replace zk client with Curator.
2. test different zk server version.
3. save data in zk without magic header.

specific, including below modules.
- register sync between `admin` and `client`.
- register bootstrap instance.
- data sync between `admin` and `bootstrap`.

hgaol:
Please assign it to me, thanks!

PR #3388, opened 2022-05-05
[ISSUR #3362] Replace zookeeper client to Curator and save data in zk without magic header

hgaol:
For #3362 , replace `zkclient` with `curator`, version 5.2.1. ~~To avoid huge PR, I plan to split with 2 PRs. For this one,~~ It mainly done following things.

- added new `zookeeperClient` class in each module and added UT.
- refactored `admin` register server class and added unit test.
- refactored `client` register client class and added UT.
- refactored `register-instance` class and added UT.
- refactored data-sync in admin and bootstrap module and added UT.
- update LICENSE file in `admin-dist` and `bootstrap-dist`.
- removed magic header when saving data in zookeeper server.

Now it works well for registration and data sync.

~~I'll implement the *data-sync* in `admin` and `bootstrap` in the next PR.~~

BTW, for some UT, I replace mocking `zkClient` with adding `TestingServer` provided by `curator-test`. It's better to help to verify the behaviors like operating with real zk server.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -109,7 +109,7 @@
         <curator-test.version>4.0.1</curator-test.version>
         <curator.version>5.2.1</curator.version>
         <wiremock.version>2.18.0</wiremock.version>
-        <zookeeper.version>3.5.6</zookeeper.version>
+        <zookeeper.version>3.6.3</zookeeper.version>
         <zkclient.version>0.10</zkclient.version>
         <shiro.version>1.8.0</shiro.version>
         <jwt.version>3.12.0</jwt.version>
--- a/pom.xml
+++ b/pom.xml
@@ -15,7 +15,8 @@
   ~ See the License for the specific language governing permissions and
   ~ limitations under the License.
   -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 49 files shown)


hgaol:
image

pom.xml
```diff
@@ -412,6 +412,27 @@
                 <version>${mail.version}</version>
             </dependency>
 
+            <dependency>
+                <groupId>org.apache.curator</groupId>
+                <artifactId>curator-framework</artifactId>
```
yu199195, line 417: add this curator in license file

shenyu-common/src/main/java/org/apache/shenyu/common/client/zookeeper/ZookeeperClient.java
```diff
@@ -0,0 +1,246 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class ZookeeperClient {
```
yu199195, line 41: why add this  in common？

hgaol: I found there are many common code snippets. To avoid dup code, I created this `ZookeeperClient` and put it in `common`.
Please let me know where it should be, and I'll move it to another module.

pom.xml
hgaol: What's the file name? I found a `LICENSE` file in project root folder but seems not it.

hgaol: resolved.

hgaol:
I've tested in below steps, it works well. Please also help to check and let me know if any problem.

0. modify configurations using zookeeper as register in `admin` and `http-example`, using zk as data sync method in `admin` and `bootstrap`. Also tested deprecated class `HttpServiceDiscovery` via adding environment variables.
1. start admin.
2. start bootstrap
3. start http example
4. check data in zk and admin portal.
5. call api to bootstrap and return value as expected.

yu199195 (approved):

Issue #3485, opened 2022-06-05
[Task] simplify the use of client annotations with @AliasFor


dragon-zhang:
### Description

For example, below is how we use `@ShenyuXXXXClient` now:
```java
@RestController
@RequestMapping("/order")
@ShenyuSpringCloudClient(path = "/order")
public class OrderController {

    /**
     * Save order dto.
     *
     * @param orderDTO the order dto
     * @return the order dto
     */
    @PostMapping("/save")
    @ShenyuSpringCloudClient(path = "/save")
    public OrderDTO save(@RequestBody final OrderDTO orderDTO) {
        orderDTO.setName("hello world spring cloud save order");
        return orderDTO;
    }

}
```
After use `@AliasFor` with `@ShenyuSpringCloudClient` like below:
```java
@Retention(RetentionPolicy.RUNTIME)
@Target({ElementType.TYPE, ElementType.METHOD})
public @interface ShenyuSpringCloudClient {

    /**
     * Path string.
     *
     * @return the string
     */
    @AliasFor(attribute = "path")
    String value() default "";

    /**
     * Path string.
     *
     * @return the string
     */
    @AliasFor(attribute = "value")
    String path() default "";

    /**
     * Rule name string.
     *
     * @return the string
     */
    String ruleName() default "";

    /**
     * Desc string.
     *
     * @return String string
     */
    String desc() default "";

    /**
     * Enabled boolean.
     *
     * @return the boolean
     */
    boolean enabled() default true;
}
```
We can simplify client annotation usage to:
```java
@RestController
@RequestMapping("/order")
@ShenyuSpringCloudClient("/order")
public class OrderController {
    
    /**
     * Save order dto.
     *
     * @param orderDTO the order dto
     * @return the order dto
     */
    @PostMapping("/save")
    @ShenyuSpringCloudClient("/save")
    public OrderDTO save(@RequestBody final OrderDTO orderDTO) {
        orderDTO.setName("hello world spring cloud save order");
        return orderDTO;
    }
    
}
```
You can refer to the implementation scheme of `shenyu-client-springcloud`, don't forget to use `org.springframework.core.annotation.AnnotatedElementUtils#findMergedAnnotation`.

And if you have some great ideas, please leave a message below, let's discuss it.

### Task List

1. - [x] simplify @ShenyuDubboClient in `alibaba dubbo`; @sunshujie1990
2. - [x] simplify @ShenyuDubboClient in `apache dubbo`; @sunshujie1990
[...]

PR #3527, opened 2022-06-11
simplify ShenyuDubboClient

weihubeats:
https://github.com/apache/incubator-shenyu/issues/3485


dragon-zhang:
@weihubeats Hi, there is no need to change so many examples. Just change one example of Apache Dubbo and Alibaba Dubbo to ensure correct works.

dragon-zhang:
please change the title to `[ISSUE #3485] simplify ShenyuDubboClient`, and check CI/CD workflow results.

dragon-zhang (approved):

PR #3596, opened 2022-06-21
[type:Integration Test] Add new test for rewrite-plugin 

erdengk:
This pr is to add a new test case for the rewrite plugin, which proves that rewrite is normal from another perspective.

- [x] Local test passed


erdengk:
@JooKS-me  please review this pr,thanks~

JooKS-me (approved):

PR #3672, opened 2022-07-06
[type:Integration Test] add integrated-test of logging-rocketmq-plugin

erdengk:
This pr is used to test whether rocketmq is integrated into shenyu.
The test case steps are as follows:
1. Add a message to rocketmq
2. Consume this information

If the consumption is successful, it means that rocketmq is integrated into shenyu. If the consumption fails, an exception will be thrown, indicating that there is a problem with rocketmq.

### DIFF
--- a/.github/workflows/integrated-test.yml
+++ b/.github/workflows/integrated-test.yml
@@ -20,6 +20,7 @@ on:
   push:
     branches:
       - master
+      - addRocketMqTest
 
 jobs:
   build:
--- a/.github/workflows/integrated-test.yml
+++ b/.github/workflows/integrated-test.yml
@@ -20,8 +20,6 @@ on:
   push:
     branches:
       - master
-      - addRocketMqTest
-
 jobs:
   build:
     strategy:
--- a/.github/workflows/integrated-test.yml
+++ b/.github/workflows/integrated-test.yml
@@ -20,6 +20,7 @@ on:
   push:
     branches:
       - master
+
 jobs:
   build:
     strategy:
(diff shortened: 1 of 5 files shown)


erdengk:
@JooKS-me  please review this pr,thanks~

shenyu-integrated-test/shenyu-integrated-test-http/docker-compose.yml
```diff
@@ -128,6 +128,47 @@ services:
...
+    volumes:
+      - ./logs:/root/logs
+    command: sh mqnamesrv
+    networks:
+      shenyu:
+        aliases:
+          -  shenyu-rocketmq
+
+  shenyu-rmqbroker:
+    image: rocketmqinc/rocketmq:4.3.2
```
JooKS-me, line 151: Why use `rocketmqinc/rocketmq`, is it different from `apache/rocketmq` ?

.github/workflows/integrated-test.yml
```diff
@@ -20,7 +20,6 @@ on:
   push:
     branches:
       - master
-
```
JooKS-me, line 23: Don't delete this line

JooKS-me (changes requested):

shenyu-integrated-test/shenyu-integrated-test-http/docker-compose.yml
erdengk: There is no difference between them, apache/rocketmq is officially released by apache. rocketmqinc/rocketmq is a private release, maybe apache/rocketmq is better here.

JooKS-me: Yep, let's change it to apache/rocketmq.

erdengk: ok

JooKS-me (approved):
Good job

PR #3673, opened 2022-07-06
Plugin modification concurrency issues #3620

### LINKED ISSUE #3620: [Task] <ShenyuWebHandler The modification of the plugin does not take into account concurrency issues and the current requests being processed>
###  ShenyuWebHandler Description

The gateway processes requests concurrently and has requests in process, but the processor's plug-in modifications are really single-threaded.

### class: org.apache.shenyu.web.handler.ShenyuWebHandler

Examples are as follows.

## Cannot gracefully process a request in progress
image

image

image

image

### Task List

1. Adopting a Copy-On-Write idea to gracefully update the plugin.

hutaishi:
Fixes #3620 

Copy a new list of plugins without affecting the current request being processed.

### DIFF
--- a/shenyu-web/src/main/java/org/apache/shenyu/web/handler/ShenyuWebHandler.java
+++ b/shenyu-web/src/main/java/org/apache/shenyu/web/handler/ShenyuWebHandler.java
@@ -104,7 +104,7 @@ public final class ShenyuWebHandler implements WebHandler, ApplicationListener<P
      *
      * @param extPlugins the ext plugins
      */
-    public synchronized void putExtPlugins(final List<ShenyuPlugin> extPlugins) {
+    public void putExtPlugins(final List<ShenyuPlugin> extPlugins) {
         if (CollectionUtils.isEmpty(extPlugins)) {
             return;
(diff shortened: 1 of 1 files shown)


shenyu-web/src/main/java/org/apache/shenyu/web/handler/ShenyuWebHandler.java
```diff
@@ -105,7 +104,7 @@ public Mono<Void> handle(@NonNull final ServerWebExchange exchange) {
      *
      * @param extPlugins the ext plugins
      */
-    public void putExtPlugins(final List<ShenyuPlugin> extPlugins) {
+    public synchronized void putExtPlugins(final List<ShenyuPlugin> extPlugins) {
```
yu199195, line 107: used volatile， maybe not used synchronized？

hutaishi: This method can be called without synchronized because it will not be called concurrently, I remove this

Issue #2303, opened 2021-11-05
[BUG] divide plugin  webclient throw exception: The connection observed an error,io.netty.handler.timeout.ReadTimeoutException: null


dongxl123:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

2021-11-04 10:58:29 [XNIO-1 I/O-3] INFO  org.apache.shenyu.plugin.httpclient.WebClientPlugin - The request urlPath is /api/v1/query?query=rate%28recv_packet_total%7Bsource%21%3D%22cffex%22%2Ctarget%3D~%22cffex.%2A%22%7D%5B5m%5D

%29&time=2021-11-04T02%3A58%3A28.564680692Z, retryTimes is 3
2021-11-04 10:58:32 [reactor-http-epoll-3] WARN  reactor.netty.http.client.HttpClientConnect - [id: 0xa0255f85, L:/10.201.220.93:34256 - R:product.thanos.10jqka.com.cn/172.20.208.115:80] The connection observed an error
io.netty.handler.timeout.ReadTimeoutException: null
2021-11-04 10:58:32 [reactor-http-epoll-3] ERROR org.apache.shenyu.plugin.httpclient.WebClientPlugin - null
io.netty.handler.timeout.ReadTimeoutException: null
2021-11-04 10:58:32 [reactor-http-epoll-3] ERROR org.apache.shenyu.web.handler.GlobalErrorHandler - [e9ba400] Resolved [ReadTimeoutException: null] for HTTP GET /prometheus/api/v1/query
2021-11-04 10:58:32 [reactor-http-epoll-3] ERROR org.springframework.boot.autoconfigure.web.reactive.error.AbstractErrorWebExceptionHandler - [e9ba400]  500 Server Errorfor HTTP GET "/prometheus/api/v1/query?query=rate

%28recv_packet_total%7Bsource%21%3D%22cffex%22%2Ctarget%3D~%22cffex.%2A%22%7D%5B5m%5D%29&time=2021-11-04T02%3A58%3A28.564680692Z"
io.netty.handler.timeout.ReadTimeoutException: null
2021-11-04 10:58:32 [reactor-http-epoll-3] ERROR reactor.core.publisher.Operators - Operator called default onErrorDropped
java.lang.NullPointerException: null
[...]

PR #3682, opened 2022-07-08
#2303 Prevent redundant timeout error by overriding the fallback

Rubenmp:
Reactor netty pollutes stderr by default when timeout exceptions are thrown, showing errors like `io.netty.handler.timeout.ReadTimeoutException: null`
There are two ways to handle this:
- First, pass a custom fallback for each flux timeout to override the default one:
```
.timeout(timeout) // Before
.timeout(timeout, Mono.error(new TimeoutException("Response took longer than timeout: " + timeout)) // After
```

- Second, disabling all netty logs with this configuration line (and manually logging web flux errors with methods like onErrorContinue, ...):
```
logging.level.reactor.netty.http.client.HttpClient=off
```

I choose the first one, it is done like that in other parts of the code. I checked all usages of _WebClient_ and all the timeouts.
The only place where timeouts are not explicitly handled is the method:
```
protected Mono<ClientResponse> doRequest(final ServerWebExchange exchange, final String httpMethod, final URI uri, final HttpHeaders httpHeaders, final Flux<DataBuffer> body);
```
but I checked all its usages and they attach safe timeouts with error handling.

The removal of this error log is justified because:
- The timeout error is still handled later with `doOnError`
- Other parts of the code handle timeouts similarly.

Make sure that:


yu199195:
@Rubenmp thanks for you contributor，looks good to me.

yu199195 (approved):

PR #3722, opened 2022-07-16
[type: feat] update docker-publish workflow, make it only push image when release

skyleaworlder:
Make sure that:

### Now

~[docker-publish workflow](https://github.com/apache/incubator-shenyu/actions/workflows/docker-publish.yml) run every time new commits merged into master.~ And this workflow can only push image to "ghcr.io", which means that images need pushed to dockerhub manually if there is a new release.

~Additionally, images on ghcr.io use "github.sha" as label.~

### This PR

Update as below:

* Add new workflow to publish image to dockerhub.
* Use "release published" as trigger event.
* Use "github.ref_name" instead of "github.sha" in new workflow. ref: [get-the-current-pushed-tag-in-github-actions](https://stackoverflow.com/questions/58177786/get-the-current-pushed-tag-in-github-actions).
* ~Still use "build-push-action", push images to multi-registries. ref: [build-push-action : push-multi-registries](https://github.com/docker/build-push-action/blob/master/docs/advanced/push-multi-registries.md)~.
* Use filter "!github.event.release.prerelease && !github.event.release.draft" to avoid redundant or unnecessary run. ref: [RESTful API of releases](https://docs.github.com/en/rest/releases/releases).

### Test

I only create a [playground repo](https://github.com/skyleaworlder/docker-plugin-example), and test in it.

I use [this workflow yaml](https://github.com/skyleaworlder/docker-plugin-example/actions/runs/2682291485/workflow), append a tag (v0.0.1-alpha.8) and [release](https://github.com/skyleaworlder/docker-plugin-example/releases/tag/v0.0.1-alpha.8).

You can check [job execute summary](https://github.com/skyleaworlder/docker-plugin-example/runs/7371362589?check_suite_focus=true), [image pushed to ghcr.io](https://github.com/skyleaworlder/docker-plugin-example/pkgs/container/normal-dockerfile):

image

and [image pushed to dockerhub](https://hub.docker.com/repository/docker/skyleaworlder/normal-dockerfile):

image

~### TODO~

~"docker/login-action" need dockerhub username and password:~

```yaml
      - name: Login to DockerHub
        uses: docker/login-action@v2
        with:
          username: ${{ secrets.DOCKERHUB_USERNAME }}
          password: ${{ secrets.DOCKERHUB_PASSWORD }}
```

~Here I use "DOCKERHUB_USERNAME" and "DOCKERHUB_PASSWORD", so it need set secrets in `settings | Security | Secrets | Actions`.~

I use "DOCKERHUB_USER" and "DOCKERHUB_TOKEN" as [cwiki - Github Secrets and Tokens](https://cwiki.apache.org/confluence/display/INFRA/Github+Secrets+and+Tokens) said.

### Need help

Actually, I'm not very confident whether my PR would work. So I would be very happy if reviewers can help to check.

### DIFF
--- a/.github/workflows/docker-publish-dockerhub.yml
+++ b/.github/workflows/docker-publish-dockerhub.yml
@@ -19,7 +19,7 @@ on:
   release:
     # pre-release or draft status don't trigger this job.
     # this job called if and only if pre-release -> release or release directly.
-    type: ["released"]
+    types: ["released"]
 
 env:
   REPOSITORY: ${{ github.repository }}
--- a/.github/workflows/docker-publish-dockerhub.yml
+++ b/.github/workflows/docker-publish-dockerhub.yml
@@ -13,20 +13,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-name: docker-publish-ghcr
+name: docker-publish-dockerhub
 
 on:
-  push:
-    branches: [ "master" ]
-    tags: [ 'v*.*.*' ]
+  release:
+    type: ["published"]
 
 env:
-  REGISTRY: ghcr.io
   REPOSITORY: ${{ github.repository }}
-  TAG: ${{ github.sha }}
+  TAG: ${{ github.ref_name }}
 
 jobs:
-  build:
+  dockerhub:
+    # pre-release or draft status don't trigger this job.
+    # this job called if and only if pre-release -> release or release directly.
+    if: "!github.event.release.prerelease && !github.event.release.draft"
     runs-on: ubuntu-latest
     permissions:
       contents: read
@@ -41,13 +42,11 @@ jobs:
       - name: Set Skip Env Var
         uses: ./.github/actions/skip-ci
 
-      - name: Login to GitHub Container Registry
+      - name: Login to DockerHub
         uses: docker/login-action@v2
-        if: env.SKIP_CI != 'true'
         with:
-          registry: ${{ env.REGISTRY }}
-          username: ${{ github.actor }}
-          password: ${{ secrets.GITHUB_TOKEN }}
+          username: ${{ secrets.DOCKERHUB_USER }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v2
@@ -79,7 +78,7 @@ jobs:
         if: env.SKIP_CI != 'true'
         run: ./mvnw -B clean -Prelease -Dmaven.javadoc.skip=true -B -Drat.skip=true -Djacoco.skip=true -DskipITs -DskipTests package
 
-      - name: Build and push (admin) (ghcr.io)
+      - name: Build and push (admin) (dockerhub)
         uses: docker/build-push-action@v3
         if: env.SKIP_CI != 'true'
         with:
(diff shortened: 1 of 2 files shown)


yu199195:
looks good

.github/workflows/docker-publish.yml
```diff
@@ -16,17 +16,19 @@
 name: docker-publish
 
 on:
-  push:
-    branches: [ "master" ]
-    tags: [ 'v*.*.*' ]
+  release:
```
yu199195, line 19: why is release？

JooKS-me:
Hi @skyleaworlder, IMO, we should publish to `ghcr.io` on every merge, and to `docker hub` on every release.

.github/workflows/docker-publish.yml
skyleaworlder: I found images are only pushed to dockerhub when release. So I'm wondering if I need to do the same when it comes to ghcr.io. Then I replace "push branches master" with "release" here.

If "push branches master" is intensinally-designed, I'll fix it.

skyleaworlder:
> Hi @skyleaworlder, IMO, we should publish to `ghcr.io` on every merge, and to `docker hub` on every release.

Oh, I see. I will revert it later.

.github/workflows/docker-publish.yml
yu199195: When releasing a version, we push the docker image, you can refer to it here: https://shenyu.apache.org/community/release-guide#finish-publishing

Not used github action.

zhc00:
hi @skyleaworlder 
these are daily docker images. they are for testing in sub repositories or others.
so i think we keep this strategy. ghrc will sweep the expired images.

.github/workflows/docker-publish.yml
JooKS-me: @yu199195 We can use actions to help us build and publish images that support multiple platforms, such as: arm64, amd64, which can make it easier for us to publish, and currently some other apache projects do the same.

And apache infra seems to provide the docker hub account and password in the GitHub settings: https://cwiki.apache.org/confluence/display/INFRA/Github+Secrets+and+Tokens

zhc00:
yes. i agree with @JooKS-me .
and i prefer to create new flow to do release flow.

.github/workflows/docker-publish.yml
skyleaworlder: @yu199195 @JooKS-me I restore the former docker-publish workflow, and create a new one to publish images to dockerhub in recent new commit.

.github/workflows/docker-publish-dockerhub.yml
```diff
@@ -0,0 +1,103 @@
...
+      packages: write
+      id-token: write
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: true
+
+      - name: Set Skip Env Var
+        uses: ./.github/actions/skip-ci
```
JooKS-me, line 43: We need to delete SKIP_CI, or it will skip the job when release.

JooKS-me (changes requested):

skyleaworlder:
Now I use "type released" instead of "published" as trigger event of docker-publish-dockerhub.yml. ("published" would make workflow run very often, because "published" event also contains concepts of several other types) ref: https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-object-40

I have tested this feature about GitHub Actions in my personal [repo](https://github.com/skyleaworlder/docker-plugin-example/actions). ref: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release

image

.github/workflows/docker-publish-dockerhub.yml
skyleaworlder: Thanks! I fix it.

JooKS-me (dismissed):
LGTM

JooKS-me:
Synchronize the progress, I created [an issue](https://issues.apache.org/jira/browse/INFRA-23509) to infra, and they gave me permission. I have now added two secrets, DOCKERHUB_TOKEN and DOCKERHUB_USER.

I have test the docker-hub account in https://github.com/apache/shenyu/pull/3753

JooKS-me (approved):

PR #3725, opened 2022-07-18
[type: refactor] support cors regex origin white list.

ningminglong:
…originRegex

Make sure that:


shenyu-common/src/main/java/org/apache/shenyu/common/config/ShenyuConfig.java
```diff
@@ -977,6 +977,9 @@ public static class CrossFilterConfig {
 
         private boolean allowCredentials;
 
+        private String originRegex;
```
qicz, line 980: move to `allowOrigin`

qicz (changes requested):
move `originRegex` to `allowOrigin`

Issue #3773, opened 2022-07-30
[Task] use `#exchangeToMono` replace `#exchange`


dragon-zhang:
### Description

use `org.springframework.web.reactive.function.client.WebClient.RequestHeadersSpec#exchangeToMono` replace `org.springframework.web.reactive.function.client.WebClient.RequestHeadersSpec#exchange`, but make sure https://github.com/apache/shenyu/issues/3686 and https://github.com/apache/shenyu/issues/3687 not happens.

### Task List

1. - [ ] use `#exchangeToMono` replace `#exchange` in `WebClientPlugin` @spicydev
2. - [x] add integrated test for https://github.com/apache/shenyu/issues/3158 @moremind 
3. - [x] add integrated test for https://github.com/apache/shenyu/issues/3687 @moremind

moremind:
maybe exchangeToMono?

spicydev:
Hi @moremind / @dragon-zhang , 
I have created below pr with changes to completely implement new WebClient request/response spec using retrieve() method instead of exchange().
https://github.com/apache/shenyu/pull/3775

I need some help in validating these client changes before we can merge the pr.

moremind:
hi,@spicydev, exchangeToMono or exchangeToFlux process request more than retrive, and shenyu process http response by response plugin. So, I think you must think a lot.

dragon-zhang:
> maybe exchangeToMono?

I remember there seemed to be a bug before. It was said that `Mono` had a size limit, and changed to `Flux` as final.

moremind:
> > maybe exchangeToMono?
> 
> I remember there seemed to be a bug before. It was said that `Mono` had a size limit, and changed to `Flux` as final.

it can change

moremind:
> > > maybe exchangeToMono?
> > 
> > 
> > I remember there seemed to be a bug before. It was said that `Mono` had a size limit, and changed to `Flux` as final.
> 
> it can change

maybe change body to byte[], and write to DataBuffer

moremind:
about lose data, maybe think `DataBufferUtils.join()`

dragon-zhang:
> about lose data, maybe use `DataBufferUtils.join()`

got, I've seen it in https://github.com/apache/shenyu/issues/3158

moremind:
i have writed use this

![image](https://user-images.githubusercontent.com/20184263/182013151-ff289ce5-e16f-44ab-8098-7bd9cf8e8229.png)

[...]

PR #3776, opened 2022-07-31
[ISSUE #3773] WebClientPlugin switched exchange() to retrieve() method

spicydev:
Make sure that:


shenyu-plugin/shenyu-plugin-httpclient/src/main/java/org/apache/shenyu/plugin/httpclient/WebClientPlugin.java
```diff
@@ -49,27 +50,29 @@ public WebClientPlugin(final WebClient webClient) {
     }
 
     @Override
-    protected Mono<ClientResponse> doRequest(final ServerWebExchange exchange, final String httpMethod, final URI uri,
+    protected Mono<ResponseEntity<Void>> doRequest(final ServerWebExchange exchange, final String httpMethod, final URI uri,
```
yu199195, line 53: if modify this， may be have modify this : WebClientMessageWriter

moremind:
maybe use exchangeToMono or exchageToFlux

shenyu-plugin/shenyu-plugin-httpclient/src/main/java/org/apache/shenyu/plugin/httpclient/WebClientPlugin.java
```diff
@@ -49,27 +50,29 @@ public WebClientPlugin(final WebClient webClient) {
...
-                    }
+                .retrieve()
+                .onStatus(HttpStatus::is2xxSuccessful, res -> {
+                    exchange.getAttributes().put(Constants.CLIENT_RESPONSE_RESULT_TYPE, ResultEnum.SUCCESS.getName());
                     exchange.getResponse().setStatusCode(res.statusCode());
                     exchange.getAttributes().put(Constants.CLIENT_RESPONSE_ATTR, res);
-                });
+                    return Mono.empty();
+                })
+                .onStatus(HttpStatus::isError, res -> {
```
dragon-zhang, line 69: use `status -> !status.is2xxSuccessful()` replace `HttpStatus::isError`.

dragon-zhang:
@spicydev ci/cd failed, please check https://github.com/apache/shenyu/runs/7596239922?check_suite_focus=true

shenyu-plugin/shenyu-plugin-httpclient/src/main/java/org/apache/shenyu/plugin/httpclient/WebClientPlugin.java
spicydev: Agree @yu199195 , I was missing this piece. we can modify the WebClientMessageWriter to use newer WebClient.ResponseSpec instead of ClientResponse to process data. what are your thoughts on the above.

shenyu-plugin/shenyu-plugin-httpclient/src/main/java/org/apache/shenyu/plugin/httpclient/WebClientPlugin.java
```diff
@@ -49,27 +50,29 @@ public WebClientPlugin(final WebClient webClient) {
...
                 .headers(headers -> headers.addAll(httpHeaders))
                 .body(BodyInserters.fromDataBuffers(body))
-                .exchange()
-                .doOnSuccess(res -> {
-                    if (res.statusCode().is2xxSuccessful()) {
-                        exchange.getAttributes().put(Constants.CLIENT_RESPONSE_RESULT_TYPE, ResultEnum.SUCCESS.getName());
-                    } else {
-                        exchange.getAttributes().put(Constants.CLIENT_RESPONSE_RESULT_TYPE, ResultEnum.ERROR.getName());
-                    }
+                .retrieve()
```
dragon-zhang, line 62: After discussion with @moremind , the performance of `#retrieve` is far inferior to `#exchangeToMono`, so please use `#exchangeToMono`. You can test the performance gap between `#retrieve` and `#exchangeToMono`.

spicydev: Hi @dragon-zhang , the performance of both exchange() & retrieve() is same, but retrieve restricts us to consume payload with it's ResponseSpec interface.

`@Override
 public ResponseSpec retrieve() {
  return new DefaultResponseSpec(exchange(), this::createRequest);
}`

Perhaps the question here is why should WebClientMessageWriter has to handle the responsibility to release client response body which is clearly causing the memory leak in case of null pointers. That responsibility we can try to delegate back to the WebClientPlugin.

dragon-zhang: You can continue to complete it according to your idea, but before the PR get merged, I will do a simple pressure test on your PR.

dragon-zhang:
please update changes from `master`.

dragon-zhang:
see https://github.com/apache/shenyu/pull/3792 ,sorry please try another task @spicydev

dragon-zhang (changes requested):
not merge

PR #3846, opened 2022-08-13
[example]Add swagger to the sample project to test the relevant functionality of the API documentation.

lianjunwei:
Add swagger to the sample project to test the relevant functionality of the API documentation.


dengliming (commented):
IMO, This doesn't seem to make sense.

Issue #3883, opened 2022-08-28
[Volunteer] When a plugin changes, it should be initialized


shihuili1218:
### Volunteer help

When plugin parameters change, the common logic should be initialized and should not be computed each time a request is processed.

e.g. When using LoadBalance, Divide、Dubbo、Http plugin needs to calculate the weight for each request. To calculate the weight, you should only need to calculate it once when the configuration changes.

I don't know much about Shenyu yet, and I don't know if there is similar situation in other places. If this optimization is necessary and there are others that need to be optimized, please supplement, I can finish it together.

My thoughts are as follows:
In `CommonPluginDataSubscriber#onSelectorSubscribe`, sending the update or delete events, performs initialization logic in the corresponding plugin.

yu199195:
Yes，I think it is necessary.  can you  provider this pr ?

shihuili1218:
> 

of course

shihuili1218:
This issue has been completed, please review
https://github.com/apache/shenyu/pull/3894

KevinClair:
> This issue has been completed, please review #3894

TotalWeight is dynamic, see org.apache.shenyu.loadbalancer.entity.Upstream#warmup

shihuili1218:
> > This issue has been completed, please review #3894
> 
> TotalWeight is dynamic, see org.apache.shenyu.loadbalancer.entity.Upstream#warmup

Oops, I see. the total weight cannot be calculated when the parameter is changed, so I will close this pr and open it again if necessary

KevinClair:
Thanks a lot, if you have more question, you can reopen this issue.

PR #3894, opened 2022-08-30
When a plugin changes, it should be initialized 

shihuili1218:
this is issue's implation. https://github.com/apache/shenyu/issues/3883

Make sure that:


shihuili1218:
TotalWeight is dynamic, the total weight cannot be calculated when the parameter is changed

PR #3897, opened 2022-08-30
polish ci job filter

dengliming:
Make sure that:


.github/workflows/ci.yml
```diff
@@ -39,32 +39,36 @@ jobs:
...
     steps:
       - name: Support longpaths
         if: ${{ matrix.os == 'windows-latest'}}
         run: git config --system core.longpaths true
       - uses: actions/checkout@v2
         with:
           submodules: true
-      - name: Set Skip Env Var
-        uses: ./.github/actions/skip-ci
+      - uses: dorny/paths-filter@v2
```
JooKS-me, line 50: It may be wrong to use third-party actions directly in the apache repository. We can add git submodules in `./.github/actions`. For example: https://github.com/apache/shenyu-helm-chart/tree/main/.github/actions

dengliming: Is there an official explanation?

JooKS-me: image
This is the error we encountered before. But it looks ok in your pr's ci 🤔, maybe we can try it first.

JooKS-me (approved):

.github/workflows/ci.yml
dengliming: em. This is a special case. If the action is generic and multiple projects are using it, we can try to request `Apache ` for permission to use it. : )

PR #3924, opened 2022-09-07
[type: build] Support user-defined jvm param in startup scripts

### LINKED ISSUE #3916: [Question] Why is shenyu-bootstrap jvm heap size fixed?
### Question

Below is a part of the shell file executed when using shenyu-bootstrap 2.5.0 docker image. (line 34 of start.sh)

`JAVA_OPTS=" -server -Xmx2g -Xms2g -Xmn1g -Xss512k -XX:+DisableExplicitGC   -XX:LargePageSizeInBytes=128m"`

I think this should be non-static and should be selectable via environment variables.

I'd like to know if this setting is the best setting for most situations or if there is a way to set it up separately.

Full file can be found here: [start.sh](https://github.com/apache/shenyu/blob/v2.5.0/shenyu-dist/shenyu-bootstrap-dist/src/main/resources/bin/start.sh)

JooKS-me:
Resolve #3916

Use environment variables to pass JVM parameters, corresponding to `SHENYU_ADMIN_JAVA_OPTS` in `shenyu-admin` and `SHENYU_BOOTSTRAP_JAVA_OPTS` in `shenyu-bootstrap`.

In docker, we just need to add `-e` to set environment variables.

In docker-compose, we just need to add `environment` config.

image

image

image

image

### DIFF
--- a/shenyu-dist/shenyu-bootstrap-dist/src/main/resources/bin/start.sh
+++ b/shenyu-dist/shenyu-bootstrap-dist/src/main/resources/bin/start.sh
@@ -31,7 +31,7 @@ LOG_FILES=${LOGS_DIR}/shenyu-bootstrap.log
 EXT_LIB=${DEPLOY_DIR}/ext-lib
 
 CLASS_PATH=.:${DEPLOY_DIR}/conf:${DEPLOY_DIR}/lib/*:${EXT_LIB}/*
-if [ -z "${SHENYU_BOOTSTRAP_JAVA_OPTS}" ]; then
+if [ -z "${BOOT_JVM}" ]; then
     JAVA_OPTS=" -server -Xmx2g -Xms2g -Xmn1g -Xss512k -XX:+DisableExplicitGC   -XX:LargePageSizeInBytes=128m"
     version=`java -version 2>&1 | sed '1!d' | sed -e 's/"//g' | awk '{print $3}'`
     echo "current jdk version:${version}"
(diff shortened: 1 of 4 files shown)


shenyu-dist/shenyu-bootstrap-dist/src/main/resources/bin/start.sh
```diff
@@ -31,16 +31,21 @@ LOG_FILES=${LOGS_DIR}/shenyu-bootstrap.log
...
+    if [[ "$version" =~ "1.8" ]];then
+        JAVA_OPTS="${JAVA_OPTS} -XX:+UseFastAccessorMethods  -XX:+UseConcMarkSweepGC -XX:+CMSParallelRemarkEnabled -XX:+UseCMSInitiatingOccupancyOnly  -XX:CMSInitiatingOccupancyFraction=70"
+    elif [[ "$version" =~ "11" ]];then
+        JAVA_OPTS="${JAVA_OPTS}"
+    elif [[ "$version" =~ "17" ]];then
+        JAVA_OPTS="${JAVA_OPTS}"
+    fi
+    echo "Use default jvm param: $JAVA_OPTS"
+else
+    JAVA_OPTS=${SHENYU_BOOTSTRAP_JAVA_OPTS}
```
yu199195, line 47: where from  ${SHENYU_BOOTSTRAP_JAVA_OPTS}？

JooKS-me: os environment variables

yu199195 (approved):

Issue #3861, opened 2022-08-18
[Task] desensitization of main fields in logs


moremind:
### Description

desensitization of main fields in logs,for example:

requestbody:{"idCard": "361222xxxxxxxxxxx00", "name":"赵x明"},this rule can config.

![image](https://user-images.githubusercontent.com/20184263/185434790-04532f80-37e4-4116-b231-e0ecae95aa41.png)

### Task List

1. [ ] desensitization of main fields in logs for all log plugin

yunlongn:
pls assign to me

moremind:
> I want to try,please assign me.

ok, I have assigned, if you have some questions, please contact me

wanghuan0606:
> 

I will contact you promptly if I have any question.

moremind:
> > 
> 
> I will contact you promptly if I have any question.

my wechat is yijia1949

PR #3927, opened 2022-09-07
[Task] desensitization of main fields in logs #3861

wanghuan0606:
**[Task] desensitization of main fields in logs #3861**

**In this pr, I mask logging data according keyword like uri,name,id et.**
**I add two new plugin handler,and three dictionary value like this:**
截屏2022-09-08 02 00 22
截屏2022-09-08 02 00 41
截屏2022-09-08 02 00 48

**User can add keyword and choose maskType when they make the logging rules,like this**:
截屏2022-09-08 02 00 03

**Now,I provide two mask types include charReplace and MD5,then user can extend their own mask types.Only need implements DataMaskInterface and create new dictionary that dictionary type is "maskType" and dictionary value is their function like this:**
截屏2022-09-08 02 51 47

截屏2022-09-08 02 46 25


moremind:
pls use two input for keyword and condition

Issue #3932, opened 2022-09-11
[Task] refactor provider client sdks


dragon-zhang:
### Description

Now there are some duplicate codes in the client SDK. This issue aim to solve the problem.

There are some useful information in https://github.com/apache/shenyu/pull/3935 and https://github.com/apache/shenyu/pull/3931 .

### Task List

1. - [x] refactor shenyu-client-alibaba-dubbo; @847850277
2. - [x] refactor shenyu-client-grpc; @totalo
3. - [x] refactor shenyu-client-springcloud; @Albertsirius
4. - [x] refactor shenyu-client-motan; @misaya295
5. - [x] refactor shenyu-client-sofa; @zjcscut
6. - [x] refactor shenyu-client-tars; @yunlongn
7. - [x] refactor shenyu-client-websocket; @lahmXu

847850277:
pls asign task 1 to me. thanks

totalo:
pls asign task 2 to me. thanks

Albertsirius:
pls assign task 3 to me. Thx

lahmXu:
Please assign task 7 to me, thanks!

zjcscut:
pls asign task 5 to me. Thx

yunlongn:
Please assign task 6 to me, thanks!

Albertsirius:
@dragon-zhang Hi, I found the `appName` assigned different value when refactoring  `shenyu-client-springcloud`:
In the original constructor of `SpringCloudClientEventListener`:
<img width="591" alt="Screen Shot 2022-09-25 at 21 36 54" src="https://user-images.githubusercontent.com/1999408/192146429-0a498201-5a8b-4858-8637-4157aa4fd9aa.png">

But in the constructor of `AbstractContextRefreshedEventListener`:
<img width="944" alt="Screen Shot 2022-09-25 at 21 38 25" src="https://user-images.githubusercontent.com/1999408/192146463-21ac9a78-d090-4d4c-ad2b-bff95d797cf6.png">

Should I keep the original or use the superclass assignment?

dragon-zhang:
> @dragon-zhang Hi, I found the `appName` assigned different value when refactoring `shenyu-client-springcloud`: In the original constructor of `SpringCloudClientEventListener`: <img alt="Screen Shot 2022-09-25 at 21 36 54" width="591" src="https://user-images.githubusercontent.com/1999408/192146429-0a498201-5a8b-4858-8637-4157aa4fd9aa.png">
[...]

PR #3937, opened 2022-09-13
[type:refactor] refactor shenyu-client-grpc

totalo:
For #3932 

Make sure that:


dragon-zhang:
please check CI

dragon-zhang:
企业微信截图_1800ce9c-275e-419b-abc6-6292c0a57ecf

totalo:
> 企业微信截图_1800ce9c-275e-419b-abc6-6292c0a57ecf

Thanks, I noticed, currently only testing as a draft.

shenyu-client/shenyu-client-grpc/src/main/java/org/apache/shenyu/client/grpc/GrpcClientEventListener.java
```diff
@@ -34,165 +34,136 @@
...
-
+    
     @Override
-    public void onApplicationEvent(@NonNull final ContextRefreshedEvent contextRefreshedEvent) {
-        Map<String, BindableService> beans = contextRefreshedEvent.getApplicationContext().getBeansOfType(BindableService.class);
-        for (Map.Entry<String, BindableService> entry : beans.entrySet()) {
+    protected Map<String, BindableService> getBeans(final ApplicationContext context) {
+        Map<String, BindableService> result = context.getBeansOfType(BindableService.class);
+        for (Entry<String, BindableService> entry : result.entrySet()) {
             exportJsonGenericService(entry.getValue());
```
dragon-zhang, line 75: `exportJsonGenericService(entry.getValue());` should be put to `#handle`.

totalo: done.

dragon-zhang:
image
I ran the version you modified and the master branch version. There is a problem when handle with `/**`, please fix it.

totalo:
hi，@dragon-zhang, I have fixed the issue with `helloService/**`. And for `access$000` it is because there is a place in the inner class to call the method of the outer class through the this reference of the outer class. The `access$000` method is a method automatically generated by the compiler because the `onError` method of the anonymous inner class in the `helloEveryOne` method accesses the `Log` object of the outer class. And only when the annotation is on the class
Since there is no way to avoid business implementation, I think our solution may be as follows for correctness:
1. Only configure the generation of non-static methods
2. Generated only for public methods
I prefer 2, what you think?

dragon-zhang:
> hi，@dragon-zhang, I have fixed the issue with `helloService/**`. And for `access$000` it is because there is a place in the inner class to call the method of the outer class through the this reference of the outer class. The `access$000` method is a method automatically generated by the compiler because the `onError` method of the anonymous inner class in the `helloEveryOne` method accesses the `Log` object of the outer class. And only when the annotation is on the class Since there is no way to avoid business implementation, I think our solution may be as follows for correctness:
> 
> 1. Only configure the generation of non-static methods
> 2. Generated only for public methods
> 
> I prefer 2, what you think?

agree with you, just go ahead.

totalo:
> ahead

Got it, You can review it, the current one is handled according to public.

totalo:
> > hi，@dragon-zhang, I have fixed the issue with `helloService/**`. And for `access$000` it is because there is a place in the inner class to call the method of the outer class through the this reference of the outer class. The `access$000` method is a method automatically generated by the compiler because the `onError` method of the anonymous inner class in the `helloEveryOne` method accesses the `Log` object of the outer class. And only when the annotation is on the class Since there is no way to avoid business implementation, I think our solution may be as follows for correctness:
> > 
> > 1. Only configure the generation of non-static methods
> > 2. Generated only for public methods
> > 
> > I prefer 2, what you think?
> 
> agree with you, just go ahead.

Got it, You can review it, the current one is handled according to public.

dragon-zhang (approved):

dragon-zhang:
Thanks for your contribution !

PR #3958, opened 2022-09-18
[type fixbug]  Apache ShenYu Admin ultra vires

li-keguo:
ShenYu Admin allows low-privilege low-level administrators to modify high-privilege administrators and low-level passwords, and low-privilege low-level administrators can create users with higher privileges than their own. There are the following fixes for this:

Repair plan：
The system defines the super administrator role, which is used to coordinate other assignments and manage other role administrators. Other role administrators cannot be assigned these super administrator exclusive permissions.

In order to adapt to old users, we also support the following features：

- [x] Whether to enable or disable the exclusive privileges of super administrators.
- [x] Super administrator privileges are customizable.

Note: Modifying user passwords and deleting users can only be done by yourself and the super administrator. This is mandatory and does not apply to the configured super administrator privileges.

### DIFF
--- a/shenyu-admin/src/main/java/org/apache/shenyu/admin/config/properties/DashboardProperties.java
+++ b/shenyu-admin/src/main/java/org/apache/shenyu/admin/config/properties/DashboardProperties.java
@@ -17,15 +17,20 @@
 
 package org.apache.shenyu.admin.config.properties;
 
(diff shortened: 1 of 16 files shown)


tuohai666:
I deleted some sensitive words, please complete the title and content again. The link in the original email has the relevant requirements.

yu199195 (approved):

PR #3976, opened 2022-09-21
[type:feat] log directly to the console

zhc00:
Formerly, we used the script `/bin/start.sh` to startup the ShenYu Admin or Bootstrap. Which logs to the file and then executes `tail` command.
but in the docker container, first, the volume is limited. Then, it is duplicated because the docker service will have one too.

Make sure that:


shenyu-dist/shenyu-admin-dist/entrypoint.sh
```diff
@@ -18,4 +18,18 @@
...
+    JAVA_OPTS=" -server -Xmx2g -Xms2g -Xmn1g -Xss256k -XX:+DisableExplicitGC  -XX:LargePageSizeInBytes=128m"
+    JAVA_OPTS="${JAVA_OPTS} -XX:+UseFastAccessorMethods  -XX:+UseConcMarkSweepGC -XX:+CMSParallelRemarkEnabled -XX:+UseCMSInitiatingOccupancyOnly  -XX:CMSInitiatingOccupancyFraction=70"
+    echo "Use default jvm options: $JAVA_OPTS"
+else
+    JAVA_OPTS=${ADMIN_JVM}
+    echo "Start with the environment variable JAVA_OPTS set: $JAVA_OPTS"
+fi
+
+echo "Starting the Apache ShenYu Admin ..."
+${JAVA_HOME}/bin/java ${JAVA_OPTS} -classpath ${CLASS_PATH} org.apache.shenyu.admin.ShenyuAdminBootstrap
```
yu199195, line 35: why not execute start.sh?

zhc00: first, `start.sh` will log to the file system.
then, this is simpler. Because we pinned the jre version.

yu199195 (approved):

PR #4040, opened 2022-10-06
Fixed Possible Json Ordering Permutations Problem in Tests

BoyuLi4:
## **Fixed Possible Json Ordering Permutations Problem in Tests**

## What is the purpose of the change
The `org.apache.shenyu.admin.service.register.ShenyuClientRegisterDivideServiceImplTest.testBuildHandle` has flakiness.

The test flakiness is due to comparisons between Json Strings and outputs from` JsonUtils.toJson()` and `GsonUtils.getInstance().toJson()`. 
However, JsonObject does not guarantee entry orders, its object is an unordered set of name/value pairs, and internal permutations may occur in the output from toJson().

There are five more similar tests that also have these kinds of flakiness. This is just an example, I will change the other five if you accept these changes.
## Brief changelog

Since the JSON tool used in the Project is GSON, so I use the GSON library `JsonParser `to convert the string back to JSON Object and compare them.  

## Verifying this change

This test avoids nested or different orders of JSON strings that cause flaky errors.
Since I didn't touch any Source code, it should be no impact on the project.
And it passed the local test that the guidelines mentioned.


yu199195 (approved):

BoyuLi4:
Thanks a lot!! can I pull the other five tests with the same problem?

yu199195:
@BoyuLi4 Hi， Sure, thanks for your contribution!

PR #4068, opened 2022-10-11
Drop-down box optimization

QinZiWeiS:
When selecting condition, change it based on the drop-down box after the first item。
fix Added Ruler page display exception
Snipaste_2022-10-10_19-43-41
after：
Snipaste_2022-10-10_19-44-05


yu199195:
can you resolve conflicts

PR #4123, opened 2022-10-21
[type: refactor] refactor AbstractShenyuPlugin#execute

tian-pengfei:
refactor AbstractShenyuPlugin#execute 

+ reduce cyclomatic complexity,
+ Increase code readability

image


shenyu-plugin/shenyu-plugin-base/src/main/java/org/apache/shenyu/plugin/base/AbstractShenyuPlugin.java
```diff
@@ -79,58 +79,33 @@ public abstract class AbstractShenyuPlugin implements ShenyuPlugin {
...
-                        cacheSelectorData(path, selectorData);
-                    }
-                    return handleSelectorIfNull(pluginName, exchange, chain);
-                } else {
-                    if (matchCacheConfig.getEnabled() && matchSelectorData.getLeft()) {
-                        cacheSelectorData(path, selectorData);
-                    }
-                }
-            } else {
-                if (StringUtils.isBlank(selectorData.getId())) {
```
yu199195, line 107: why remove this，
 if SelectorData not null ，&& selectorData.getId() is null   ，need 
return handleSelectorIfNull(pluginName, exchange, chain);

tian-pengfei: thats not gonna happen,because when cached selectorData.getId() is null ,obtainMatchedSelectorData will return null
image

image

Issue #4072, opened 2022-10-12
[BUG] Shenyu-Admin > BasicConfig > Plugin list page 2 cannot display


Once2012:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I do not know when to start, the second page of  BasicConfig > Plugin cannot be displayed（empty white page），but the first and third page is OK（Page size is 12）

F12 of Chrome show this:
![1665555847736](https://user-images.githubusercontent.com/19165279/195266985-98990f57-d14a-4be6-93de-cb855529f4f9.png)

### Environment

```markdown
ShenYu version(s):2.5.0
```

Once2012:
And when I open Request plugin，all page of Shenyu-Admin can not display，so I have to close the Request plugin by modifing Database table (plugin.enable => 0）.
The Request plugin is in Plug in list Page 2，maybe it cause the problem?

The stacktrace of Shenyu-Admin is below:
![image](https://user-images.githubusercontent.com/19165279/195269103-ea1e1af1-e0df-44fe-ace5-3bcc2edeed13.png)

yu199195:
can you execute  right SQL ？ about some plugin~

PR #4146, opened 2022-10-28
[ISSUE #4072] fix plugin update bug by modifying config field setter

gngsn:
### Problem
When updating plugin corresponding to (Menu) BasicConfig > Plugin page, config field is updated BLANK, not null. The config data is saved BLANK in database and it makes difference from initial setting. In addition, this causes JSON parsing errors when return to front-side.

스크린샷 2022-10-29 오전 1 17 42

스크린샷 2022-10-29 오전 1 19 30

스크린샷 2022-10-29 오전 1 24 29

스크린샷 2022-10-29 오전 1 19 45


yu199195:
@gngsn  thanks for you contributor，But CI has a error about  shenyu checkstyle, you need modify it ：

> Error:  src/main/java/org/apache/shenyu/admin/model/dto/PluginDTO.java:[122,9] (whitespace) WhitespaceAfter: 'if' is not followed by whitespace.

yu199195 (approved):

PR #4176, opened 2022-11-11
nacos_npe

wzhangNJ:
shenyu-cleint register nacos's service, Not registered nacos's config. 
readData is null， cause subscribeMetadata method npe.


wzhangNJ (commented):
shenyu-cleint register nacos's service, Not registered nacos's config.
readData is null， cause subscribeMetadata method npe.

KevinClair (approved):

PR #4089, opened 2022-10-15
[BUG #4080]fix bug When using post requests, you can add parameters after the ur…

ableYang123:
[BUG] When using post requests, you can add parameters after the url to crack signature authentication #4080

Issue #4080, opened 2022-10-14
[BUG] When using post requests, you can add parameters after the url to crack signature authentication 


ableYang123:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When using post requests, you can add parameters after the url to crack signature authentication 

### Steps To Reproduce

When I catch a correct request on the network
![image](https://user-images.githubusercontent.com/29156983/195748688-e25c42d7-2372-4d0a-bc7f-be5f0b825ad0.png)

I try to tamper with the parameter content, the program will report a signature error
![image](https://user-images.githubusercontent.com/29156983/195748794-91ef7efa-04d6-4a8c-a4c9-6ac2651671f9.png)

However, when I concatenate the previous parameters after the url, I can crack the parameter signature 
![image](https://user-images.githubusercontent.com/29156983/195748894-63d541c6-70a3-4f18-9c72-d7345decafe0.png)

### Environment

```markdown
ShenYu version(s):2.5.0
```

### Anything else?

The problem here is that the parameter after the URL overrides the parameter with the same name in the request body 
![image](https://user-images.githubusercontent.com/29156983/195742048-e6fb99f8-9c54-4a46-81d9-c3e6fe3fbad1.png)
should put the request header and url parameters on one map and the request body on the other

yunlongn:
There is no way to prevent the user from entering the same parameter name in parameter and body.

yunlongn:
can you perfect him.

ableYang123:
OK,I trry

PR #4187, opened 2022-11-14
[type:bugfix]Fixed signing function of  the sign plugin When there is no request body

tian-pengfei:
Make sure that:

when I tested signing function of the sign plugin,I found It is not the same as the function description of the document.
should be :
```tex
path/springcloud/class/annotation/gettimestamp1668393864220version1.0.0EA8919756AFF40A3896AFEA161E23B29
```
actual:
0530a3b79d788f385781a6480e457fd

But I found this commit(#4089) from git history ,its not good for no request body in signing

**in addition**
I think the commit #4089 is not good enough for solving the bug #4080

I have some ideas
We could define a simple json format to calculate the sign

```json
{
    "header":"path/api/service/abc?id=123timestamp1571711067186version1.0.0506EEB535CF740D7A755CB4B9F4A1536",
    "body":"id123nameorder"
}
```
or
```json
{
    "header":{
        "path":"path/api/service/abc?id=123",
        "timestamp":"1571711067186",
        "version":"1.0.0",
        "appKey":"506EEB535CF740D7A755CB4B9F4A1536"
    },
    "body":{
        "id":"path/api/service/abc?id=123",
        "name":"order"
    }
}
```
the body is optional
```java
sign = DigestUtils.md5DigestAsHex(json.getBytes()).toUpperCase()
```

**in addition**
I found there are too many parameters in request header, I think we could redesign with reference to JWT format


shenyu-plugin/shenyu-plugin-sign/src/main/java/org/apache/shenyu/plugin/sign/service/DefaultSignService.java
```diff
@@ -126,6 +126,6 @@ private Pair<Boolean, String> sign(final ShenyuContext shenyuContext, final Serv
     }
 
     private String buildExtSignKey(final String signKey, final ShenyuContext shenyuContext) {
-        return String.join("", Constants.TIMESTAMP, shenyuContext.getTimestamp(), Constants.PATH, shenyuContext.getPath(), Constants.VERSION, "1.0.0", signKey);
+        return String.join("", Constants.PATH, shenyuContext.getPath(), Constants.TIMESTAMP, shenyuContext.getTimestamp(), Constants.VERSION, "1.0.0", signKey);
```
yu199195, line 129: Why change this order? According to the documentation?

tian-pengfei: image

after https://github.com/apache/shenyu/pull/4089 ,Actually its funticon doesnt meet the documentation description,I change that ,it can be compatible  when there is no request body. but  its still incompatible ,when there have request body

tian-pengfei:
I think the best way is to redesign header parameters  and the calculation method of  signature

tian-pengfei:
@yu199195  Sir ,I  design a simple structrue . I would like to know what do you think , I hope you can give me some advices 

header:
```json
{
    "alg":"HMD5",
    "version":"1.0.0",
    "appKey":"506EEB535CF740D7A755CB4B9F4A1536",
    "timestamp":"1571711067186"
}
```
HMD5-->HMAC-MD5   , HS256--->HMAC-SHA-256 ,etc

> token = base64urlEncoding(header) + '.' + base64urlEncoding(signature)
```tex
signature = HMAC(
  base64urlEncoding(header) + Relative URL + Body*,
  secret
);

* indicate Optional , it depends on config
 Relative URL = path [ "?" query ] [ "#" fragment ]  eg: /apache/shenyu/pulls?q=is%3Apr+is%3Aclosed
```

li-keguo (approved):

shenyu-plugin/shenyu-plugin-sign/src/main/java/org/apache/shenyu/plugin/sign/service/DefaultSignService.java
tian-pengfei: image



In #4242, I changed it back according to the current(2.5.1) document . The comments above are based on the 2.5.0 documentation

PR #4239, opened 2022-12-05
fix #4237:delete rule cache BaseData#RULE_MAP when delete rule

### LINKED ISSUE #4237: [BUG] Caches in BaseDataCache#RULE_MAP will not be removed when delete rule

### Current Behavior

When i delete a rule which belong to a selector，the local cache in a currentMap(BaseDataCache#RULE_MAP) will not be removed. This may affect the matching of rules when use the same condition as the deleted rules, it will chose the wrong config. And it also consume memory. And i found when update the rule at NacosCacheHandler#updatePluginMap or in other handler, it will only unSubscribe and onSubscribe the current rules, the deleted rules will not be handled. 
```java
protected void updateRuleMap(final String configInfo) {
        try {
            List<RuleData> ruleDataList = GsonUtils.getInstance().toObjectMapList(configInfo, RuleData.class).values()
                    .stream().flatMap(Collection::stream)
                    .collect(Collectors.toList());
            ruleDataList.forEach(ruleData -> Optional.ofNullable(pluginDataSubscriber).ifPresent(subscriber -> {
                subscriber.unRuleSubscribe(ruleData);
                subscriber.onRuleSubscribe(ruleData);
            }));
        } catch (JsonParseException e) {
            LOG.error("sync rule data have error:", e);
        }
}
```

### Expected Behavior

the local cache will be removed when delete the rule.

### Steps To Reproduce

1. create a new selector and add two rule in divide plugin.
image

2. do invoke and debug at AbstractShenyuPlugin.
image
Now, there are two cache records for this selector

3. delete one rule.
image

4. do invoke and debug again.
image
The cache of the rule named intfc2 has not be removed.

### Environment

```markdown
ShenYu version(s):2.5.0,master
```

### Debug logs

_No response_

### Anything else?

_No response_

simelyWWei:
Make sure that:


yu199195 (approved):

yu199195:
long time, it be closed

PR #4266, opened 2022-12-14
ShenyuResult  MediaType.APPLICATION_JSON_UTF8

key9527:
建议这里还是用MediaType.APPLICATION_JSON_UTF8，实测在对接某些老旧系统时，这里不指定charset，对方接收会乱码。


shenyu-plugin/shenyu-plugin-api/src/main/java/org/apache/shenyu/plugin/api/result/ShenyuResult.java
```diff
@@ -70,7 +70,7 @@ default MediaType contentType(ServerWebExchange exchange, Object formatted) {
         if (Objects.nonNull(clientResponse) && clientResponse.headers().contentType().isPresent()) {
             return clientResponse.headers().contentType().get();
         }
-        return MediaType.APPLICATION_JSON;
+        return MediaType.APPLICATION_JSON_UTF8;
```
yu199195, line 73: Are they different, can they be dynamically configured or returned based on front-end incoming？

key9527: this code is used for WebFluxResultUtils.result(exchange, error);   for example：
`
        if (CollectionUtils.isEmpty(upstreamList)) {
            LOG.error("divide upstream configuration error： {}", selector);
            Object error = ShenyuResultWrap.error(exchange, ShenyuResultEnum.CANNOT_FIND_HEALTHY_UPSTREAM_URL);
            return WebFluxResultUtils.result(exchange, error);
        }
`
 If I customize the error message in chinese，when this msg returned ,  The receiver cannot recognize this charset。because this is shenyu returned， not back-end service。

Issue #4276, opened 2022-12-19
[BUG] After the HTTP request passes through the shenyu proxy, the cookie is repeated


wls981:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

![image](https://user-images.githubusercontent.com/15735480/208342780-813cd52f-8413-4c5a-bdb2-41a8b55f1a52.png)
![image](https://user-images.githubusercontent.com/15735480/208342814-413f690b-1521-495a-98c6-315ccd589986.png)
![image](https://user-images.githubusercontent.com/15735480/208343471-cb8a9dde-1702-444e-ba68-d32fe8fa96b9.png)

As shown in the picture above, after being proxied by the divide plug-in, the cookie in the returned HTTP response header is repeated.  

shenyu 2.4.3 and 2.5.0 can be reproduced.

### Expected Behavior

The cookie should not be repeated, because I find that after an http url passes through the gateway, 
the <Transfer-Encoding: chunked> header in the response header is also repeated, which causes the nginx in front of the gateway to report an error.

### Steps To Reproduce

Download shenyu-2.5.0-bootstrap-bin.tar.gz and shenyu-2.5.0-admin-bin.tar.gz directly from the official website, and simply configure the divide plug-in code to reproduce.

### Environment

```markdown
ShenYu version(s):2.4.3  、 2.5.0
```

### Debug logs

none

yu199195:
The response header is duplicated, is it only the cookie and the rest will be duplicated?

PR #4283, opened 2022-12-25
[ISSUE #4276] remove redundant cookie setting

RayayChung:
fix: #4276 
Make sure that:


shenyu-plugin/shenyu-plugin-httpclient/src/main/java/org/apache/shenyu/plugin/httpclient/WebClientPlugin.java
```diff
@@ -63,8 +63,7 @@ protected Mono<ClientResponse> doRequest(final ServerWebExchange exchange, final
                         .flatMap(bytes -> Mono.fromCallable(() -> Optional.ofNullable(bytes))).defaultIfEmpty(Optional.empty())
                         .flatMap(option -> {
                             final ClientResponse.Builder builder = ClientResponse.create(response.statusCode())
-                                    .headers(headers -> headers.addAll(response.headers().asHttpHeaders()))
-                                    .cookies(cookies -> cookies.addAll(response.cookies()));
```
yu199195, line 67: why remove cookies setting？

RayayChung: Setting cookies is just part of setting headers. So the previous version codes may lead setting cookies twice.

yu199195: Is there a time when cookies is not passed from the header?

RayayChung: the data of `response.cookies()` is the subset of `response.headers().asHttpHeaders()`, therefore setting headers is also setting cookies in the sam time. I have never met the case that cookies is not present in headers. Is this case exists?

yu199195 (approved):

PR #4359, opened 2023-02-07
Fix enable oauth2 but the plugin does not execute problem

Zhoutzzz:
Fix while open oauth2 plugin but it does not execute problem

Make sure that:


shenyu-plugin/shenyu-plugin-oauth2/src/main/java/org/apache/shenyu/plugin/oauth2/OAuth2Plugin.java
```diff
@@ -70,8 +68,8 @@ public String named() {
 
     @Override
     public boolean skip(final ServerWebExchange exchange) {
-        Boolean skipStatus = exchange.<Boolean>getAttribute(OAUTH2_ENABLE);
-        return skipStatus == null || skipStatus;
+        Boolean skipStatus = exchange.<Boolean>getAttribute("enable");
+        return skipStatus == null || !skipStatus;
```
li-keguo, line 72: `!Boolean.TRUE.equals(exchange.getAttribute("enable"))`

not enable is skip. 
That might make the meaning clearer

li-keguo (approved):

PR #4366, opened 2023-02-08
[ISSUE #4356] optimize UpstreamCheckService

### LINKED ISSUE #4356: [Task] optimize UpstreamCheckService
### Description

In the k8s environment, due to the ip change after the client container is restarted, the UpstreamCheckService health check is easy to fail. Due to the use of single-thread health check, it is easy to slow down the health check of other nodes, and eventually cause the OOM of shenyu-admin.

This task needs you to use muti-thread to optimize `org.apache.shenyu.admin.service.impl.UpstreamCheckService#scheduled`, you can refer to the implementation of `org.apache.shenyu.loadbalancer.cache.UpstreamCacheManager`.

Please remember to allow users to configure the number of threads, maybe you should also do some modify to `org.apache.shenyu.loadbalancer.cache.UpstreamCheckTask` to support config.

Good luck !

### Task List

- [ ] optimize UpstreamCheckService;
- [ ] support config the number of threads in `UpstreamCheckTask`;

iwangjie:
Fix #4356 

Make sure that:


shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/UpstreamCheckService.java
```diff
@@ -243,56 +258,91 @@ public void replace(final String selectorId, final List<CommonUpstream> commonUp
...
         } catch (Exception e) {
             LOG.error("upstream scheduled check error -------- ", e);
         }
     }
 
+    private void doCheck() {
+        // check zombie
+        ZOMBIE_SET.parallelStream().forEach(this::checkZombie);
+        // check up
+        UPSTREAM_MAP.forEach(this::check);
```
dragon-zhang, line 272: Reserve the judgment and remove `parallelStream`.

dragon-zhang (approved):

PR #4380, opened 2023-02-13
fix #4378 plugin rule update did not take effect in time.

### LINKED ISSUE #4378: [BUG] Plugin rule update did not take effect in time.
### Bug Description

After the plug-in matching rule is updated, the new rule of the request gateway does not take effect.

synchronous invalid.

#### Context

After I modified the rule in shenyu admin, the synchronization gateway was triggered, and the gateway did not seem to handle it correctly 
 
The gateway uses prefix tree to store rules, and the processing logic of updating action is putNode, so the historical rules are not cleaned up, which leads to the request interface still using the old rules after updating rules

Divide plugin, the source address can still be accessed after the Condition changes, which is fatal:
image

Similarly, WAF plugin
After the firewall rule switch changes, it does not take effect, which is also fatal:

image

image

iwangjie:
fix #4378 

shenyu-admin update rule data. shenyu-bootstrap not effective.

Make sure that:

### DIFF
--- a/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/RuleServiceImpl.java
+++ b/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/RuleServiceImpl.java
@@ -152,11 +152,17 @@ public class RuleServiceImpl implements RuleService {
         Assert.notNull(before, "the updated rule is not found");
         RuleDO ruleDO = RuleDO.buildRuleDO(ruleDTO);
         final int ruleCount = ruleMapper.updateSelective(ruleDO);
+
+        // need old data for cleaning
+        final List<RuleConditionDO> beforeRuleCondition = ruleConditionMapper.selectByQuery(new RuleConditionQuery(ruleDO.getId()));
+
         //delete rule condition then add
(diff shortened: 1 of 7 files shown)


shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/RuleServiceImpl.java
```diff
@@ -152,11 +152,17 @@ public int update(final RuleDTO ruleDTO) {
         Assert.notNull(before, "the updated rule is not found");
         RuleDO ruleDO = RuleDO.buildRuleDO(ruleDTO);
         final int ruleCount = ruleMapper.updateSelective(ruleDO);
+
+        // need old data for cleaning
+        final List<RuleConditionDO> beforeRuleCondition = ruleConditionMapper.selectByQuery(new RuleConditionQuery(ruleDO.getId()));
```
yu199195, line 157: update beforeRuleCondition ？  why？

iwangjie: because the gateway uses the prefix tree to handle the cache, I can not clean up the history cache if Conditions change, so I pass the pre-modified Conditions to the past for precision cleaning.

moremind (approved):

PR #4443, opened 2023-03-06
Use high-performance clocks instead of frequent system calls.

iwangjie:
` System.currentTimeMillis () ` frequently switches from user mode to kernel mode 
 
The system has only one global clock source, and high concurrency or frequent access will cause serious contention 
 
`ShenyuClock` Accuracy of 1 MILLISECONDS, compatible with most scenarios, especially in high-frequency access.
 
The operation can be completed in user mode, which reduces the loss of entering kernel mode.

https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8185891

🤔I think it can be used in the load balancer, Seek community views.


yu199195 (dismissed):

shenyu-common/src/main/java/org/apache/shenyu/common/utils/ShenyuClock.java
```diff
@@ -0,0 +1,101 @@
...
+     */
+    private static ShenyuClock instance() {
+        return InstanceHolder.INSTANCE;
+    }
+
+    /**
+     * Schedule clock updating.
+     */
+    private void scheduleClockUpdating() {
+        ScheduledThreadPoolExecutor scheduler = new ScheduledThreadPoolExecutor(1, runnable -> {
```
yu199195, line 64: used ShenyuThreadFactory

iwangjie: good Idea. i fixed.

yu199195 (changes requested):
wait

PR #4522, opened 2023-03-31
[ISSUE #4508]Tolerant Path Slash. fix #4508

### LINKED ISSUE #4508: [Question] Consider compatibility adjacent slashes.
### Question

RFC 3986 said:

If a URI contains an authority component, then the path component must either be empty or begin with a slash ("/") character. If a URI does not contain an authority component, then the path cannot begin with two slash characters ("//"). In addition, a URI reference (Section 4.1) may be a relative-path reference, in which case the first path segment cannot contain a colon (":") character. The ABNF requires five separate rules to disambiguate these cases, only one of which will match the path substring within a given URI reference. We use the generic term "path component" to describe the URI substring matched by the parser to one of these rules.

---
But Tomcat and Nginx, as a high-occupancy web container, violates the constraint

image

**Whether or not Shenyu as a gateway needs to tolerate this has improved compatibility and less surprise, helping the corporate infrastructure group to keep the“Non-canonical” writing of business systems unchanged, i think this is necessary because violating this specification is less costly for a company with many existing services than complying with it, and we should be lenient, right?**

**If this is necessary, please assign it to me🙋🏻‍♀️ and I will support it.**

iwangjie:
## before
image

## after

image

Make sure that:


shenyu-bootstrap/src/main/java/org/apache/shenyu/bootstrap/config/CollapseSlashesConfiguration.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.net.URI;
+
+/**
+ * Tolerant Path Slash.
+ */
+@Configuration
+public class CollapseSlashesConfiguration {
+
+    @Bean
```
yu199195, line 35: I think a switch needs to be added here, otherwise all requests will go through this webfilter

iwangjie: > I think a switch needs to be added here, otherwise all requests will go through this webfilter

Hi. do you mean a rule-level switch or a gateway-level switch?

yu199195: gateway-level ， in  application.yaml

iwangjie: > gateway-level ， in application.yaml

OK, see what you mean. I'll fix it.

shenyu-bootstrap/src/main/java/org/apache/shenyu/bootstrap/config/CollapseSlashesConfiguration.java
```diff
@@ -0,0 +1,56 @@
...
+    @ConditionalOnProperty(
+            value = "shenyu.switchConfig.collapseSlashes",
+            havingValue = "true",
+            matchIfMissing = true
```
yu199195, line 40: matchIfMissing = false

iwangjie: ok. I deleted it to apply the default false.

li-keguo (approved):

PR #4606, opened 2023-05-03
【Type:feature 】The SPEL in the mock plugin is secure by default

li-keguo:
work list:

- [x] Use SimpleEvaluationContext as the default secure SPEL.
- [x] Provides a Standard Expression Generator, which is not loaded by default.


tian-pengfei:
Well done! maybe standardSPELExpression can completely replace old one . I think we can remove old one. Have you ever try different `MethodSecurityEvaluationContext`.  are threre any differences between `StandardEvaluationContext` and `MethodSecurityEvaluationContext`?

li-keguo:
> Well done! maybe standardSPELExpression can completely replace old one . I think we can remove old one. Have you ever try `MethodSecurityEvaluationContext`. are threre any differences between `StandardEvaluationContext` and `MethodSecurityEvaluationContext`?

The original has been replaced with a secure implementation. see https://docs.spring.io/spring-framework/docs/5.0.6.RELEASE/javadoc-api/org/springframework/expression/spel/support/SimpleEvaluationContext.html

StandardExpressionGenerator is not loaded by default and needs to be enabled by the user.

Without studying the MethodSecurityEvaluationContext, but only providing some clues, he came from spring-security.

tian-pengfei (approved):

Issue #4613, opened 2023-05-05
[Volunteer]  Support Admin Internationalization


yu199195:
## Background
Shenyu is a native API gateway for service proxy, protocol translation and API governance. It can manage and maintain the API through Shenyu-admin, and support internationalization in Chinese and English. Unfortunately, Shenyu-admin is only internationalized on the front end. The message prompt returned by the back-end interface is still in English. Therefore, we need to implement internationalization support for the back-end interface.This will lay a good foundation for shenyu to move towards more language support.

## Relevant skills
Related skills spring resources
Spring Internationalization
Front-end react framework
API reference
```
            java.util.Locale;
            org.springframework.context.MessageSource;
            org.springframework.context.support.ResourceBundleMessageSource; 
Interface effect example
            ## zh request example
            POST http://localhost:9095/plugin
            Content-Type: application/json
            Location: cn-zh
            X-Access-Token: xxx
            {
            "name": "test-create-plugin",
            "role": "test-create-plugin",
            "enabled": true,
            "sort": 100
            }
            Respone
            {
            "code": 600,
            "message": "未登录"
            }
            
            ### en request example
            POST http://localhost:9095/plugin
            Content-Type: application/json
            Location: en
            X-Access-Token: xxx
            {
            "name": "test-create-plugin",
            "role": "test-create-plugin",
            "enabled": true,
            "sort": 100
            }
            Respone
            {
            "code": 600,
            "message": "token is error"
            } 
```

## Task List
- [ ] The task discussed with the tutor how to achieve the internationalization of shenyu-admin background
- [ ] Some prompt message translation
- [ ] Get through the internationalization of front-end, obtain the client region information through http protocol, support the language of the corresponding region.
- [ ] Leave the extension of other multi-language internationalization support interface, so as to facilitate the localization transformation of subsequent users.

Difficulty: Major
Potential mentors:
Keguo Li, mail: likeguo@apache.org
Project Devs, mail: dev@shenyu.apache.org

peng-heng:
I want to try to do it.

peng-heng:
![image](https://github.com/apache/shenyu/assets/11962364/75e25f01-c239-4673-875b-93b3bfe986d4)
 please fix this test case. e2e / e2e (postgres) (pull_request)  will filed.

December-Pb:
Is this task still open? I noticed there are already 3 commits.

PR #4640, opened 2023-05-17
[Volunteer] Support Admin Internationalization 

peng-heng:
#4613


li-keguo:
hi It's not a good realization.
We should use spring's internationalization.
see

https://docs.spring.io/spring-framework/docs/5.1.3.RELEASE/spring-framework-reference/core.html#spring-core

Generally, in the web scenario, the locale context is obtained through the LocaleContextHolder.
```java

messageSource.getMessage("test.message",   LocaleContextHolder.getLocale())
```
see :AcceptHeaderLocaleResolver

peng-heng:
it's ok . i will closed this pr. and think about spring's internationalization.

PR #4729, opened 2023-06-12
[type:fix]fix SofaServiceEventListenerTest.testBuildMetaDataDTO

tian-pengfei:
Make sure that:

I think the `SofaServiceEventListenerTest.testBuildMetaDataDTO` method has a problem.

Sometimes the expected one's timeMillis is diffrent from the actual's ,Because of the difference of creation time

image

5cf5a5971672a56fcf3cd68a29f71c2

image

image

So, Removes comparing timeMillis  in MetaDataRegisterDTO#equals


dragon-zhang (approved):
LGTM

dragon-zhang (changes requested):
the `equals` maybe also need changes, too ?

tian-pengfei:
> the `equals` maybe also need changes, too ?

That depends on whether comparing two objects to be equal should take into account the time they was created

tian-pengfei:
> the `hashCode` maybe also need changes, too ?

yes I will fix it

dragon-zhang (approved):

Issue #4686, opened 2023-05-30
[Task] Client Active Offline


loongs-zhang:
### Description

Currently, shenyu only supports passive offline, during which the front-end will have a clear perception (i.e. response 408), indicating that it is better to support active offline.

The idea is roughly:
1. The client utilizes the mechanism provided by `ShenyuClientShutdownHook` to call the offline interface of shenyu admin;
2. In the offline interface of shenyu admin, we update the database and send events;

### Task List

- [x] springmvc client
- [x] ~~spring cloud client wait for https://github.com/apache/shenyu/issues/4680~~
- [ ] springmvc client active offline support shenyu-admin cluster
- [x] ZookeeperClientRegisterRepository active offline @xuziyang
- [x] NacosClientRegisterRepository active offline @847850277 
- [x] EtcdClientRegisterRepository active offline @xuziyang
- [x] ConsulClientRegisterRepository active offline @847850277 
- [x] ApolloClientRegisterRepository active offline @xuziyang

weiweily:
hello， when this will be ok？

dragon-zhang:
> hello， when this will be ok？

At least 2.6.0.

dragon-zhang:
After discuss, we think the spring cloud active offline is not needed.

xuziyang:
I would like to take on this job.

847850277:
please asign  `ConsulClientRegisterRepository active offline` and `  NacosClientRegisterRepository active offline` to me.

xuziyang:
please asign  ApolloClientRegisterRepository active offline to me.

loongs-zhang:
see https://github.com/apache/shenyu/issues/5324

loongs-zhang:
https://github.com/apache/shenyu/issues/5324

PR #4806, opened 2023-07-05
[ISSUE #4686] ZookeeperClientRegisterRepository active offline

xuziyang:
[ISSUE #4686] ZookeeperClientRegisterRepository active offline


dragon-zhang:
need impl `offline` for `ZookeeperClientRegisterRepository`.

xuziyang:
> need impl `offline` for `ZookeeperClientRegisterRepository`.

Personally, I believe that there is no need to implement the offline () method for the following reasons:

When the client side service stops, ShenyuClientShutdownHook will be triggered, which will close zkClient. After closing zkClient, the registered URI information will naturally be deleted by zk, and shenyu admin can sense the offline of the service.

dragon-zhang (approved):
I saw the `org.apache.shenyu.register.client.zookeeper.ZookeeperClientRegisterRepository#registerURI`, you are correct.

PR #4807, opened 2023-07-05
[type: feature] e2e-spring-cloud change e2e action

HaiqiQin:
Add testing for other modules of e2e.


.github/workflows/e2e.yaml
```diff
@@ -142,12 +142,66 @@ jobs:
           bash ./shenyu-e2e/script/storage_init.sh
           ./mvnw -B -f ./shenyu-e2e/pom.xml -pl shenyu-e2e-case/shenyu-e2e-case-http -Dstorage=${{ matrix.storage }} test
 
+  e2e-other:
```
moremind, line 145: e2e-case

moremind (changes requested):

.github/workflows/e2e.yaml
HaiqiQin: I will change it.

PR #4867, opened 2023-07-18
[type: fix] fix client shutdown close method call twice

847850277:
The client's close will be called twice when it is closed.
Because of Spring When the bean is destroyed, it will proactively call the `close` or `shutdown` methods. At the same time, Shenyu also defined ShenyuClientShutdownHook. Therefore, renaming the `close` method to ` closeRepository` solves the problem of being called twice.

image

Make sure that:

### DIFF
--- a/shenyu-client/shenyu-client-core/src/main/java/org/apache/shenyu/client/core/shutdown/ShenyuClientShutdownHook.java
+++ b/shenyu-client/shenyu-client-core/src/main/java/org/apache/shenyu/client/core/shutdown/ShenyuClientShutdownHook.java
@@ -55,7 +55,7 @@ public class ShenyuClientShutdownHook {
 
     public ShenyuClientShutdownHook(final ShenyuClientRegisterRepository repository, final ShenyuRegisterCenterConfig config) {
         String name = String.join("-", hookNamePrefix, String.valueOf(hookId.incrementAndGet()));
-        Runtime.getRuntime().addShutdownHook(new Thread(repository::close, name));
+        Runtime.getRuntime().addShutdownHook(new Thread(repository::closeRepository, name));
         LOG.info("Add hook {}", name);
         ShenyuClientShutdownHook.props = config.getProps();
     }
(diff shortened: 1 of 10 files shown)


yu199195:
client shutdown close method call twice ？

847850277:
> client shutdown close method call twice ？ how to fixed？

yes。
1、`ShenyuClientRegisterRepository` bean destry the `close` methd will be call.
2、and `Runtime.getRuntime().addShutdownHook(new Thread(repository::closeRepository, name));` will be call.

`By default, JavaConfig configured beans are used. If there is a close or shutdown method, this method will be automatically executed when the bean is destroyed. If you do not want to execute this method, add @ Bean (destroyMethod="") to prevent the destruction method from starting`

i try rename `close` methd to `closeRepository` can solve spring call default close method.

dragon-zhang:
> client shutdown close method call twice ？ how to fixed？

```java
@Target({ElementType.METHOD, ElementType.ANNOTATION_TYPE})
@Retention(RetentionPolicy.RUNTIME)
@Documented
public @interface Bean {

	//......

	/**
	 * The optional name of a method to call on the bean instance upon closing the
	 * application context, for example a {@code close()} method on a JDBC
	 * {@code DataSource} implementation, or a Hibernate {@code SessionFactory} object.
	 * The method must have no arguments but may throw any exception.
	 * <p>As a convenience to the user, the container will attempt to infer a destroy
	 * method against an object returned from the {@code @Bean} method. For example, given
	 * an {@code @Bean} method returning an Apache Commons DBCP {@code BasicDataSource},
	 * the container will notice the {@code close()} method available on that object and
	 * automatically register it as the {@code destroyMethod}. This 'destroy method
	 * inference' is currently limited to detecting only public, no-arg methods named
	 * 'close' or 'shutdown'. The method may be declared at any level of the inheritance
	 * hierarchy and will be detected regardless of the return type of the {@code @Bean}
	 * method (i.e., detection occurs reflectively against the bean instance itself at
	 * creation time).
	 * <p>To disable destroy method inference for a particular {@code @Bean}, specify an
	 * empty string as the value, e.g. {@code @Bean(destroyMethod="")}. Note that the
	 * {@link org.springframework.beans.factory.DisposableBean} callback interface will
	 * nevertheless get detected and the corresponding destroy method invoked: In other
	 * words, {@code destroyMethod=""} only affects custom close/shutdown methods and
	 * {@link java.io.Closeable}/{@link java.lang.AutoCloseable} declared close methods.
	 * <p>Note: Only invoked on beans whose lifecycle is under the full control of the
	 * factory, which is always the case for singletons but not guaranteed for any
	 * other scope.
	 * @see org.springframework.beans.factory.DisposableBean
	 * @see org.springframework.context.ConfigurableApplicationContext#close()
	 */
	String destroyMethod() default AbstractBeanDefinition.INFER_METHOD;
}
```

The `close` in `ShenyuClientRegisterRepository` will be called by spring, so just change the `close` method name.

dragon-zhang (dismissed):

Issue #4928, opened 2023-07-30
[BUG] <The operation 'unpublished' failed for any apidoc.>


lianjunwei:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Select a certain apidoc, then open the modify page and operate "unpublished". The error message is as follows:
nested exception is org.apache.ibatis.exceptions.TooManyResultsException: Expected one result (or null) to be returned by selectOne(), but found: 2

![image](https://github.com/apache/shenyu/assets/13819823/6345c416-3520-477b-9461-8884d6ec7ba6)

### Expected Behavior

Successfully unpublished.

### Environment

```markdown
ShenYu version(s):master.
```

PR #4929, opened 2023-07-30
<ISSUE #4928>Fix the error that occurred when submitting 'unpublished' to apidoc.

lianjunwei:
detail see : #4928 
modify logic: When modifying apidoc, it is supported to delete registered apis, but not when creating.

Make sure that:


shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/ApiServiceImpl.java
```diff
@@ -124,7 +128,7 @@ private String update(final ApiDTO apiDTO) {
             }
             if (ApiStateEnum.PUBLISHED.getState() == apiDO.getState()) {
                 register(apiDO);
-            } else if (ApiStateEnum.OFFLINE.getState() == apiDO.getState()) {
+            } else if (ApiStateEnum.UNPUBLISHED.getState() == apiDO.getState()) {
```
dragon-zhang, line 131: The state judge should not change.

shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/ApiServiceImpl.java
```diff
@@ -176,12 +180,17 @@ private void unregister(final ApiDO apiDO) {
                     .collect(Collectors.toList()));
         }
         //clean selector
-        Optional.ofNullable(selectorService.findByName(apiDO.getContextPath()))
+        Optional.ofNullable(selectorService.findByNameAndPluginName(apiDO.getContextPath(), PluginEnum.DIVIDE.getName()))
```
dragon-zhang, line 183: Other rpc types need to be considered.

shenyu-admin/src/main/java/org/apache/shenyu/admin/mapper/SelectorMapper.java
```diff
@@ -88,9 +88,9 @@ public interface SelectorMapper extends ExistProvider {
      * select select by name.
      *
      * @param name the name
-     * @return selector do
+     * @return selector do list
      */
-    SelectorDO selectByName(String name);
+    List<SelectorDO> selectByName(String name);
```
moremind, line 93: new method do this? selectListByName or findByName?

moremind (changes requested):

shenyu-admin/src/main/java/org/apache/shenyu/admin/mapper/SelectorMapper.java
lianjunwei: Yes. The selectByName method is currently only used in one place. I plan to remove it later because the original method may generate errors during use (using 'contextpath' as a conditional query usually results in 2 records, one from the proxy proxy plugin and the other from the contextpath plugin). This PR only quickly fixes discovered bugs.

moremind: use a new method instead of changing the method

moremind (changes requested):

shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/ApiServiceImpl.java
lianjunwei: All proxy plugin IDs should be checked here. In order to quickly fix errors, I only specified the divide plugin. I plan to provide a method to obtain the IDs of all proxy type plugins in the plugin's enumeration class.
But I found two enumeration classes about plugin plugins.  `org.apache.shenyu.e2e.model.Plugin` and `org. apache. shenyu. common. enums. PluginEnum `.
But only the former has a plugin ID. I want to merge `org. apache. shenyu. e2e. model. Plugin` and `org. apache. shenyu. common. enums. PluginEnum`, but the scope of the changes is a bit large, so I want to keep it until the next refactoring. If this PR considers other proxy plugins, I can make minor changes.

shenyu-admin/src/main/java/org/apache/shenyu/admin/mapper/SelectorMapper.java
lianjunwei: > use a new method instead of changing the method

That's right.

shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/ApiServiceImpl.java
lianjunwei: For apidoc, the OFFLINE case is currently not implemented. In the next version, I plan to implement the OFFLINE function based on the application granularity of the apidoc module. The unregister() method should belong to the UNPUBLISHED case, not OFFLINE, so I modified the conditional judgment.

shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/ApiServiceImpl.java
lianjunwei: > Other rpc types need to be considered.

I have already implemented it, push pr immediately.

lianjunwei: Hi, please review if it's okay. @dragon-zhang

shenyu-admin/src/main/java/org/apache/shenyu/admin/service/SelectorService.java
```diff
@@ -87,47 +87,58 @@ default int createOrUpdate(SelectorDTO selectorDTO) {
...
     /**
      * find selector by name.
      *
      * @param name the name
      * @return selector do
+     * @deprecated sice 2.6.0  Deprecated. By querying under this condition, multiple data are usually obtained.
+     *              Therefore, it is recommended to: {@linkplain SelectorService#findListByName(java.lang.String)}
      */
+    @Deprecated
     SelectorDO findByName(String name);
```
lianjunwei, line 132: The findByName() method should return multiple entries, and the original usage of accepting one entry is incorrect. So I have marked @ Deprecated (not recommended for others to use and minimize inappropriate use).
At present, only one place has been called (`org.apache.shenyu.admin.listener.zookeeper.HttpServiceDiscovery # updateSelectorHandler`), and HttpServiceDiscovery has long been marked  @ Deprecated by others. Although this class may not have been actually called, I still handle compatibility. #4928

shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/ApiServiceImpl.java
lianjunwei: Hi, @dragon-zhang , it has been adjusted. Is this okay?
image

lianjunwei: hi, @dragon-zhang .The original meaning of 'offline' has been preserved. Simultaneously adding optimization: When modifying apidoc, it supports deleting registered apis, but it is not supported when creating.

dragon-zhang (approved):

moremind (approved):

PR #4941, opened 2023-08-01
[fix:bug] Fix admin sync bug

yunlongn:
Make sure that:


shenyu-admin/src/main/java/org/apache/shenyu/admin/service/converter/AbstractSelectorHandleConverter.java
```diff
@@ -68,7 +68,7 @@ public String handler(final String handle, final List<CommonUpstream> aliveList)
      */
     @Override
     public <T extends CommonUpstream> List<T> updateStatusAndFilter(final List<T> existList, final List<? extends CommonUpstream> aliveList) {
-        if (CollectionUtils.isEmpty(aliveList) || CollectionUtils.isEmpty(existList)) {
+        if (aliveList == null) {
```
moremind, line 71: use CollectionUtils.isEmpty(?

moremind (changes requested):

shenyu-common/src/main/java/org/apache/shenyu/common/dto/convert/selector/DubboUpstream.java
```diff
@@ -246,8 +246,6 @@ public boolean equals(final Object o) {
         }
         DubboUpstream that = (DubboUpstream) o;
         return port == that.port
-                && weight == that.weight
-                && warmup == that.warmup
```
dragon-zhang, line 250: why remove these ?

yunlongn: These two are not the key factors. See `DivideUpstream#equals`

shenyu-admin/src/main/java/org/apache/shenyu/admin/service/converter/AbstractSelectorHandleConverter.java
yunlongn: If the aliveList is an empty array, follow-up logic is required

moremind: Objects.isNull

yunlongn: i will modify him

dragon-zhang (changes requested):
`hashCode` also need modify.

dragon-zhang (approved):

moremind (approved):

PR #4943, opened 2023-08-01
[type:feat] Increase the default stack size of shenyu-admin thread in docker

xcsnx:
Make sure that:


xcsnx:
> Why do we modify the `-Xss` parameter, can you provide more information?


Because it needs to adapt to the Euler system.
When docker runs the shenyu-admin image on the Euler system, the following error will appear "The stack size specified is too small, Specify at least 328k".

JooKS-me (approved):

PR #4944, opened 2023-08-02
[type:fix] Polish Code.

GOODBOY008:
Polish code


shenyu-admin/src/main/java/org/apache/shenyu/admin/service/manager/impl/SwaggerDocParser.java
```diff
@@ -317,7 +317,7 @@ protected RefInfo getResponseRefInfo(final JsonObject docInfo) {
 
     private RefInfo getRefInfo(final JsonObject jsonObject) {
         JsonElement refElement;
-        boolean isArray = Objects.isNull(jsonObject.get("type")) ? false : "array".equals(jsonObject.get("type").getAsString());
+        boolean isArray = !Objects.isNull(jsonObject.get("type")) && "array".equals(jsonObject.get("type").getAsString());
```
dragon-zhang, line 320: maybe use `Objects.nonNull`.

Issue #5105, opened 2023-09-01
[Question] shenyu-spring-boot-starter pom, why are there multiple plugin dependencies?


xuziyang:
### Question

Isn't it up to the user to decide what plugin to use?

crudboy:
Maybe this can answer your question?
https://shenyu.apache.org/docs/deployment/deployment-custom/#

PR #5106, opened 2023-09-01
[ISSUE #5105] Remove plugin dependencies in shenyu-spring-boot-starter

xuziyang:


shenyu-bootstrap/pom.xml
```diff
@@ -208,6 +208,52 @@
         </dependency>
         <!--shenyu cache plugin end-->
 
+        <!--if you use http proxy start this-->
+        <dependency>
+            <groupId>org.apache.shenyu</groupId>
```
moremind, line 213: this is default plugin

moremind (changes requested):

shenyu-bootstrap/pom.xml
xuziyang: All right.

PR #5197, opened 2023-10-09
alert report support config admin cluster multi servers

tomsun28:
alert report support config admin cluster multi servers

```
shenyu:
  alert:
    enabled: true
    admins: localhost:9095,182.33.4.2:9095
```

Make sure that:

- [ ] You have read the [contribution guidelines](https://shenyu.apache.org/community/contributor-guide).
- [ ] You submit test cases (unit or integration tests) that back your changes.
- [ ] Your local test passed `./mvnw clean install -Dmaven.javadoc.skip=true`.

PR #5207, opened 2023-10-11
shenyu admin register itself and others can call admin api through gateway

tomsun28:
- shenyu admin register itself and others can call admin api through gateway divide plugin
- alert report support config admin cluster register gateway api path

https://github.com/apache/shenyu/pull/5197 this support:  alert report support config admin cluster multi servers

```
shenyu
  alert:
    enabled: true
    # the shenyu admin servers or gateway path switch one to config
    admins: localhost:9095
    # the shenyu admin servers or gateway path switch one to config
    gateway: localhost:9195/shenyu/admin/alert/report
```

Make sure that:

### DIFF
--- a/shenyu-admin/pom.xml
+++ b/shenyu-admin/pom.xml
@@ -313,6 +313,12 @@
             <artifactId>shenyu-discovery-zookeeper</artifactId>
             <version>${project.version}</version>
         </dependency>
+
+        <dependency>
+            <groupId>org.apache.shenyu</groupId>
+            <artifactId>shenyu-spring-boot-starter-client-springmvc</artifactId>
+            <version>${project.version}</version>
+        </dependency>
     </dependencies>
 
     <profiles>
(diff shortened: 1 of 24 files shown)


shenyu-admin/pom.xml
```diff
@@ -313,6 +313,12 @@
             <artifactId>shenyu-discovery-zookeeper</artifactId>
             <version>${project.version}</version>
         </dependency>
+
+        <dependency>
+            <groupId>org.apache.shenyu</groupId>
+            <artifactId>shenyu-spring-boot-starter-client-springmvc</artifactId>
+            <version>${project.version}</version>
```
lianjunwei, line 320: This solution of introducing client jars feels too heavy and not elegant. It should be better to use a listener (such as org. apache. shenyu. admin. listener. Application Start Listener) to directly call the internal method to initiate the registration process after the admin is started.

lianjunwei: After thinking about it, the "sdk+annotation" solution is perfect.

tomsun28:
close. do not call shenyu admin api through shenyu gateway

PR #5241, opened 2023-10-19
Fix flaky

wang3820:
### Flakiness in `org.apache.shenyu.admin.service.PermissionServiceTest#testGetPermissionMenu`
**Nondex** was used to check and locate the flakiness in the test. The test can be reproduced using the following command:
```shell
mvn install -pl shenyu-admin -am -DskipTests
mvn -pl shenyu-admin test -Dtest=org.apache.shenyu.admin.service.PermissionServiceTest#testGetPermissionMenu
mvn -pl shenyu-admin edu.illinois:nondex-maven-plugin:2.1.1:nondex -Dtest=org.apache.shenyu.admin.service.PermissionServiceTest#testGetPermissionMenu
``` 

### Issue
This test is straight forward. It creates an expected result and compare it to the actual result from testing. The issue lies in side of the `getPermissionMenu` function, which internally calls `ResourceUtil.buildMenu` and it uses hashmap to build a list: 
```java
Map<String, PermissionMenuVO.MenuInfo> menuInfoMap = metaList.stream()
                .map(PermissionMenuVO.MenuInfo::buildMenuInfo)
                .filter(menuInfo -> Objects.nonNull(menuInfo) && StringUtils.isNotEmpty(menuInfo.getId()))
                .collect(Collectors.toMap(PermissionMenuVO.MenuInfo::getId, Function.identity(), (value1, value2) -> value1));
```
`Collectors.toMap` uses hashmap which creates the nondeterminism, as when `assertThat(result, is(expectedResult))` is called, the order of values in `result.getCurrentAuth()`. `result.getAllAuth()`, `result.getMenu()` could be nondeterministic while the order of lists in `expectedResult` is fixed.
```java
public static <T, K, U> Collector<T, ?, Map<K, U>> toMap(Function<? super T, ? extends K> keyMapper, Function<? super T, ? extends U> valueMapper, BinaryOperator<U> mergeFunction) {
        return toMap(keyMapper, valueMapper, mergeFunction, HashMap::new);
    }
```

## Fix
The order of the lists in the expected result is fixed but the the order of the lists in the actual result is not. Although the orders are different, the content are still the same. As each object contains 3 lists only, `containsInAnyOrder` is used to see if the content of each 3 lists match.

Make sure that:


li-keguo:
Excellent submission instructions

li-keguo (approved):

PR #5422, opened 2024-01-26
[type:feature] peak ewma lb

impactCn:
Make sure that:


kimmking (dismissed):

kimmking:
fix ci first @impactCn

kimmking:
The main logic is the same with P2C.

Aias00:
Review note (PMC Aias00): open since Jan 2024 and now conflicts with master. Given its age, please rebase if you intend to pursue it, or we will close it as stale.

Aias00 (changes requested):
## Review (PMC Aias00) — `[type:feature] peak ewma lb`

**Decision: REQUEST CHANGES**

Thanks for contributing a Peak EWMA load balancer — the algorithm and the Finagle/Dubbo attribution in the Javadoc are good. But this PR has gone stale against master and needs a refresh before it can merge.

### 🔴 Blocking — does not compile against current master

`PeakEWMALoadBalancer` implements

```java
@Override
protected Upstream doSelect(List<Upstream> upstreamList, String ip) {
```

but `AbstractLoadBalancer` now declares a *different* abstract signature:

```java
protected abstract Upstream doSelect(List<Upstream> upstreamList, LoadBalanceData data);
```

There is no `doSelect(List, String)` overload left, so the `@Override` fails and `shenyu-loadbalancer` will not build. The SPI signature changed (`String ip` → `LoadBalanceData data`) some time after this PR was opened; please rebase and migrate the implementation. (`LoadBalanceData` carries the ip, so the algorithm itself barely changes.)

Note this also explains why the PR currently reports **no CI checks at all** — nothing has run against it.

### 🟠 Should fix

2. **No tests.** Every other load balancer in `shenyu-loadbalancer` has a unit test (`RoundRobinLoadBalancerTest`, `HashLoadBalancerTest`, `P2cLoadBalancerTest`, `ShortestResponseLoadBalancerTest`, …). A new SPI needs at least: selection returns a member of the input list, penalty behaviour on first probe, and convergence when one upstream's latency spikes.
3. **`upstreamMetricMap` is unbounded and never evicted.**
   ```java
   private Map<Upstream, Metric> upstreamMetricMap = new ConcurrentHashMap<>();
   ```
   Upstream instances are recreated on every selector/rule sync, so this grows without bound for the life of the JVM. ShenYu has been actively bounding its caches (see #7158, #7159); please use a bounded map (e.g. `WindowTinyLFUMap`), key on something stable like the upstream URL, and/or prune entries not present in the current `upstreamList`.
4. **`getCost()` holds a lock without `try/finally`.**
   ```java
   ewmaLock.lock();
   observe();
   ...
   ewmaLock.unlock();
   ```
   If `observe()` (or `upstream.isHealthy()`) ever throws, the lock is never released and that upstream is dead forever. Wrap in `try { ... } finally { ewmaLock.unlock(); }`.
5. **`containsKey` → `get` → `put` is not atomic**; two threads can build two `Metric`s for the same upstream and one silently overwrites the other, discarding its EWMA history. Use `upstreamMetricMap.computeIfAbsent(upstream, Metric::new)`. Also make the field `final`.
6. `Metric.upstream` and `ewmaLock` should be `private final`; `upstream` is read from `observe()` under the lock but is never reassigned.

### 🟡 Non-blocking

7. The SPI resource file still has no trailing newline after the new `peakEWMA=...` line — worth fixing while you are editing it.
8. `DECAY_TIME = 600` is hard-coded. Consider making it configurable via the rule handle, consistent with how P2C exposes its parameters.

Once it compiles against the current `doSelect` signature and has tests, happy to take another look — the algorithm itself looks right.

PR #5632, opened 2024-08-27
set default extPlugin path to relative path.

initialdmg:
Current default value of extPlugin path `/ext-lib/` is not consistent with described in deployment document，which resulting custom plugin jars auto loading failed. On Windows, bootstrap binary package's startup context  is `bin` directory, ext plugin relative path is not correct too.

Make sure that:


shenyu-web/src/main/java/org/apache/shenyu/web/loader/ShenyuPluginPathBuilder.java
```diff
@@ -30,7 +30,7 @@ public final class ShenyuPluginPathBuilder {
 
     private static final String PLUGIN_PATH = "plugin-ext";
     
-    private static final String DEFAULT_EXT_PLUGIN_PATH = "/ext-lib/";
+    private static final String DEFAULT_EXT_PLUGIN_PATH = "ext-lib/";
```
yu199195, line 33: Why can't  use '/ ' at the beginning?

initialdmg: The absolute path of `/ext-lib/` is not in the deployment directory. On Linux, it's just pointing to the `/ext-lib` folder. On Windows it is in the root of the deployment driver. A log point added to confirm this. We can find this in test case too: 
https://github.com/apache/shenyu/blob/1f5b5cda230c10b577712918db95c4028d902345/shenyu-web/src/test/java/org/apache/shenyu/web/loader/ShenyuPluginPathBuilderTest.java#L42-L49

Aias00:
Review note (PMC Aias00): open since Aug 2024. If this relative extPlugin-path change is still wanted, please rebase onto current master and request a re-review; otherwise it will be closed as stale.

Aias00 (approved):
## Review (PMC Aias00) — `set default extPlugin path to relative path`

**Decision: APPROVE**

Small, coherent fix. Making the default ext-plugin path relative instead of absolute-rooted is the right call: `/ext-lib/` only ever resolved correctly if the process had write access to (and actually used) the filesystem root — on Windows it silently becomes `<drive>:\ext-lib`.

### Correctness checks

**1. `start.bat` changes are behaviour-preserving.**
```
 cd %~dp0
+cd ..
-set LOG_HOME=%~dp0/../logs
+set LOG_HOME=./logs
-set CLASS_PATH=".;..\conf;..\lib\*;..\ext-lib\*"
+set CLASS_PATH=".;.\conf;.\lib\*;.\ext-lib\*"
```
After the added `cd ..` the working directory is the dist root, so:
- `./logs` ≡ previous `%~dp0/../logs` → both `<dist>/logs` ✓
- `.\conf` / `.\lib\*` / `.\ext-lib\*` ≡ previous `..\conf` / `..\lib\*` / `..\ext-lib\*` ✓

It also makes Windows consistent with `start.sh`, which has always done `cd \`dirname $0\`; cd ..`.

**2. The relative `ext-lib/` resolves correctly in every supported launch path.** I checked all three:
- `start.sh` — `cd ..` → CWD is the dist root, `ext-lib/` → `<dist>/ext-lib` ✓
- `start.bat` (after this PR) — same ✓
- `docker/entrypoint.sh` — already relies on `DEPLOY_DIR=$(pwd)` and `./conf`, so CWD is the dist root ✓

and all of them put `${DEPLOY_DIR}/ext-lib` on the classpath, so the resolved directory is the one that actually holds the jars.

**3. `DEFAULT_EXT_PLUGIN_PATH` change is safe.** `"ext-lib/"` now falls through `getPluginPath` to `new File("ext-lib/")`, resolved against the dist root. The new `LOG.info("shenyu extPlugin path: {}", ...)` in `ShenyuLoaderService` is a genuinely useful startup diagnostic for exactly this class of bug, and `getPluginFile` has no side effects, so calling it once for logging is harmless.

### Non-blocking nits

1. `ShenyuPluginPathBuilder` now calls
   ```java
   ShenyuPluginPathBuilder.class.getResource(DEFAULT_EXT_PLUGIN_PATH);
   ```
   with a **relative** name. `Class#getResource` resolves non-`/`-prefixed names against the *class's package*, so this now looks up `org/apache/shenyu/web/loader/ext-lib/`, which will never exist — the lookup always returns `null` and falls through to the literal default. Behaviour is unchanged in practice (the old absolute `/ext-lib/` almost never resolved either), but the intent would be clearer as `getResource("/" + DEFAULT_EXT_PLUGIN_PATH)` or via the classloader. Worth a follow-up.
2. The two `start.bat` files duplicate this logic; nothing to do now, just noting it.

### CI

The only failing check is `build (shenyu-integrated-test-sdk-http)` from run `27319442694`, which is dated **2026-06-11** — three months stale and its logs have expired, so it cannot be diagnosed. Most sibling jobs in that run were `CANCELLED`. Please push a trivial commit (or re-run) so we get a current green signal before merge; I don't believe the failure relates to this change, but it needs confirming.

Nice small fix — thanks.

Issue #5529, opened 2024-04-02
[Task]  Add end-to-end (E2E) testing for logging plugins.


xcsnx:
### Description

Currently, there are many logging plugins in Shenyu that lack end-to-end (E2E) testing. It is necessary to improve the E2E testing for these logging plugins now. You can refer to the E2E testing of the RocketMQ logging plugin: https://github.com/apache/shenyu/pull/5439

### Task List

- [ ] 1. Logging-ElasticSearch Plugin
- [ ] 2. Logging-RabbitMq Plugin
- [ ] 3. Logging-Kafka Plugin
- [ ] 4 .Logging-Pulsar Plugin
- [ ] 5. Logging-Clickhouse Plugin
- [ ] 6. Logging-RocketMq Plugin (Finished)

RayayChung:
pls assign logging-kafka to me

Divyansh200102:
I would like to try

HsiuChi:
Please assign part 1,2 to me

ling0900:
Please assign one or more to me, I would like to do

zzzk1:
Hi @xcsnx, can you assign `Logging-Kafka Plugin` part to me?

xcsnx:
> Hi @xcsnx, can you assign `Logging-Kafka Plugin` part to me?

Certainly, but it seems that someone has already attempted to work on Kafka, yet there are still some issues. If you're interested, you can build upon their work to achieve the goal. For more details, please refer to the PR: https://github.com/apache/shenyu/pull/5540.

478320:
pls assign part 2 to me

kamiingithub:
Hi @xcsnx, pls assign part 1 and 5 to me

no-regret666:
Hello, @xcsnx ,i would like to try part 1,can you assign it to me?

xcsnx:
> Hello, [@xcsnx](https://github.com/xcsnx) ,i would like to try part 1,can you assign it to me?


of course

wjwang00:
Hi @xcsnx ,If Task 4: Logging-Pulsar Plugin is not yet completed, I would like to try to finish it.

SouthwestAsiaFloat:
Hello @xcsnx ,if task 5 is not assigned at present, I'd like to take this task. Very appreciate if you could let me have this on me 😊

PR #5645, opened 2024-08-31
[type:feat] add Logging-Kafka Plugin

jakiuncle:
https://github.com/apache/shenyu/issues/5529


Aias00 (changes requested):
if you want to add new e2e test, you should add a new module, not change exist modules

Issue #5623, opened 2024-08-19
[Task] ShenYu namespace development


xcsnx:
### Description

Currently, managing multiple user requirements with ShenYu often involves deploying multiple ShenYu Admin instances and sets of ShenYu Gateway. To ensure data independence, each gateway typically connects to only one ShenYu Admin. However, managing multiple ShenYu Admin instances increases both user and operational costs.

With ShenYu Admin's namespace feature, users can deploy a single ShenYu Admin cluster to manage multiple ShenYu Gateways. The namespace management allows for better control of ShenYu Gateway and ShenYu Client data registration, reducing both user and operational costs.

#### 1.Shenyu-Admin:
- [x] 1.1 Add namespace module @xcsnx 
- [x] 1.2 Transform the plugin module into the plugin template module @xcsnx 
- [x] 1.3 Add a new plugin management module (plugin templates as instances under namespaces) @xcsnx 
- [x] 1.4 Modify the selector, rule, and metadata modules @xcsnx 
- [x] 1.5 Modify the discovery modules @xcsnx 
- [x] 1.6 Modify the alert modules @478320 
- [x] 1.7 Modify the auth modules @xcsnx 
- [x] 1.8 Modify the apiDoc modules（Wait for the version 1.4 task to be completed.）@478320 
- [x] 1.9 Modify the data-permission modules @xcsnx 

#### 2.Shenyu-client:
- [x] 2.1 Modify the shenyu-client @moremind 
#### 3.Sync data :
- [x] 3.1 Modify Http  @xcsnx 
- [x]  3.2 Modify Websocket  @Aias00 
- [x] 3.3 Modify PathDataSync @moremind 
- [x] 3.4 Modify NodeDataSync @moremind

xcsnx:
Modifications to the Shenyu-Admin module can refer to: https://github.com/apache/shenyu/pull/5604

moremind:
pls assign to me middleware sync

478320:
Please assign part 1.8 to me

xcsnx:
> Please assign part 1.8 to me @xcsnx

This subtask is under discussion, and the transformation plan has not yet been determined.

478320:
> > Please assign part 1.8 to me @xcsnx
> 
> This subtask is under discussion, and the transformation plan has not yet been determined.

Thank you for your response. I understand that subtask 1.8 is still under discussion. If there's anything I can help with, please let me know. Thank you!

xcsnx:
[...]

PR #5685, opened 2024-09-29
[type:refactor]Decouple apidoc module from rule and selector modules

478320:
decouple apidoc module from rule and selector modules, link https://github.com/apache/shenyu/issues/5623

Make sure that:
 - [X] Your local test passed ./mvnw clean install -Dmaven.javadoc.skip=true.


shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/ApiServiceImpl.java
```diff
@@ -110,6 +109,7 @@ public String createOrUpdate(final ApiDTO apiDTO) {
      * @return update message
      */
     private String update(final ApiDTO apiDTO) {
+        //cdbuild apiDO
```
yu199195, line 112: why add this？

shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/ApiServiceImpl.java
```diff
@@ -126,11 +126,6 @@ private String update(final ApiDTO apiDTO) {
                 tagRelationMapper.deleteByApiId(apiDO.getId());
                 tagRelationMapper.batchInsert(tags);
             }
-            if (ApiStateEnum.PUBLISHED.getState() == apiDO.getState()) {
```
yu199195, line 129: why remove this？

478320: Hello, my previous submission failed to check the instability of ci. I hope I can re-trigger ci through some modifications. If these modifications cause bad effects, I will delete them

shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/ApiServiceImpl.java
478320: Hello, my previous submission failed to check the instability of ci. I hope I can re-trigger ci through some modifications. If these modifications cause bad effects, I will delete them

shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/ApiServiceImpl.java
478320: Hello, in order to realize that the user can directly use the api document after adding it in the visual operation interface, shenyu Gateway creates the default selector and rule when calling to create the api document, so the user can directly use the document for testing. But this also causes the api document to be coupled with selector and rule. To delete this part, users need to manually create the api document and test the need to manually add selector, rule, metadate, but solve the coupling problem of api and selector and rule. Users who use annotations such as @ShenyuGetMapping, as well as those who use swagger, will not be affected by the automatic creation of selectors and so on

478320: When creating apidoc, this part determines whether the document is in the published state, and if the document is in the published state, the selector is registered, etc. If it is in the offline state, the corresponding selector content is deleted, which will lead to the coupling of apidoc and selector. Deleting this part can release their coupling. The removal of this feature will have the following effects, the document publication and logout will only be displayed as a status without additional functions, the document logout and online need to manually add and delete selectors and other content, for the creation of api documents, Users using @RequestMapping and other annotations and users using swagger will automatically create selectors and other content through annotations, which will not affect the convenience brought by automation. To add api documents in the visual interface, you need to manually create corresponding selector, rule, and Metadata. For the offline api document, all users need to manually delete the corresponding selector, rule, and Metadata

moremind (approved):

Issue #5780, opened 2024-11-13
[BUG] Fix the issue with deleting the "Host" header


titxue:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Directly deleting the "Host" header causes the custom host in the requests plugin to fail.

![image](https://github.com/user-attachments/assets/31e50dbd-4668-4068-97e4-94171e8e4fd9)
![image](https://github.com/user-attachments/assets/0efd44d2-6adf-4ee9-a847-6daf864f9fc6)

### Expected Behavior

The "Host" header should be customizable, rather than being restricted to the reverse proxy address.

![image](https://github.com/user-attachments/assets/cb15e30b-f700-4865-85e9-56249d330b39)

### Steps To Reproduce

1. Go to **HttpProcess** → **Request** → **Rules** → **Headers**.
2. Adding a custom **Host** value to the headers does not take effect. 

### Environment

```markdown
ShenYu version(s):2.6.1
```

### Anything else?

[The Host field is not set correctly when the upstream is a reverse proxy.](https://github.com/apache/shenyu/issues/4381)

moremind:
In your second picture, why does your host have to be github.com? please add my wechat: yijia1949

JerryDtj:
能把这个问题指派给我吗,虽然我时间不多,但是我想试试

PR #5781, opened 2024-11-13
Fix the issue with deleting the "Host" header

titxue:
#5780


yu199195:
why delete 'Host'  from  http header

titxue:
> why delete 'Host' from http header

titxue:
> why delete 'Host' from http header

The original logic ends up removing the ‘Host’ header, which causes any custom ‘Host’ header functionality in the request plugin to fail. My recommendation is to remove the code that deletes ‘Host’ to allow the plugin to function correctly, increasing flexibility without forcing the removal of the ‘Host’ header. Alternatively, this could be made configurable, either in the request setup or the service discovery. In any case, forcing the deletion of ‘Host’ in code is not ideal. Please also refer to the details in my submitted issue for further clarification.

titxue (commented):
Remove the delete host operation

moremind:
not right

PR #5799, opened 2024-11-21
[type:bugfix] fix namespace bugs

Aias00:
fix namespace bugs

Make sure that:

### DIFF
--- a/shenyu-admin/src/main/java/org/apache/shenyu/admin/controller/NamespacePluginController.java
+++ b/shenyu-admin/src/main/java/org/apache/shenyu/admin/controller/NamespacePluginController.java
@@ -55,7 +55,7 @@ import java.util.List;
 /**
  * this is namespace plugin controller.
  */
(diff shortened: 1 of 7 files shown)


shenyu-admin/src/main/java/org/apache/shenyu/admin/controller/NamespacePluginController.java
```diff
@@ -55,7 +55,7 @@
 /**
  * this is namespace plugin controller.
  */
-@RestApi("/namespacePlugin")
+@RestApi("/namespace_plugin")
```
dengliming, line 58: You should keep consistent with other places. e.g. `namespace-plugin` would be better.

Aias00: sounds good

yu199195 (approved):

yu199195 (approved):

Issue #5844, opened 2024-12-11
[BUG] <The nacos data proxy.selector and discovery.divide for divide selector delete failed when the divide selector be deleted>


Wweiei:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

In divide plugin, the selector's selector config、proxy_selector and discovery config will be sync to nacos, when use nacos as data sync center. When delete the selector, the relevant information for the selector should be delete at the same time. But currently, only the selector config has been removed, the proxy_selector and discovery config have not been deleted.
1、add one selector as follows, the selector id is 1866733538682441728.
![image](https://github.com/user-attachments/assets/f9b32256-a37f-425e-b659-f4d7b0af4a4d)
![image](https://github.com/user-attachments/assets/0154afa5-b823-44de-835a-fc70fab1bc3c)
2、nacos data.
![image](https://github.com/user-attachments/assets/3b0138a7-4aa0-4d0c-bbdf-55a9ee7d9caa)
3、delete selector, the proxy.selector config and discovery.divide config still exists.
![image](https://github.com/user-attachments/assets/b43281d1-fac8-4d41-b870-64f5b2688c45)

### Expected Behavior

The divide selector's proxy.selector and discovery config should be deleted when the selector is deleted.

### Steps To Reproduce

1. check out to master branch and pull the lastest code.
2. use nacos as data sync center.
3. add one divide selector.
4. delete the selector.

### Environment

```markdown
ShenYu version(s): branch master latest
```

### Anything else?

1. The proxy.selector is deleted failed is due to the ProxySelectorDTO missing the id value and pluginName set the wrong value when unbindDiscovery .
2. When deleting the selector, there is no method to delete the discovery configuration.

yu199195:
good job～

PR #5845, opened 2024-12-11
[type:fix] Delete proxy.selector and discovery  nacos config when delete the divide selector.

Wweiei:
Fix the issue #5844. Delete proxy.selector and discovery  nacos config when delete the divide selector.

Make sure that:

### DIFF
--- a/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/SelectorServiceImpl.java
+++ b/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/SelectorServiceImpl.java
@@ -294,12 +294,14 @@ public class SelectorServiceImpl implements SelectorService {
             if (Objects.nonNull(discoveryDO)) {
                 final DiscoveryProcessor discoveryProcessor = discoveryProcessorHolder.chooseProcessor(discoveryDO.getType());
                 ProxySelectorDTO proxySelectorDTO = new ProxySelectorDTO();
+                proxySelectorDTO.setId(selector.getId());
                 proxySelectorDTO.setName(selector.getName());
-                proxySelectorDTO.setPluginName(pluginMap.getOrDefault(selector.getId(), ""));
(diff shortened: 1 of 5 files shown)


shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/SelectorServiceImpl.java
```diff
@@ -294,12 +294,14 @@ private void unbindDiscovery(final List<SelectorDO> selectors, final List<Plugin
             if (Objects.nonNull(discoveryDO)) {
                 final DiscoveryProcessor discoveryProcessor = discoveryProcessorHolder.chooseProcessor(discoveryDO.getType());
                 ProxySelectorDTO proxySelectorDTO = new ProxySelectorDTO();
+                proxySelectorDTO.setId(selector.getId());
                 proxySelectorDTO.setName(selector.getName());
-                proxySelectorDTO.setPluginName(pluginMap.getOrDefault(selector.getId(), ""));
+                proxySelectorDTO.setPluginName(pluginMap.getOrDefault(selector.getPluginId(), ""));
```
yu199195, line 299: why is  pluginMap.getOrDefault(selector.getPluginId(), "")？

Wweiei:
```
Map<String, String> pluginMap = ListUtil.toMap(pluginDOS, PluginDO::getId, PluginDO::getName);
```
The key for pluginMap is pluginId，the value for pluginMap is pluginName.
At this step the pluginMap's value is the divide plugin's id as key and the divide plugin'name as the value.
image
If use the selectorId as key, it will get null value, instead of "divide".

yu199195 (approved):

shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/SelectorServiceImpl.java
yu199195: thanks, great job.

PR #5951, opened 2025-03-07
[feat] add jdbc properties

Aias00:
add jdbc properties, for mysql 8+ 
image

Make sure that:

### DIFF
--- a/shenyu-admin/src/main/resources/application-mysql.yml
+++ b/shenyu-admin/src/main/resources/application-mysql.yml
@@ -20,7 +20,7 @@ shenyu:
 
 spring:
   datasource:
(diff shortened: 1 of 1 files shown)


yu199195:
mysql 5.7，  need add this properties？

Aias00:
> mysql 5.7， need add this properties？

Setting allowPublicKeyRetrieval=true does not cause errors when connecting to a MySQL 5.7 server. This parameter is necessary when the server uses sha256_password and SSL is not enabled, as it prevents the "Public Key Retrieval is not allowed" error. If mysql_native_password is used, the parameter is ignored and does not cause any issues. Users should adjust this setting based on the actual authentication plugin and SSL configuration, but overall, **this setting is safe.**

yu199195 (approved):

yu199195 (approved):

PR #5989, opened 2025-04-01
[fix] fix http check

Aias00:
fix http check
fix when set checkd = false, still check
image

Make sure that:


moremind (dismissed):

Aias00:
@moremind  i did somt change, it is my bad. the default value should return false, or the upstream data will not be recorded

PR #6072, opened 2025-07-28
optimize: organize nacos dependencies to infra nacos module

yuluo-yx:
使用注解 `@NestedConfigurationProperty` 组合 nacos properties 配置


yuluo-yx:
refer #6082

PR #6140, opened 2025-09-06
Fix the issue where the cache is not properly deleted after removing a selector in the Nacos data synchronization method.

### LINKED ISSUE #6121: [BUG] <title>Using Nacos data synchronization, after deleting the selector, the data cannot be synchronized to the gateway

### Current Behavior

After deleting the selector, the data cannot be synchronized to the gateway, resulting in an error that the rule does not exist. Restarting the application service to automatically register with admin is also useless, and the gateway issue must be resolved by restarting it

### Expected Behavior

_No response_

### Steps To Reproduce

1. Delete the selector in the divide plugin plug-in of shenyu admin, and then call the application service to report the following error: divide: rule not found
2. Restart the application service and automatically register with shenyu admin. Both the selector and rule registration are normal, but there are still errors when accessing the application service, divide：rule not found
3. Restart shenyu gateway to solve the problem

### Environment

```markdown
ShenYu version(s):2.7.0.1，2.7.0.2
```

### Debug logs

_No response_

### Anything else?

_No response_

17661152:
Fixes #6121

I have implemented the fix on the 2.7.0.2 version used in our company and have already deployed it to the production environment.


Aias00 (approved):
LGTM，thx

PR #6263, opened 2025-12-27
Fix: Set isClosing flag in closeGracefully method as documented

airajena:
## Description
This PR fixes a FIXME comment in `ShenyuSseServerTransportProvider.java`. The javadoc for `closeGracefully()` claimed to use the `isClosing` flag to prevent new connections during shutdown, but the flag was never actually being set.

## Problem
The FIXME comment (line 196-198) noted:
```java
// FIXME: This javadoc makes claims about using isClosing flag but it's not
// actually
// doing that.
```

The `closeGracefully()` method was not setting `isClosing = true`, which means new connections could potentially be accepted during the graceful shutdown process.

## Solution
- Set `isClosing = true` at the start of `closeGracefully()` method
- Update the javadoc to accurately describe the behavior
- Remove the outdated FIXME comment

## Changes
**File:** `shenyu-plugin/shenyu-plugin-mcp-server/src/main/java/org/apache/shenyu/plugin/mcp/server/transport/ShenyuSseServerTransportProvider.java`

```java
// Before
public Mono<Void> closeGracefully() {
    return Flux.fromIterable(sessions.values())
        ...
}

// After
public Mono<Void> closeGracefully() {
    isClosing = true;  // Now properly set!
    return Flux.fromIterable(sessions.values())
        ...
}
```

## Testing
- ✅ `./mvnw compile -pl shenyu-plugin/shenyu-plugin-mcp-server -am -DskipTests` passes

## Impact
- **Risk:** Low - This is a bug fix that ensures proper shutdown behavior
- **Backwards Compatible:** Yes - No API changes

### DIFF
--- a/shenyu-disruptor/src/main/java/org/apache/shenyu/disruptor/DisruptorProviderManage.java
+++ b/shenyu-disruptor/src/main/java/org/apache/shenyu/disruptor/DisruptorProviderManage.java
@@ -130,7 +130,7 @@ public class DisruptorProviderManage<T> {
         disruptor.setDefaultExceptionHandler(new IgnoreExceptionHandler());
         disruptor.start();
         RingBuffer<DataEvent<T>> ringBuffer = disruptor.getRingBuffer();
-        provider = new DisruptorProvider<>(ringBuffer, disruptor, isOrderly);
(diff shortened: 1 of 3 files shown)


Aias00 (approved):
LGTM

Issue #6265, opened 2025-12-30
[BUG] When setting RealURL in configurationShenyuContext, parameter information cannot be added.


MaMengzhen:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

In the configurationShenyuContext method, when setting the RealURL, parameter information cannot be added and the original path information should be set.

<img width="2586" height="1946" alt="Image" src="https://github.com/user-attachments/assets/908df535-0f25-4948-bfb6-ecfa0a0d32e3" />

<img width="2574" height="1930" alt="Image" src="https://github.com/user-attachments/assets/98549517-18b7-4209-9d2b-2bfe8069d5ff" />

### Expected Behavior

In the configurationShenyuContext method, when setting the RealURL, parameter information cannot be added and the original path information should be set.

### Steps To Reproduce

1. Convert an existing GET request interface into an MCP tool.
2. Test the MCP tool.
3. Enter the configurationShenyuContext method and set the RealURL.
4. Entering the buildRequestUri method, Build the final request uri.

### Environment

```markdown
ShenYu version(s):2.7.1-SNAPSHOT
```

PR #6266, opened 2025-12-30
fix mcp config shenyu context error

MaMengzhen:
In the configurationShenyuContext method, when setting the RealURL, parameter information cannot be added and the original path information should be set. #6265 

Make sure that:

### DIFF
--- a/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/UpstreamCheckService.java
+++ b/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/UpstreamCheckService.java
@@ -222,6 +222,8 @@ public class UpstreamCheckService {
             if (!exists.isPresent()) {
                 upstreams.add(commonUpstream);
             } else {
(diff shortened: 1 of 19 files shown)


shenyu-plugin/shenyu-plugin-mcp-server/src/main/java/org/apache/shenyu/plugin/mcp/server/callback/ShenyuToolCallback.java
```diff
@@ -612,6 +610,23 @@ private void configureShenyuContext(final ServerWebExchange decoratedExchange,
...
     }
 
+    /**
+     * Safely extracts raw path from a URI string, falling back to original string on failure.
+     *
+     * @param path the path string to process
+     * @return raw path if URI parsing succeeds, otherwise the original string
+     */
+    private String extractRawPath(final String path) {
+        if (org.apache.commons.lang3.StringUtils.isEmpty(path)) {
```
Aias00: do not use full name , use simple name pls

PR #6275, opened 2026-01-21
replace stream().forEach() to forEach()

VampireAchao:
replace stream().forEach() to forEach()

Make sure that:


shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/TagServiceImpl.java
```diff
@@ -184,7 +184,7 @@ private void recurseUpdateTag(final Map<String, TagDO> allData, final Map<String
             return;
         }
         List<String> subTagIds = relationMap.get(id);
-        subTagIds.stream().forEach(tagId -> {
+        subTagIds.forEach(tagId -> {
```
Aias00, line 187: ensure that `subTagIds` is not null to avoid a `NullPointerException`.

Aias00 (approved):

PR #6313, opened 2026-04-01
fix(plugin-base): short-circuit cached selector and rule matches

Aias00:
## Summary
- reuse cached selector matches in `AbstractShenyuPlugin` instead of re-running default selector matching on a cache hit
- reuse cached rule matches on a cache hit while preserving the existing empty-sentinel fallback behavior
- add `AbstractShenyuPluginTest` coverage for selector-cache-hit and rule-cache-hit short-circuit paths

## Test Plan
- [x] `./mvnw -q -pl shenyu-plugin/shenyu-plugin-base -am -Dtest=AbstractShenyuPluginTest -DfailIfNoTests=false -DskipITs -Drat.skip=true -Djacoco.skip=true -Dcheckstyle.skip=true -Dspotbugs.skip=true test`


shenyu-plugin/shenyu-plugin-base/src/main/java/org/apache/shenyu/plugin/base/AbstractShenyuPlugin.java
```diff
@@ -95,13 +95,15 @@ public Mono<Void> execute(final ServerWebExchange exchange, final ShenyuPluginCh
...
         SelectorData selectorData = obtainSelectorDataCacheIfEnabled(path);
-        // handle Selector
-        if (Objects.nonNull(selectorData) && StringUtils.isBlank(selectorData.getId())) {
-            return handleSelectorIfNull(pluginName, exchange, chain);
-        }
-        selectorData = defaultMatchSelector(exchange, selectors, path);
-        if (Objects.isNull(selectorData)) {
-            return handleSelectorIfNull(pluginName, exchange, chain);
+        if (Objects.nonNull(selectorData)) {
+            if (StringUtils.isBlank(selectorData.getId())) {
```
moremind, line 99: merge the condition

moremind (changes requested):

Issue #6071, opened 2025-07-28
[Task] add metadata map to model upstream


Aias00:
### Description

add metadata map to model upstream

### Task List

1. add metadata map to model upstream

ww783:
Hi, I'd like to work on this issue. Could you please assign it to me?

Aias00:
> Hi, I'd like to work on this issue. Could you please assign it to me?

yeap, welcome, feel free to ask if u got any problems

PR #6317, opened 2026-04-03
[ISSUE #6071] Add gray release support for divide plugin

ww783:
## Summary

* add gray release fields to `DivideRuleHandle`
* add `GrayLoadBalancer` for gray routing and traffic splitting
* add metadata support for divide discovery upstreams
* preserve admin-managed `gray` and `metadata` during discovery upstream updates

## Migration

* existing 2.7.1 deployments need to run the new `2.7.1-upgrade-2.7.2-*` script
  before starting the updated admin image

## Testing

* `./mvnw -pl shenyu-common,shenyu-loadbalancer,shenyu-plugin/shenyu-plugin-proxy/shenyu-plugin-divide -am -DskipITs -DfailIfNoTests=false -Dtest=DivideRuleHandleTest,PluginEnumTest,UpstreamCacheManagerTest,GrayLoadBalancerTest,DividePluginTest,DivideUpstreamDataHandlerTest test -Dmaven.javadoc.skip=true`
* `./mvnw -pl shenyu-admin -am -DskipTests -Dmaven.javadoc.skip=true compile`
* `./mvnw -pl shenyu-web -am -DskipTests -Dmaven.javadoc.skip=true compile`

Ref #6071


yu199195:
Hi， can you add my wechat： xixy199195 ？

damingqicai:
如何取消订阅，辛苦回复一下，谢谢






大明奇才
***@***.***



        



         原始邮件
         
       
发件人：xiaoyu ***@***.***&gt;
发件时间：2026年5月11日 11:40
收件人：apache/shenyu ***@***.***&gt;
抄送：Subscribed ***@***.***&gt;
主题：Re: [apache/shenyu] [ISSUE #6071] Add gray release support for divide plugin (PR #6317)



yu199195 left a comment (apache/shenyu#6317)

Hi， can you add my wechat： xixy199195 ？

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS&nbsp;or Android.

Aias00 (changes requested):
Dual-lane review (code-reviewer + architect) of #6317 (head `7e54812c`). Requesting changes.

**High — regresses the #6412 cache-pollution fix.** `DividePlugin.doExecute` still does `upstream.setUrl(specifyDomains.get(0))` after selecting from `UpstreamCacheManager` (cached, shared instance), so a `Specify-Domain` header permanently overwrites a cached upstream's URL for all subsequent requests. Master commit `d4db5e60d` (#6412, merged 2026-07-05) replaced exactly this with filter-first + rebuild-new-`Upstream`. This PR (last commit 2026-06-22) predates #6412, so it will regress that fix unless rebased. Please rebase onto current master and reconcile `doExecute` with #6412's handling, and decide/document the gray-vs-specifyDomain precedence (today gray runs on the full pool and the selected URL is then overwritten, so `Specify-Domain` silently bypasses gray routing). Add a gray + specifyDomain test.

**High (architect) — silent semantic regression of the existing `props.gray` exclusive-pool behavior.** Master's `DivideUpstreamDataHandler` (from #5763) submits gray upstreams as an *exclusive* pool when any `props.gray=true` exists. This PR removes that branch and always submits a mixed pool; gray upstreams only matter when the new rule handle sets `grayEnabled=true`. Any deployment relying on the existing gray flag (e.g. a custom discovery provider injecting `gray:true` into props) will silently start sending production traffic to gray upstreams under the normal LB, with no migration note. For a non-breaking 2.7.1→2.7.2, please add a compat fallback (preserve exclusive-pool behavior when `grayEnabled=false` but gray upstreams exist) or an explicit behavior-change note.

**Medium — `matchMetadata` matches ALL gray upstreams when `value` is null.** `GrayLoadBalancer.matchMetadata` returns `true` when `metadataMatch.getValue()` is null, and `DividePlugin.buildGrayConfig` builds a `MetadataMatch` whenever `grayMetadataKey` is set (value defaults to null). So a rule with the metadata key set but value blank partitions against every gray upstream, defeating version partitioning. Treat a blank value as "no metadata filter" (skip) or require both key and value at config time.

**Medium — pure percent-based canary is impossible.** `shouldRouteToGray` returns false when `conditions` is empty *before* consulting `percent`, and `grayPercent` defaults to 0, so `grayEnabled=true` + `grayPercent=50` with no condition routes 0% to gray with no warning. Either allow percent-only routing or log a warning when gray is enabled without a condition/percent>0.

**Architect design items (non-blocking but please address):**
- **Dual gray-identity model.** `GrayLoadBalancer` requires `Upstream.isGray()==true` **and** a metadata match; admin now exposes the boolean (via props) and the metadata column independently with no enforced invariant. Partial config silently degrades. Collapse to one identity model — preferably metadata-only partitioning — or derive `isGray()` from the configured metadata at sync time.
- **Abstraction/schema mismatch.** `GrayConfig` carries `List<GrayCondition>` (OR semantics) but `DivideRuleHandle` flattens to a single condition and `buildGrayConfig` emits a ≤1-element list — the list path is dead and forces a future schema migration. Either make the rule handle a `List<GrayCondition>` now, or drop the list from `GrayConfig`.
- **Layering.** Header/cookie/query/ip condition matching is reimplemented inside `GrayLoadBalancer` with a bespoke operator set, parallel to Shenyu's existing rule/selector match engine. Reuse `MatchStrategy`/`Condition`, or reconsider whether gray routing belongs in the LB at all vs. a separate selector+rule pointing at a gray upstream group.

**Low.** On retry, `DefaultRetryStrategy.resend` re-enters `GrayLoadBalancer.select` and re-rolls `ThreadLocalRandom.nextInt(100) < percent`, so a gray request may fail over to the normal pool — consider caching the route-to-gray decision in the exchange for the request lifetime. Also, gray selection is unit-tested in `GrayLoadBalancerTest` but not through the real `DividePlugin` with a mixed gray/normal pool (`doExecuteGrayModeShouldExposeGrayConfigAndLoadBalanceAttribute` only asserts the attribute is set, using non-gray-flagged upstreams) — a plugin-level test asserting the gray upstream is actually selected from a mixed pool would guard the wiring.

Positives: defensive copying in `Upstream.setMetadata`/Builder, fail-closed `matchCondition` try/catch, NULL-metadata compatibility in `DiscoveryTransfer`/`DivideUpstreamDataHandler`, and the `toString` comma fix. The retry path correctly re-uses the persisted `GrayConfig` exchange attribute.

Aias00 (changes requested):

Issue #2216, opened 2021-10-13
[Volunteer] sync data center support different ShenYu instance


qicz:
### Volunteer help

Sometimes, We just have one zookeeper or nacos cluster, but there are many ShenYu instances for different env or business, and the data is isolated. Maybe ShenYu sync data center save the data's identity with instance info.

eg：
```
shenyu
   ├──{instanceName or Instance Id}
   ├    ├──regsiter
   ├    ├     ├──metadata
   ├    ├     ├     ├──${rpcType}
   ├    ├     ├     ├      ├────${contextPath}
   ├    ├     ├     ├               ├──${ruleName} : save metadata data of MetaDataRegisterDTO
   ├    ├     ├──uri
   ├    ├     ├     ├──${rpcType}
...
```

Issue #2215, opened 2021-10-13
[Question] <title> Does it support a project to register HTTP and Dubbo interfaces to the gateway at the same time?


mobaiLv:
### Question

Does it support a project to register HTTP and Dubbo interfaces to the gateway at the same time?

dingjian2:
no  support both ,    but you could write a new starter  integration  springmvc and dubbo starter

PR #6330, opened 2026-04-17
[fix]: Fix ReactorNettyWebSocketClient builder reuse in WebSocket plugin configuration

zhaoguhong:
### Motivation

`WebSocketPluginConfiguration` currently passes a mutable `WebsocketClientSpec.Builder` in a way that may cause builder state to be reused across WebSocket handshakes.

Following the same approach discussed in Spring Cloud Gateway issue #2215 / PR #2216, each handshake should use a fresh builder instance to avoid cross-request state leakage.

### Description

- Updated `WebSocketPluginConfiguration#reactorNettyWebSocketClient` to provide a `Supplier<WebsocketClientSpec.Builder>` that creates a **new builder on each invocation**.
- Extended the existing `testReactorNettyWebSocketClient` (as requested) to verify behavior by checking that the internal builder supplier returns different builder instances across calls.

### Related Links

- https://github.com/spring-cloud/spring-cloud-gateway/issues/2215
- https://github.com/spring-cloud/spring-cloud-gateway/pull/2216


Aias00 (approved):
LGTM

PR #6339, opened 2026-05-12
feat(api-doc): support request parameter parsing for RPC types (Dubbo/SOFA/TARS/gRPC)

### LINKED ISSUE #5940: [BUG] <title>@ApiDoc注解下的API在注册时无法把请求、返回body注册进去

### Current Behavior

比如如图这个api，应该在document中得到OrderDTO、OrderDTO的结构信息，可是在页面上却是空的（如图）

Image

Image

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
ShenYu version(s):2.7.1-SNAPSHOT
```

### Debug logs

_No response_

### Anything else?

_No response_

### LINKED ISSUE #6338: [BUG] API Document generation does not parse request parameters for RPC types (Dubbo/SOFA/TARS/gRPC)

### Current Behavior

The API document generation in `ApiDocRegistrarImpl.getDocument()` uses `OpenApiUtils.generateDocumentParameters()` for all RPC types without distinction. This method only supports parsing Spring Web annotations (`@RequestParam`, `@RequestPart`, path variables) for HTTP-style parameters. For non-HTTP RPC types (Dubbo, SOFA, TARS, gRPC), it fails to extract any meaningful request parameter information, since these RPC types do not use Spring Web annotations.

As a result, the generated API document for RPC types contains `responseType` (response details) but has empty/missing request parameter details — the document only has response, no request.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
ShenYu version(s):
```

### Debug logs

_No response_

### Anything else?

_No response_

eye-gu:
close #6338 
close #5940

Previously, API document generation only supported Spring Web annotation-based parameter parsing (`@RequestParam`, `@RequestPart`, path variables), which only works for HTTP/WebSocket/Spring Cloud. RPC types (Dubbo, SOFA, TARS, gRPC) had no request parameter information in generated documents.

### Changes

**OpenApiUtils - RPC-aware document generation**
- Add `buildDocumentJson()` that dispatches by RPC type:
  - **Dubbo/TARS**: parse request parameters from Java method signatures
  - **gRPC**: parse request/response from `StreamObserver` method pattern
  - **Dubbo + Protobuf**: parse protobuf message fields via reflection for both request and response
  - HTTP/WebSocket/Spring Cloud: unchanged Spring MVC annotation parsing
- Unify document fields to `requestParameters` / `responseParameters`

**SOFA - enable API document support**
- Refactor `SofaServiceEventListener` to extend `AbstractContextRefreshedEventListener` common flow
- Replace custom `handler()` / `onApplicationEvent()` with standard `getCorrectedClass()` + `buildApiPath()` hooks
- SOFA services now go through the unified API doc registration pipeline

**Examples**
- Add `@ApiModule` / `@ApiDoc` annotations to Dubbo protobuf example services

### Impact

- Dubbo/SOFA/TARS/gRPC API documents now include detailed request parameter schemas
- SOFA module gains API document generation capability
- No breaking change for HTTP/WebSocket/Spring Cloud document generation

Make sure that:


shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/ApiServiceImpl.java
```diff
@@ -243,12 +243,14 @@ public ApiVO findById(final String id) {
...
-                apiVO.setRequestParameters(docItem.getRequestParameters());
-                apiVO.setResponseParameters(docItem.getResponseParameters());
-                apiVO.setBizCustomCodeList(docItem.getBizCodeList());
+                if (Objects.nonNull(docItem)) {
+                    apiVO.setRequestHeaders(docItem.getRequestHeaders());
+                    apiVO.setRequestParameters(docItem.getRequestParameters());
+                    apiVO.setResponseParameters(docItem.getResponseParameters());
+                    apiVO.setBizCustomCodeList(docItem.getBizCodeList());
+                }
             }
```
moremind, line 254: else {? how to resolve?}

moremind (changes requested):

shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/ApiServiceImpl.java
eye-gu: @moremind I think when document is blank or docItem is null, it means the API has no document data, so the page simply won't display any API Doc content — which is the expected behavior. The null checks (StringUtils.isNotBlank + Objects.nonNull) added in this PR already make the code safer than before. No additional else branch is needed here.

shenyu-client/shenyu-client-core/pom.xml
```diff
@@ -81,5 +81,55 @@
...
+                <groupId>kr.motd.maven</groupId>
+                <artifactId>os-maven-plugin</artifactId>
+                <version>1.6.2</version>
+            </extension>
+        </extensions>
+        <plugins>
+            <plugin>
+                <groupId>org.xolstice.maven.plugins</groupId>
+                <artifactId>protobuf-maven-plugin</artifactId>
+                <version>0.6.1</version>
```
moremind: use common version

moremind (changes requested):

shenyu-client/shenyu-client-core/pom.xml
eye-gu: Done.

Aias00 (commented):
Following up on this PR. Head consistency check first: the effective diff against master is empty — `gh pr diff 6339` and `gh api pulls/6339/files` both return nothing, and `pr-6339^{tree}` equals the master base tree `b94a6997b`. Tracing this, the PR's content was already squash-merged via #6379 (commit `ca83c6974`, merged 2026-06-12), whose body lists every commit on this branch. So all seven existing review threads are auto-resolved/outdated (the lines they referenced no longer exist in any diff). The remaining `CHANGES_REQUESTED` and `BEHIND` states are stale. Recommendation: close this PR and track follow-ups separately, since merging #6339 as-is is a no-op.

For the already-landed code on master, two follow-up issues worth a separate issue/PR:

1. Response-side repeated protobuf message fields lose nested schemas. In `OpenApiUtils.parseRepeatedProtobufField` (`OpenApiUtils.java:521-541`), a repeated `MESSAGE` field sets `elementType.setType("object")` + `setDescription(fullMsgName)` but never recurses or sets `refs`, so the element has no inner field schema. The request-side equivalent `parseRepeatedProtobufFieldSchema` (`:917-936`) does call `resolveProtobufMessageFieldSchema` and recurses. `OpenApiUtilsTest.java:145-149` only asserts `getType()=="object"` on the `addresses` element and never checks nested `getRefs()`, so the gap is untested. Suggested fix: mirror the request side by calling `resolveProtobufMessageField(elementType, ...)` for the `MESSAGE` branch and add a test asserting `street`/`city` on the repeated-message response element.

2. Inherited DTO fields are omitted. `parseClass` (`:692`) and `parseClassSchema` (`:803`), plus the parameterized variants (`:731`, `:841`), iterate only `getDeclaredFields()` with no superclass walk, so RPC request/response DTOs that extend a base class drop base-class fields from the generated doc. Suggested fix: walk `getSuperclass()` up to `Object`/framework bases and accumulate fields; add a test with a DTO extending a base DTO.

Minor: array element naming is `ITEMS` on the response side (`:668`, `:714`) but `items` on the request/schema side (`:827`, `:931`) — pick one (lowercase `items` matches OpenAPI) for consistency.

The SOFA listener refactor (`SofaServiceEventListener`) and the `handleClass`→`buildApiPath(method, superPath, null)` flow look correct — the parent only enters `handleClass` when `superPath.contains("*")`, so the `@Nullable` third argument is never dereferenced on the non-wildcard branch.

eye-gu:
@Aias00 PTAL.

Aias00 (approved):
Reviewed #6339 — Support request parameter parsing for RPC types (Dubbo/SOFA/TARS/gRPC) in API doc generation.

## Verdict: APPROVE

Genuine, well-tested improvement. (Note: I previously commented that this looked already-merged via #6379 and suggested closing — that tracing was wrong; `collectInstanceFields` does **not** exist on `master` and master still contains the `"ITEMS"` literal, so this PR's changes are real and not present upstream yet.)

### What it does
- **Inherited fields now documented.** New `collectInstanceFields(Class)` walks the class hierarchy (subclass-first, stopping at `Object` and protobuf base classes) collecting non-static instance fields. Previously only the most-derived class's own `getDeclaredFields()` were used, so superclass fields were silently omitted from generated RPC API docs. New test `testParseReturnTypeInheritedFields` covers a `SubDto extends BaseDto` (`name` + `id`) case.
- **Naming consistency.** Array element name `"ITEMS"` → `"items"` across `parseType`/`parseParameterizedType`/`parseGenericArrayType` and the protobuf paths, and `parseRepeatedProtobufField` now resolves the nested message element via `resolveProtobufMessageField` instead of leaving `elementType` nameless. Tests updated accordingly (e.g. `getRefs().get(6)` now asserts `items` with nested refs).
- **Cleanup.** Removes the unused `typeVariableMap` parameter threaded through every protobuf parse method, and adds a `tags` repeated field to the test proto to exercise the repeated-message path.

### Verification
- CI is green: `build (17/19/20/21, ubuntu + windows)`, all integrated-test jobs, `Analyze (java)`, `CodeQL`, `check-license-header`, `changes` all pass.
- The new/updated tests in `OpenApiUtilsTest` exercise inherited fields, protobuf field types, repeated fields, and the `items` naming — sufficient coverage for the changed behavior.

### Non-blocking follow-ups (optional, not required for merge)
- `parseParameterizedType` still uses `methodParam.getType()` (raw `Class<?>`) for RPC request parameters, so `List<String>` / `Map<K,V>` params render as plain `object` rather than accurate array/map schemas. Consider switching to `getParameterizedType()` + handling raw `Collection`/`Map` (a Copilot suggestion). The current behavior is correct for concrete field types; this is a schema-accuracy enhancement.
- For `Map` schemas, `refs` entries currently get no `name`; setting `"key"`/`"value"` (or an `additionalProperties` shape) would render more cleanly.

Approving — the feature works and is tested; the above are nice-to-haves for a later pass.

PR #6348, opened 2026-05-22
feat(client):add many path register for spring mvc client.

### LINKED ISSUE #6327: [Question] 是否支持多前缀注册到同一个controller
### Question

现在需要使用新的前缀，旧的前缀在切换过程需要保留；我要怎么做才可以实现在同一个controller里面注册两个路由前缀规则呢？
现在这样子进行配置，只有第一个前缀可以正常注册到devide的rule：
```
@RequestMapping(value = {
        "/mypath1/config",  
        "/other-path/config",
})
@Tag(name= "地址配置")
@ShenyuRequestMapping(path = {
        "/mypath1/config",  
        "/other-path/config",
})
public class MyConfigController {
...
}
```

wy471x:
Feat([#6327](https://github.com/apache/shenyu/issues/6327)): Support many path register into same controller.

Make sure that:

All changes:
1. SpringMvcClientEventListener

Unit Tests:
1.testBuildApiDocSextetDefaultProducesConsumes
2.testBuildApiDocSextetExplicitProducesConsumesAndMethod
3.testBuildApiDocSextetMultipleMethodsProducesConsumes

close #6327


wy471x:
Hi, @Aias00, when you have time, could you please help review this PR? Thank you.

Aias00:
@wy471x hi, pls fix the review

Aias00 (changes requested):
Thanks for the PR. The main route registration path looks like the right direction, but I think we should add one regression test before merging.

The new logic expands class-level paths in `SpringMvcClientEventListener#handle()` and also changes API doc generation to iterate `buildApiSuperPaths()`. However, the current unit tests only assert `buildApiSuperPaths()` and `buildApiDocSextet()`, while the integration test verifies gateway routing only.

Please add a publisher-level/unit test that captures the emitted `MetaDataRegisterDTO` and `ApiDocRegisterDTO` for a controller with multiple class-level paths, and assert the exact count and paths, for example both:

- `/mvc/multipath/v1/greet`
- `/mvc/multipath/v2/greet`

This is important because the registration contract is now "one Java method can produce multiple registered paths", while `metaDataMap` is still keyed by `Method` and API doc generation depends on that registration flow. Even if `ApiExt` is path-independent today, this behavior should be locked by tests.

Also, please consider avoiding the suffix-based fallback heuristic in `buildApiPath()` (`superPath.endsWith(formatPath(p))`) or at least cover it with tests. It is hard to reason about method-vs-class annotation provenance from string shape alone.

wy471x:
> Thanks for the PR. The main route registration path looks like the right direction, but I think we should add one regression test before merging.
> 
> The new logic expands class-level paths in `SpringMvcClientEventListener#handle()` and also changes API doc generation to iterate `buildApiSuperPaths()`. However, the current unit tests only assert `buildApiSuperPaths()` and `buildApiDocSextet()`, while the integration test verifies gateway routing only.
> 
> Please add a publisher-level/unit test that captures the emitted `MetaDataRegisterDTO` and `ApiDocRegisterDTO` for a controller with multiple class-level paths, and assert the exact count and paths, for example both:
> 
> * `/mvc/multipath/v1/greet`
> * `/mvc/multipath/v2/greet`
> 
> This is important because the registration contract is now "one Java method can produce multiple registered paths", while `metaDataMap` is still keyed by `Method` and API doc generation depends on that registration flow. Even if `ApiExt` is path-independent today, this behavior should be locked by tests.
> 
> Also, please consider avoiding the suffix-based fallback heuristic in `buildApiPath()` (`superPath.endsWith(formatPath(p))`) or at least cover it with tests. It is hard to reason about method-vs-class annotation provenance from string shape alone.

I have made the corresponding fixes and added relevant test cases for coverage. Please review them, thank you.

Aias00 (approved):
LGTM,thx

Aias00:
thx for pr, have my wechat? aias00

PR #6361, opened 2026-06-08
Stabilize k8s ingress startup healthchecks

Aias00:
## Summary
- wait for k8s ingress deployments to finish rollout before HTTP health probes
- add bounded curl timeouts and shared healthcheck logic for k8s ingress cases
- keep cluster diagnostics running after healthcheck failures and include pod descriptions

## Root cause
The linked grpc ingress CI run timed out in `Wait for k8s Cluster Start up` before tests started. The original healthcheck immediately curled `localhost:30095/actuator/health` after applying the kind resources, without waiting for deployment rollout and without per-request curl timeouts. When startup was slow or the endpoint accepted a connection but did not respond, a curl process could hang until the GitHub Actions 15-minute step timeout killed the job.

## Verification
- `bash -n shenyu-integrated-test/k8s-ingress-healthcheck.sh shenyu-integrated-test/shenyu-integrated-test-k8s-ingress-{grpc,http,apache-dubbo,websocket}/script/healthcheck.sh`
- `git diff --check`
- local fake `kubectl` + HTTP 200 success-path run
- local fake `kubectl` + unreachable-service failure-path run

## Not tested
- full GitHub Actions kind matrix run locally


Aias00:
Updated this PR to cover the additional `shenyu-integrated-test-http` CI failure from run 27113257022 as well.\n\nThat failure was `SentinelPluginTest.testFallbackUri` timing out after 180s. The test configured a plain fallback URI, which exercises an external 302 redirect fallback after a POST request. To avoid the flaky redirect boundary, the test now uses internal `fallback:` dispatch and the example accepted endpoint also supports POST for that fallback target.\n\nAdditional local verification:\n- `./mvnw -f shenyu-examples/shenyu-examples-http/pom.xml -DskipTests -Dmaven.javadoc.skip=true compile`\n- `./mvnw -f shenyu-integrated-test/pom.xml -pl shenyu-integrated-test-http -am -DskipTests -Pit -Dmaven.javadoc.skip=true test-compile`

Aias00:
Follow-up for the latest k8s ingress grpc failure: the new helper used `kubectl rollout status deployment --all`, but the workflow installs kubectl v1.24.14 and that command fails immediately with `unknown flag: --all` in job 80017461502.

I pushed `a479c8226` to keep the readiness wait compatible with that pinned kubectl version: it now enumerates deployments with `kubectl get deployment -n shenyu-ingress -o name` and runs `kubectl wait --for=condition=Available` for each deployment before the bounded HTTP health probes.

Local verification:
- `bash -n` for the shared helper and the four ingress healthcheck scripts
- `git diff --check` for the helper
- fake kubectl/curl success path
- fake kubectl/curl failed service path
- fake kubectl no-deployment guard path

Aias00:
Second follow-up after monitoring the new run: the kubectl compatibility issue is gone. The next failure was in `shenyu-integrated-test-k8s-ingress-apache-dubbo`, where the helper reached the bounded curl loop and `30095` became healthy, but `31187` stayed unreachable.

Root cause from the diagnostic dump: the Dubbo example pod was being restarted by kubelet because its liveness probe hit `localhost:8081/actuator/health`, while the app and service use port `8011`. I pushed `e2d530051` to align that probe with `8011`. The SOFA k8s example had the same stale `8081` probe while serving on `28011`, so that probe is aligned too.

Local verification:
- `git diff --check` for the changed manifests
- Ruby YAML parsing for the changed manifests

The local environment has no kube-apiserver, so `kubectl apply --dry-run=client` cannot complete here; the PR CI run is the end-to-end validation.

Aias00:
Third follow-up: after the liveness probe fix, apache-dubbo no longer restarted. The pod was Ready with restart count 0, but `localhost:31187` still returned 000 for every healthcheck attempt.

Root cause: the kind cluster config only mapped `30095` to the GitHub runner host. The ingress services.list files probe NodePorts via `localhost`, so each probed NodePort also needs an `extraPortMappings` entry. I pushed `072122525` to map the active ingress healthcheck ports: HTTP `31189`, gRPC `31184`, WebSocket `31188`, and apache-dubbo `31187`. I also fixed the latent SOFA services.list port from `31184` to its actual `31112` and mapped it.

Local verification:
- all ingress services.list localhost ports have matching kind hostPort mappings
- Ruby YAML parsing for changed kind configs/manifests
- `git diff --check`

Aias00:
Fourth follow-up: apache-dubbo now passes healthcheck and tests. The remaining failure moved to gRPC.

Root cause for gRPC: the example app serves HTTP health at `55290/grpc/actuator/health` because `server.servlet.context-path=/grpc`. The k8s liveness probe missed the `/grpc` context path, and services.list probed `31184`, which maps service port `8080`; the HTTP server port `55290` is exposed as NodePort `31183`.

I pushed `6ed35d3ca` to align all three pieces:
- liveness: `http://localhost:55290/grpc/actuator/health`
- services.list: `http://localhost:31183/grpc/actuator/health`
- kind hostPort mapping: `31183`

Local verification remains:
- all ingress services.list localhost ports have matching kind hostPort mappings
- Ruby YAML parsing for changed manifests
- `git diff --check`

Aias00:
Follow-up for the latest websocket ingress failure (job 80021225750):

- The preflight healthcheck reached the websocket example NodePort but got repeated 404s from `http://localhost:31188/new/feature/get/mapping/path`.
- That path is not implemented by `shenyu-example-spring-annotation-websocket`; the real test coverage is the websocket connection to `ws://localhost:30095/myWs`.
- Updated the preflight URL to `http://localhost:31188/actuator/health`, which is the same stable endpoint already used by the example's Kubernetes liveness probe.
- Increased the websocket example liveness initial delay from 30s to 60s because diagnostics showed an early startup liveness failure/restart before the service became ready.

Verification before push:
- `git diff --check`
- Ruby YAML parse for the websocket Kubernetes manifest
- kind hostPort mapping consistency check for all ingress service lists
- `./mvnw -f shenyu-integrated-test/pom.xml -pl shenyu-integrated-test-k8s-ingress-websocket -am -DskipTests -Pit -Dmaven.javadoc.skip=true test-compile`

Aias00:
Second websocket follow-up after run 27115745087 / job 80022215094:

- The stale `/new/feature/get/mapping/path` issue is confirmed fixed, but probing `http://localhost:31188/actuator/health` still failed with repeated 404s.
- The diagnostic logs show the ingress controller registered the backend as a websocket upstream: `protocol: ws://`, `upstreamUrl: 10.244.0.6:8001`.
- That means the websocket preflight should not use an HTTP actuator request through the websocket backend path. The real test path is `ws://localhost:30095/myWs`, and the stable preflight signal is that the ingress controller itself is up.

New commit `edb7ace53`:
- changes websocket preflight to `http://localhost:30095/actuator/health`
- removes the now-unused kind hostPort mapping for `31188`
- adds a `startupProbe` to the websocket example so liveness does not restart the app while it is still starting

Verification before push:
- `git diff --check`
- Ruby YAML parse for websocket Kubernetes manifest and kind config
- kind hostPort mapping consistency check for all ingress service lists
- `./mvnw -f shenyu-integrated-test/pom.xml -pl shenyu-integrated-test-k8s-ingress-websocket -am -DskipTests -Pit -Dmaven.javadoc.skip=true test-compile`

Aias00:
Follow-up for https://github.com/apache/shenyu/actions/runs/27116098121:

The websocket case now passed the gateway healthcheck step with the protocol-correct `localhost:30095/actuator/health` probe, but then failed in Run test because the backend example deployment in the `default` namespace was still starting. The diagnostics showed the backend pod at about 37s old with startupProbe warnings, and the ingress log had an empty websocket rule handle before reporting upstream configuration errors.

I pushed `87964af39` to make the shared k8s ingress healthcheck wait for deployments in both `shenyu-ingress` and `default` before polling service endpoints. This keeps the websocket probe protocol-correct while restoring an explicit readiness gate for the backend examples.

Aias00:
Follow-up for https://github.com/apache/shenyu/actions/runs/27116433626:

The new failure was in `shenyu-integrated-test-k8s-ingress-apache-dubbo` during the healthcheck step. The shared helper correctly waited for `shenyu-ingress` deployments, but then failed on `default` with `no deployments found in namespace default`. That namespace is expected to be empty for apache-dubbo because its example, ingress controller, and ZooKeeper workloads all live in `shenyu-ingress`.

I pushed `153603824` so `shenyu-ingress` remains required while `default` is optional. This keeps the websocket race fix, where `default` does contain a backend deployment, without failing ingress cases that do not use that namespace.

Aias00:
Follow-up for https://github.com/apache/shenyu/actions/runs/27116688369:

`http`, `grpc`, and `apache-dubbo` passed on `153603824`, so the shared helper and optional namespace handling are working for those cases. The remaining failure was `shenyu-integrated-test-k8s-ingress-websocket`: the healthcheck waited ten minutes for the default backend deployment, but the pod stayed `0/1` and restarted after `Startup probe failed` events.

Root cause: `shenyu-example-spring-annotation-websocket` has Kubernetes probes for `http://localhost:8001/actuator/health`, but the module did not include `spring-boot-starter-actuator`, unlike the native/reactive websocket examples. I pushed `4611d4f69` to add the actuator starter so the existing startup/liveness probes target a real health endpoint.

Aias00:
Follow-up for https://github.com/apache/shenyu/actions/runs/27117367880:

The websocket backend is now healthy: the healthcheck waited for the default deployment, it became available, and `localhost:30095/actuator/health` returned 200. The remaining failure moved to `Run test`: `WebsocketPluginTest` got `WebsocketNotConnectedException`.

The ingress controller log showed the cause: `EndpointsReconciler` repeatedly threw `NullPointerException` at `needUpdateSelectorId.addAll(selectorIdList)` because it always looked up `divide` selectors for endpoint updates. For websocket ingress there is no divide selector, so the websocket selector kept its empty upstream handle and the plugin logged `websocket upstream configuration error`.

I pushed `e2bfcb988` so endpoint reconciliation skips missing selector caches and refreshes both `divide` and `websocket` selector handles with plugin-appropriate upstream JSON. Added `EndpointsReconcilerTest` to cover the websocket endpoint update path.

Aias00:
Follow-up for the PR rollup failure after the ingress fix: `it / shenyu-integrated-test-http` failed in `SentinelPluginTest.testFallbackUri` with `expected: <202> but was: <404>`.

Root cause: the test used `fallback:/http/test/request/accepted`, but `fallback:` dispatch is handled locally by the gateway `DispatcherHandler`; it is not a fresh divide upstream request. In CI this resolved as a missing backend static resource (`No static resource http/test/request/accepted`) instead of exercising the sentinel fallback endpoint.

I pushed `7dc8849cc` to point the fallback URI at the existing gateway-local `fallback:/fallback/sentinel` endpoint and assert the sentinel fallback enum code/message, so the test now verifies the intended behavior without depending on an upstream route re-entry.

Local verification:
- `./mvnw -B -f shenyu-integrated-test/pom.xml -pl shenyu-integrated-test-http -am -DskipTests -Dmaven.javadoc.skip=true test-compile`
- `git diff --check`

Aias00:
Follow-up for run https://github.com/apache/shenyu/actions/runs/27118273553: the sentinel fallback test reached `fallback:/fallback/sentinel`, but the response was 405 because `FallbackHandler` preserves the original request method. The original request is POST, while `DefaultFallbackController` only exposed GET mappings.

I pushed `e042d2fbd` to make the default fallback endpoints method-agnostic `@RequestMapping` targets and added a POST regression test for `/fallback/sentinel`.

Local verification:
- `./mvnw -B -pl shenyu-web -DskipITs -DskipTests=false -Dtest=DefaultFallbackControllerTest -Dmaven.javadoc.skip=true test`
- `./mvnw -B -f shenyu-integrated-test/pom.xml -pl shenyu-integrated-test-http -am -DskipTests -Dmaven.javadoc.skip=true test-compile`
- `git diff --check`

Aias00:
Follow-up on the latest `it` failures: both failed before running integration tests while docker-maven-plugin was pulling Docker Hub base images:

- `alpine:latest`: `context deadline exceeded`
- `centos:7`: `context deadline exceeded`

This is independent of the code/test fixes, but it still makes the PR easy to fail because the workflow clears runner Docker caches before every matrix job. I pushed `671b80a01` to pre-pull the admin/bootstrap base images with retries before Maven starts, so docker-maven-plugin can use local cached layers.

Local verification:
- YAML parse for `.github/workflows/integrated-test.yml`
- `git diff --check`

Aias00:
Follow-up for run https://github.com/apache/shenyu/actions/runs/27119187870: the Docker pull issue was gone, and sentinel fallback returned the intended payload:

`{"code":429,"message":"SentinelPlugin fallback success, please check your service status!"}`

The remaining failure was in the test assertion: it deserialized the response into `ResultBean`, whose field is `msg`, while the fallback response uses `message`. I pushed `d229e0996` to assert the fallback response as a map against the actual `code` and `message` fields.

Local verification:
- `./mvnw -B -f shenyu-integrated-test/pom.xml -pl shenyu-integrated-test-http -am -DskipTests -Dmaven.javadoc.skip=true test-compile`
- `git diff --check`

Aias00:
Follow-up after the latest failures on head 16900d9ca:

* Root cause: the remaining red e2e/ci failures were still pre-test infrastructure failures from actions/setup-java@v4 downloading Temurin JDKs from GitHub releases and receiving HTTP 504. k8s-examples-http also had a transient k3s binary download failure.
* Fix: added actions/setup-java-with-retry and routed workflow-level setup-java calls through it; wrapped k8s-examples-http k3s installation in explicit retries.
* Verification on the latest PR run: it passed all 11 jobs, it-k8s passed all 4 jobs, k8s-examples-http passed, e2e build/images/storage/case jobs all passed, and ci matrix jobs for 17/19/20/21 Linux plus 17 Windows all passed. The only remaining pending checks at the time of this comment are the queued ci/e2e aggregate gate jobs; there are no completed failing checks.

yuluo-yx (approved):

PR #6375, opened 2026-06-09
fix(request): make add operations append existing values

### LINKED ISSUE #6360: [BUG] Request plugin add* operations overwrite existing values like set*

### Current Behavior

In the Request plugin, `addParameters`, `addHeaders`, and `addCookies` behave the same as their corresponding `set*` operations when the target key already exists.

For example, when the original request contains:

```text
?tag=old
```

and the Request plugin is configured with:

```json
{
  "parameter": {
    "addParameters": {
      "tag": "new"
    }
  }
}
```

the downstream service receives:

```text
tag=new
```

instead of preserving the original value.

The current implementation routes both `add*` and `set*` operations through the same logic based on `set()`, resulting in identical behavior.

### Expected Behavior

`add*` operations should preserve existing values and append the configured value, while `set*` operations should continue to overwrite existing values.

Example:

```text
Original request:
?tag=old

addParameters(tag=new)
```

Expected:

```text
tag=old&tag=new
```

or an equivalent multi-value representation.

The same semantic distinction should apply to headers and cookies.

### Steps To Reproduce

1. Enable the Request plugin.

2. Configure a Request plugin rule:

```json
{
  "parameter": {
    "addParameters": {
      "tag": "new"
    }
  }
}
```

3. Send a request containing an existing query parameter:

```bash
curl "http://localhost:9195/http/request-bug/echo?tag=old"
```

4. Observe that the downstream service receives:

```text
tag=new
```

instead of preserving the original value and appending the new value.

### Environment

```markdown
ShenYu version: 2.4.0
Apache ShenYu master branch
Module: shenyu-plugin-request
Local standalone deployment
```

### Debug logs

RequestHandle documents addHeaders as "need to be appended new header value",
while setHeaders is documented as covering/overwriting the value.

ModifyResponsePlugin also implements:

addHeaders -> HttpHeaders.add(...)
setHeaders -> HttpHeaders.set(...)

which suggests that add and set are intended to have different semantics.

### Anything else?

I would be happy to work on a PR if this behavior is confirmed as a bug😄.

SouthwestAsiaFloat:
## Description

This PR fixes the inconsistency between add* and set* operations in the Request Plugin.

Previously, addParameters, addHeaders, and addCookies internally used set() operations, causing existing values to be overwritten and making them behave the same as their corresponding set* operations.

This behavior is inconsistent with the documented semantics in RequestHandle and the implementation of ModifyResponsePlugin, where:

- add* operations append values.
- set* operations overwrite values.

## Changes

- Use queryParams.add(...) for addParameters.
- Use headers.add(...) for addHeaders.
- Use cookies.add(...) for addCookies.
- Keep existing setParameters, setHeaders, and setCookies behavior unchanged.
- Add/update unit tests covering requests that already contain the target key.

## Tests

- [x] `./mvn -pl shenyu-plugin/shenyu-plugin-request test`

## Checklist

- [x] I have read the contribution guidelines.
- [x] I submit test cases that back my changes.
- [x] My local tests passed.

Fixes #6360

### DIFF
--- a/shenyu-plugin/shenyu-plugin-request/src/main/java/org/apache/shenyu/plugin/request/RequestPlugin.java
+++ b/shenyu-plugin/shenyu-plugin-request/src/main/java/org/apache/shenyu/plugin/request/RequestPlugin.java
@@ -123,10 +123,10 @@ public class RequestPlugin extends AbstractShenyuPlugin {
             return;
         }
         if (MapUtils.isNotEmpty(shenyuReqHeader.getAddHeaders())) {
-            shenyuReqHeader.getAddHeaders().entrySet().forEach(s -> this.fillHeader(s, headers));
+            shenyuReqHeader.getAddHeaders().entrySet().forEach(s -> this.addHeader(s, headers));
         }
         if (MapUtils.isNotEmpty(shenyuReqHeader.getSetHeaders())) {
-            shenyuReqHeader.getSetHeaders().entrySet().forEach(s -> this.fillHeader(s, headers));
+            shenyuReqHeader.getSetHeaders().entrySet().forEach(s -> this.setHeader(s, headers));
         }
         if (MapUtils.isNotEmpty(shenyuReqHeader.getReplaceHeaderKeys())) {
             shenyuReqHeader.getReplaceHeaderKeys().entrySet().forEach(s -> this.replaceHeaderKey(s, headers));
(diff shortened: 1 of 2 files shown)


Aias00 (approved):
LGTM

Aias00:
pls add my wechat: aias00

Issue #6398, opened 2026-06-26
[Task] Decouple JWT signing key from user password hash


Aias00:
### Description

DashboardUserServiceImpl currently uses the SHA-512 hash of the user's password as the HMAC-SHA256 signing key for JWT tokens (JwtUtils.generateToken(userName, getPassword(), ...)). This design has two weaknesses:

If the database is compromised, an attacker can use the stored password hash directly as a JWT signing key without needing to crack the original password.
Changing a user's password silently invalidates all existing sessions for that user with no explicit revocation mechanism.

Affected Files:

shenyu-admin/src/main/java/org/apache/shenyu/admin/utils/JwtUtils.java
shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/DashboardUserServiceImpl.java
shenyu-admin/src/main/resources/application.yml

Proposed Change:

Introduce a shenyu.jwt.secret-key configuration property in application.yml that holds a randomly generated secret independent of user credentials.
Update JwtUtils.generateToken() and JwtUtils.verifyToken() to use this configured secret instead of the password hash.
Generate a default random value at startup if the property is not configured, with a warning log recommending explicit configuration.

Acceptance Criteria:

JWT tokens are signed with a configured or randomly generated secret key, not the user's password hash.
Changing a user's password does not affect existing JWT validity.
shenyu.jwt.secret-key is documented in the configuration reference.

hengyuss:
Hi, @Aias00  I've submitted a PR to address it. Could you please review when you have time? Thanks!

PR #6407, opened 2026-06-27
Fix/decouple jwt signing key from user password hash

hengyuss:
#6398 

decouple jwt signing key from user password hash, 
add secret-key in application.yml to configure the key of jwt secret key, the code will generate a random string as secretKey if user doesn't configure it.
 i don't update JwtUtils.generateToken() and JwtUtils.verifyToken() . the two function receive the argument as secretKey and i think the behavior is correct. So that i change two place that invoke these two funcitons.
Make sure that:


hengyuss:
the head ref contain hidden characters.

Issue #6401, opened 2026-06-26
[Task] Add response body size limit in SwaggerImportServiceImpl.fetchSwaggerDoc()


Aias00:
### Description

SwaggerImportServiceImpl.fetchSwaggerDoc() uses response.body().string() to read the full HTTP response into memory without any size constraint. On the MCP import path, the response is subsequently parsed three times (Gson + OpenAPIV3Parser + JsonParser), multiplying memory usage. An authenticated Admin user can supply a URL pointing to a large file, causing JVM heap exhaustion and crashing the Admin process.
Affected File:

shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/SwaggerImportServiceImpl.java
Proposed Change:

Replace the unbounded response.body().string() call with a size-limited counted read:
javaprivate static final long MAX_SWAGGER_BODY_BYTES = 10L * 1024 * 1024; // 10 MB

private String fetchSwaggerDoc(final String swaggerUrl) throws IOException {
    // Check Content-Length header first, then count bytes during streaming read
    // Throw IllegalArgumentException if limit exceeded
}
Acceptance Criteria:

Requests to /swagger/import or /swagger/import/mcp with a URL pointing to a file larger than the configured limit return a 400 Bad Request response.
The size limit is configurable via application.yml.
Legitimate Swagger/OpenAPI documents within the size limit are imported without issue.

SouthwestAsiaFloat:
Hello, I'd like to handle this task😄

PR #6411, opened 2026-06-28
fix(admin): limit swagger import response body size

SouthwestAsiaFloat:
#6401

  ### Description

  This PR limits the Swagger/OpenAPI import response body size in shenyu-admin.

  It adds a configurable `shenyu.swagger.max-body-size` setting with a default value of 10 MB, rejects oversized Swagger responses both by `Content-Length` and during streaming, and returns HTTP 400 for
  invalid or oversized import requests.

  ### Tests

  Added tests in:
  - `SwaggerImportControllerTest`
  - `SwaggerImportServiceImplTest`

  The tests cover oversized Swagger response rejection, HTTP 400 responses for invalid import requests, successful small-body imports, unknown `Content-Length` streaming limits, exact-limit bodies, empty
  response bodies, and response charset handling.

Make sure that:

### DIFF
--- a/shenyu-admin/src/main/java/org/apache/shenyu/admin/controller/SwaggerImportController.java
+++ b/shenyu-admin/src/main/java/org/apache/shenyu/admin/controller/SwaggerImportController.java
@@ -24,7 +24,6 @@ import org.apache.shenyu.admin.service.SwaggerImportService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.http.HttpStatus;
-import org.springframework.http.ResponseEntity;
 import org.springframework.validation.annotation.Validated;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestBody;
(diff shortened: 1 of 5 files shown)


shenyu-admin/src/main/java/org/apache/shenyu/admin/controller/SwaggerImportController.java
```diff
@@ -53,17 +55,22 @@ public SwaggerImportController(final SwaggerImportService swaggerImportService)
      * @return the result of swagger import
      */
     @PostMapping("/import")
-    public ShenyuAdminResult importSwagger(@Valid @RequestBody final SwaggerImportRequest request) {
+    public ResponseEntity<ShenyuAdminResult> importSwagger(@Valid @RequestBody final SwaggerImportRequest request) {
```
Aias00: why change response type?

shenyu-admin/src/main/java/org/apache/shenyu/admin/controller/SwaggerImportController.java
```diff
@@ -74,14 +81,19 @@ public ShenyuAdminResult importSwagger(@Valid @RequestBody final SwaggerImportRe
      * @return the result of config import
      */
     @PostMapping("import/mcp")
-    public ShenyuAdminResult importMcpConfig(@Valid @RequestBody final SwaggerImportRequest request) {
+    public ResponseEntity<ShenyuAdminResult> importMcpConfig(@Valid @RequestBody final SwaggerImportRequest request) {
```
Aias00: why change response type?

Aias00 (changes requested):

shenyu-admin/src/main/java/org/apache/shenyu/admin/controller/SwaggerImportController.java
SouthwestAsiaFloat: I changed the response type because the PR description says that invalid or oversized import requests should return HTTP 400. The original return type (ShenyuAdminResult) only carries the business error code in the response body, and when returned directly, Spring MVC still responds with HTTP 200 by default. Using ResponseEntity<ShenyuAdminResult> allows the endpoint to return both an HTTP 400 status and the existing ShenyuAdminResult response body.

If the expected behavior is to keep the existing admin API convention (i.e. returning HTTP 200 with ShenyuAdminResult.code = 400 for business errors), I’m happy to change it back and update the implementation accordingly.

My apologies if I misunderstood the expected behavior.

Aias00: pls change to ShenyuAdminResult

SouthwestAsiaFloat: Done

Aias00 (approved):
LGTM

Issue #6422, opened 2026-07-15
updateAppSecret error! always  return  app key not existed  AppAuthController#updateSk


KANLON:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

The `GET /appAuth/updateSk` endpoint always fails validation with the error `updateSk.appKey: app key not existed`, regardless of whether the provided `appKey` actually exists in the database.

**Root cause:** The `@Existed` validation on the `appKey` parameter uses `AppKeyProvider`, whose `existed()` method calls `appAuthMapper.existed(key)`. However, `AppAuthMapper.existed()` queries the `id` (primary key) column instead of the `app_key` column:

```xml
<!-- Currently used (WRONG) - queries by id -->
<select id="existed" resultType="java.lang.Boolean">
    SELECT true FROM app_auth WHERE id = #{id} LIMIT 1
</select>

<!-- Should use this - queries by app_key -->
<select id="appKeyExisted" resultType="java.lang.Boolean">
    SELECT true FROM app_auth WHERE app_key = #{appKey} LIMIT 1
</select>
```

Since the passed `appKey` value is checked against the `id` column, the query never matches, so the validation always fails.

**Fix:** In `AppKeyProvider.existed()`, call `appAuthMapper.appKeyExisted(key)` instead of `appAuthMapper.existed(key)`.

### Environment

```markdown
ShenYu version(s):
```

yu199195:
Hi，can you fixed it ？

KANLON:
> Hi，can you fixed it ？

ok，i pull a pr later

KANLON:
> Hi，can you fixed it ？

https://github.com/apache/shenyu/pull/6425  the pr

KANLON:
sorry， i find  the bug only in version   2.5.1

PR #6425, opened 2026-07-15
Fix the method for verifying the existence of the appkey

KANLON:
https://github.com/apache/shenyu/issues/6422  

The `GET /appAuth/updateSk` endpoint always fails validation with the error `updateSk.appKey: app key not existed`, regardless of whether the provided `appKey` actually exists in the database.

**Root cause:** The `@Existed` validation on the `appKey` parameter uses `AppKeyProvider`, whose `existed()` method calls `appAuthMapper.existed(key)`. However, `AppAuthMapper.existed()` queries the `id` (primary key) column instead of the `app_key` column:

```xml
<!-- Currently used (WRONG) - queries by id -->
<select id="existed" resultType="java.lang.Boolean">
    SELECT true FROM app_auth WHERE id = #{id} LIMIT 1
</select>

<!-- Should use this - queries by app_key -->
<select id="appKeyExisted" resultType="java.lang.Boolean">
    SELECT true FROM app_auth WHERE app_key = #{appKey} LIMIT 1
</select>
```

Since the passed `appKey` value is checked against the `id` column, the query never matches, so the validation always fails.

**Fix:** In `AppKeyProvider.existed()`, call `appAuthMapper.appKeyExisted(key)` instead of `appAuthMapper.existed(key)`.

Make sure that:

### DIFF
--- a/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/provider/AppKeyProvider.java
+++ b/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/provider/AppKeyProvider.java
@@ -37,6 +37,6 @@ public class AppKeyProvider implements ExistProvider {
     
     @Override
     public Boolean existed(final Serializable key) {
-        return appAuthMapper.existed(key);
+        return appAuthMapper.appKeyExisted(key);
     }
 }


Aias00:
pls add my wechat: aias00

PR #6456, opened 2026-07-25
fix: support IPv6 address parsing in upstream URL handling

### LINKED ISSUE #6442: [BUG] IPv6 upstream addresses are parsed incorrectly by colon splitting
### Current Behavior

Several admin/common runtime paths parse upstream URLs by splitting on `:`. This breaks valid IPv6 upstream addresses.

Evidence:

- `shenyu-common/src/main/java/org/apache/shenyu/common/utils/UpstreamCheckUtils.java:67-73`
- `shenyu-admin/src/main/java/org/apache/shenyu/admin/service/manager/impl/LoadServiceDocEntryImpl.java:153-158`
- `shenyu-admin/src/main/java/org/apache/shenyu/admin/transfer/DiscoveryTransfer.java:352-361`

Examples:

```java
hostPort = StringUtils.split(http[1], Constants.COLONS);
...
String[] upstreamUrlArr = discoveryUpstreamData.getUrl().split(":");
...
.map(url -> url.split(":", 2))
```

For valid IPv6 literals such as `[2001:db8::1]:8080` or `http://[2001:db8::1]:8080`, colon splitting truncates the host and/or attempts to parse part of the IPv6 literal as a port. That can produce a failed health check, wrong discovery upstream data, or `NumberFormatException`.

### Expected Behavior

Valid IPv6 upstream addresses should be parsed with URI/host-port aware logic and should work in health checks, discovery conversion, and API document loading.

### Steps To Reproduce

Use an upstream URL such as:

```text
[2001:db8::1]:8080
http://[2001:db8::1]:8080
```

Then trigger one of these paths:

- upstream health check via `UpstreamCheckUtils.checkUrl(...)`
- discovery upstream conversion via `DiscoveryTransfer.mapToDiscoveryUpstreamData(...)`
- API document instance loading via `LoadServiceDocEntryImpl.getInstances2(...)`

### Suggested Fix

Use `java.net.URI` or a host-port parser that supports bracketed IPv6 literals. Avoid raw `split(":")` for URLs or host:port strings.

wy471x:
Replace raw split(":") with IPv6-aware host:port parsing using bracket
notation detection and java.net.URI to avoid breaking IPv6 addresses
like [2001:db8::1]:8080 in health checks, discovery, and API doc loading.

<!--
-->
Make sure that:

## Summary of All Changes
         
### Problem                                                                            
                    
  Four files parsed upstream URLs/host:port strings by splitting on :. This broke    
  IPv6 addresses like [2001:db8::1]:8080 — internal colons caused host truncation,   
  wrong port parsing, or NumberFormatException.                                      
                                                                                     
### Files Changed                                                        
   
  Core Logic — shenyu-common                                                                                                                                                         
                                                                                                                                                                                     
  IpUtils.java — New parseHostPort() method that correctly handles:                                                                                                                  
  - IPv4 with/without port (192.168.1.1:8080, 192.168.1.1)                                                                                                                           
  - IPv6 bracket notation with/without port ([::1]:9090, [2001:db8::1])
  - IPv6 zone IDs ([fe80::1%eth0]:8080) — stripped before validation to avoid platform-dependent UnknownHostException
  - Hostnames with/without port (example.com:8080, example.com)                                                      
  - Null/blank input → IllegalArgumentException                                                                                                                                      
  - Non-numeric port → IllegalArgumentException
  - Port out of range (0 or >65535) → IllegalArgumentException                                                                                                                       
  - Trailing junk after ] → IllegalArgumentException                                                                                                                                 
  - Empty port defaults to "80"                                                                                                                                                      
                                                                                                                                                                                     
  UpstreamCheckUtils.java — Adapted to use the new parseHostPort().                                                                                                                  
                                                                                                                                                                                     
  Wrapper & Callers — shenyu-admin                                                                                                                                                   
                                                                                                                                                                                     
  CommonUpstreamUtils.java — Two new utility methods:                                                                                                                                
  - normalizeUrl(url) — Normalizes upstream URLs by adding default port 80 when missing, and wrapping bare IPv6 addresses in brackets
  - parseHostPort(url) — Delegates to IpUtils.parseHostPort()                                                                                                                        
                                                             
  DiscoveryTransfer.java — Uses normalizeUrl() when converting discovery upstream DTOs, plus null/blank guards on URL and properties.                                                
                                                                                                                                                                                     
  DiscoveryUpstreamServiceImpl.java — Transaction atomicity (@Transactional), host:port fallback logic, normalization on both create and update paths.                               
                                                                                                                                                                                     
  LoadServiceDocEntryImpl.java — Adapted to use new parseHostPort().                                                                                                                 
                  
  UpstreamCheckService.java — Adapted to use new parseHostPort().                                                                                                                    
                  
  DiscoveryDataChangedEventSyncListener.java — Null guard on discoveryUpstreams list. 

### Notes
Scope: shenyu-admin and shenyu-common only. 
Follow-up: Gateway-side consumers (grpc, tars) to be handled in subsequent work.

close [#6442](https://github.com/apache/shenyu/issues/6442)


Aias00:
I found a compatibility issue in the IPv6 URL handling.

This PR changes the canonical serialized IPv6 upstream URL from the old bare form (`host:port`) to bracketed form (`[host]:port`) in `CommonUpstreamUtils.buildUrl(...)`. However, discovery persistence, dedupe, status updates, and deletes still use exact `upstream_url` string matching.

That means an existing persisted IPv6 upstream written by older code, for example `2001:db8::1:8080`, will not match the new `[2001:db8::1]:8080` value after upgrade. `nativeCreateOrUpdate(...)` can insert a duplicate instead of updating the old row, and status/delete paths can leave the old row stale because the SQL and in-memory comparisons still key on exact `upstream_url`.

There are a few parser edge cases worth covering at the same time:

- Bare IPv6 without brackets, e.g. `2001:db8::1`, is split on the last colon and becomes host `2001:db8:` + port `1`.
- Already bracketed host input passed to `buildUrl`, e.g. `[2001:db8::1]`, becomes `[[2001:db8::1]]:8080`.
- Bracketed malformed input like `[2001:db8::1` throws a low-level `StringIndexOutOfBoundsException`, while `[2001:db8::1]junk` is silently accepted as port 80.

Could we add an upgrade-safe canonicalization strategy here, either by migrating old `upstream_url` values to the new bracketed form or by making lookup/update/delete/dedupe tolerant of both legacy and canonical forms until data is normalized? Tests for legacy bare IPv6 rows plus the malformed parser cases would catch the risky paths.

wy471x:
> I found a compatibility issue in the IPv6 URL handling.
> 
> This PR changes the canonical serialized IPv6 upstream URL from the old bare form (`host:port`) to bracketed form (`[host]:port`) in `CommonUpstreamUtils.buildUrl(...)`. However, discovery persistence, dedupe, status updates, and deletes still use exact `upstream_url` string matching.
> 
> That means an existing persisted IPv6 upstream written by older code, for example `2001:db8::1:8080`, will not match the new `[2001:db8::1]:8080` value after upgrade. `nativeCreateOrUpdate(...)` can insert a duplicate instead of updating the old row, and status/delete paths can leave the old row stale because the SQL and in-memory comparisons still key on exact `upstream_url`.
> 
> There are a few parser edge cases worth covering at the same time:
> 
> * Bare IPv6 without brackets, e.g. `2001:db8::1`, is split on the last colon and becomes host `2001:db8:` + port `1`.
> * Already bracketed host input passed to `buildUrl`, e.g. `[2001:db8::1]`, becomes `[[2001:db8::1]]:8080`.
> * Bracketed malformed input like `[2001:db8::1` throws a low-level `StringIndexOutOfBoundsException`, while `[2001:db8::1]junk` is silently accepted as port 80.
> 
> Could we add an upgrade-safe canonicalization strategy here, either by migrating old `upstream_url` values to the new bracketed form or by making lookup/update/delete/dedupe tolerant of both legacy and canonical forms until data is normalized? Tests for legacy bare IPv6 rows plus the malformed parser cases would catch the risky paths.

@Aias00 Hi, I have fixed the issues you mentioned. Please take a look when you have time.

Aias00 (changes requested):
Reviewed #6456 — IPv6 address parsing in upstream URL handling.

Thanks for this — the IPv6-aware parsing in `IpUtils.parseHostPort` and the `URI`-based `checkUrl` rewrite look correct (verified `URI.getHost()` returns the bracketed IPv6 form and `InetSocketAddress` accepts it). But there's a blocking issue plus a few consistency gaps to address before merge.

**Blocker — CI is red (NPE regression in `nativeCreateOrUpdate`)**

`DiscoveryUpstreamServiceImpl.nativeCreateOrUpdate` now calls `CommonUpstreamUtils.normalizeUrl(...)` unconditionally, and `normalizeUrl` → `IpUtils.parseHostPort` does `upstreamUrl.startsWith("[")` with no null guard. The existing `DiscoveryUpstreamServiceTest.testNativeCreate` builds a DTO with no URL set, so the PR now throws:

```
java.lang.NullPointerException: Cannot invoke "String.startsWith(String)" because "upstreamUrl" is null
  at DiscoveryUpstreamServiceTest.testNativeCreate(DiscoveryUpstreamServiceTest.java:205)
  at DiscoveryUpstreamServiceTest.testNativeCreateOrUpdate(DiscoveryUpstreamServiceTest.java:127)
```

This is the actual cause of the `build (17, windows-latest)` FAILURE (`ci` workflow); the ubuntu matrix was cancelled by fail-fast. `mergeStateStatus` is BLOCKED. Beyond the test, it's a runtime regression: master inserted the row (with null url) without throwing, the PR now 500s. The new `DiscoveryUpstreamServiceImplTest` always sets a URL via `buildDTO`, so it masks the regression; the pre-existing `DiscoveryUpstreamServiceTest` was not updated.

Fix options: null/blank-guard `IpUtils.parseHostPort` (or `normalizeUrl`), and add the same exception guard to the service path that you already added to the listener's `handleAdded/Updated/Deleted`. Then fix `testNativeCreate` to set a URL (or assert the new typed exception).

**Should-fix**

1. Listener swallows exceptions inside `@Transactional`. `handleUpdated`/`handleDeleted` catch `Exception`, log, and continue, but `onChange` is `@Transactional(rollbackFor=Exception.class)` — so partial batches now commit instead of rolling back (master propagated → rollback). Either rethrow after logging to keep atomicity, or drop the `@Transactional` and document per-item best-effort; don't swallow-and-commit silently.

2. Service-layer write paths lack the host:port fallback. `deleteBySelectorIdAndUrl` and `changeStatusBySelectorIdAndUrl` only exact-match the normalized URL; the listener additionally falls back to `matchByHostAndPort` for old-format records. So old `example.com` (no port) or unbracketed `2001:db8::1:8080` rows can't be deleted/status-toggled via the register/REST path until migrated, while the event path can. Reuse `matchByHostAndPort` in both service methods (delete/update by id) so all write paths match.

3. `mapToCommonUpstream` is now throwable in unguarded streams. Replacing `url.split(":")[0]` with `parseHostPort(url)[0]` means malformed bracketed input (`[broken`, `[]`, `[localhost]`) throws `IllegalArgumentException`. Callers in `UpstreamCheckService` (lines 482-484, 509) map with no try/catch — one bad row breaks the whole DIVIDE stream. Skip-and-log per item, or return null on parse failure and filter.

**Nits**

- `normalizeUrl` does `Integer.parseInt(parts[1])` but `parseHostPort` doesn't validate the port token is numeric (`example.com:abc` → uncaught NumberFormatException in the service path). Validate the port in `parseHostPort` or guard callers.
- IPv6 coverage stops at admin/common. `ShenyuResolverHelper` (grpc) and `PrxInfoUtil` (tars) still `split(":")` and will NumberFormatException on `[ipv6]:port`. Pre-existing (unbracketed form was already broken), but please scope the PR description to admin/common, or open a follow-up for the gateway-side consumers.

Tests: the new `IpUtilsTest` / `CommonUpstreamUtilsTest` cases are good and cover IPv4/IPv6/hostname/bracket edge cases. Please also add round-trip `normalizeUrl` tests for hostname-without-port and IPv6-with-zone. And re-run CI with `fail-fast: false` so the ubuntu matrix actually executes.

Aias00 (commented):
Follow-up after the new commits (d05fec8 + f0a26bf). Thanks — the substantive issues are addressed. Per-finding:

- **#1 blocker (null-url NPE in `nativeCreateOrUpdate` + CI red): fixed.** `IpUtils.parseHostPort` and `CommonUpstreamUtils.normalizeUrl` now throw `IllegalArgumentException("Invalid upstream URL, null or blank")` on blank/null, and `testNativeCreate` now sets a URL. The originally-red `k8s-examples-http` is passing again. Verified in commit f0a26bf.
- **#2 listener swallowed exceptions inside `@Transactional`: fixed.** `handleUpdated`/`handleDeleted` now `throw e` after logging (commit d05fec8), so `onChange` rolls back on failure — atomicity restored.
- **#3 service write paths lacked host:port fallback: fixed.** `deleteBySelectorIdAndUrl` and `changeStatusBySelectorIdAndUrl` now check `effect == 0` and fall back to `matchByHostAndPort` + delete/update by id, mirroring the listener path (d05fec8).
- **#4 `mapToCommonUpstream` throwable in unguarded streams: fixed.** `mapToCommonUpstream` and `mapToDiscoveryUpstreamData` now wrap `parseHostPort` in try/catch, return `null` + log warn on `IllegalArgumentException`, and all three stream call sites (`updateSelectorHandler`, `fetchUpstreamData`, `onDiscoveryUpstreamUpdated`) `.filter(Objects::nonNull)`. `mapToDiscoveryUpstreamData`'s null-url path now returns `null` instead of throwing (d05fec8).
- **#5 non-numeric port `NumberFormatException`: partially fixed.** Blank/null is now a typed `IllegalArgumentException`. A non-numeric port (`example.com:abc`) still makes `normalizeUrl`'s `Integer.parseInt(parts[1])` throw `NumberFormatException`. In the transfer/stream paths that's now harmless (NFE is a subclass of IAE, caught and skipped), but the direct `normalizeUrl` call in `nativeCreateOrUpdate` still surfaces an untyped NFE → 500. Minor — consider validating the port token in `parseHostPort` (throw typed IAE) so the service path fails cleanly too. Not a blocker.
- **#6 IPv6 coverage stops at admin/common (`ShenyuResolverHelper`/`PrxInfoUtil` still `split(":")`): not addressed.** Out of scope for this PR as flagged; please open a follow-up for the grpc/tars gateway-side consumers, or note in the PR description that this PR scopes to admin/common.

One thing to look at before merge: the `build` matrix is currently red on all OS jobs (windows + ubuntu 17/19/20/21), but it's **not a code failure** — the log shows `##[error]The operation was canceled.` ~7 min in, with every surefire suite passing (`Tests run: X, Failures: 0, Errors: 0`) and no compile errors. It looks like a job timeout/cancellation, not a regression from these commits. Please re-run CI; if it reproduces, the `build` workflow's `timeout-minutes` may need bumping (the reactor is large and shared runners are slow). Once `build` is green I'm happy to approve — no code blockers remain.

Aias00 (commented):
Correction to my previous comment — I was wrong that the `build` red was a timeout. I dug into the per-job conclusions of run 30756841941 and the real failure is on the **Windows** job; the four ubuntu jobs were cancelled by `fail-fast` (the `build` matrix uses the default `fail-fast: true`), which is why their logs end with `The operation was canceled.` mid-build. So the red is real test failures in this PR's own code, not an infra timeout. Please ignore my earlier "re-run and approve" line — re-running will fail the same way until these are fixed.

The Windows job (`build (17, windows-latest)`) ends with `BUILD FAILURE` — `Tests run: 1474, Failures: 0, Errors: 3`. The three errors:

1. **`DiscoveryUpstreamServiceTest.testNativeCreateOrUpdate:128 -> testNativeUpdate:200`** — `java.lang.IllegalArgumentException: Invalid upstream URL, null or blank`.
   Your `f0a26bf` blank-guard in `parseHostPort`/`normalizeUrl` throws on a null/blank URL, and you fixed `testNativeCreate` to set `192.168.1.1:8080` — but `testNativeUpdate` (also called by `testNativeCreateOrUpdate` at line 128) still builds a DTO with no URL set, so the new guard throws. Fix: set a URL in `testNativeUpdate` too (the blank-guard is the right behavior; the test was relying on the old null-tolerant path).

2. **`CommonUpstreamUtilsTest.testNormalizeUrlIpv6WithZoneIdAndPortRoundTrip:506`** and
   **`testNormalizeUrlIpv6WithZoneIdWithoutPortRoundTrip:515`** — `java.lang.IllegalArgumentException: Invalid IPv6 address: fe80::1%lo0`, `Caused by: java.net.UnknownHostException: no such interface lo0`.
   These new tests use `fe80::1%lo0`, where `lo0` is the macOS loopback interface name. Linux and Windows CI runners have no `lo0` interface, so the zone-id resolution fails and the test errors. So these tests are platform-dependent and break on every non-macOS runner (they'd fail on Linux too — windows just happened to reach the module first).

This points at a **product gap, not just a test bug**: `parseHostPort` resolves the IPv6 zone id to a real network interface, so on a Linux/Windows gateway an upstream URL containing a link-local IPv6 with a zone id (e.g. `fe80::1%eth0`) will fail to parse at runtime. For an IPv6-support PR, link-local + zone ids being unsupported on the production platform is a real correctness gap — please either parse the zone id without resolving it to an interface (treat it as an opaque scope id), or explicitly document that link-local zone-id upstreams are unsupported and skip these tests on CI rather than hardcoding a macOS interface name.

So: the blocker stands (CI red, 3 real failures). Keeping changes-requested. Once `testNativeUpdate` sets a URL and the zone-id case is either implemented without interface resolution or the tests are made platform-agnostic, `build` should go green.

wy471x:
> Follow-up after the new commits ([d05fec8](https://github.com/apache/shenyu/commit/d05fec8182f31c1131831c520572aa9abee00216) + [f0a26bf](https://github.com/apache/shenyu/commit/f0a26bf6d6621f2a8885fb0df22c2711133efe71)). Thanks — the substantive issues are addressed. Per-finding:
> 
> * **#1 blocker (null-url NPE in `nativeCreateOrUpdate` + CI red): fixed.** `IpUtils.parseHostPort` and `CommonUpstreamUtils.normalizeUrl` now throw `IllegalArgumentException("Invalid upstream URL, null or blank")` on blank/null, and `testNativeCreate` now sets a URL. The originally-red `k8s-examples-http` is passing again. Verified in commit [f0a26bf](https://github.com/apache/shenyu/commit/f0a26bf6d6621f2a8885fb0df22c2711133efe71).
> * **#2 listener swallowed exceptions inside `@Transactional`: fixed.** `handleUpdated`/`handleDeleted` now `throw e` after logging (commit [d05fec8](https://github.com/apache/shenyu/commit/d05fec8182f31c1131831c520572aa9abee00216)), so `onChange` rolls back on failure — atomicity restored.
> * **#3 service write paths lacked host:port fallback: fixed.** `deleteBySelectorIdAndUrl` and `changeStatusBySelectorIdAndUrl` now check `effect == 0` and fall back to `matchByHostAndPort` + delete/update by id, mirroring the listener path ([d05fec8](https://github.com/apache/shenyu/commit/d05fec8182f31c1131831c520572aa9abee00216)).
> * **#4 `mapToCommonUpstream` throwable in unguarded streams: fixed.** `mapToCommonUpstream` and `mapToDiscoveryUpstreamData` now wrap `parseHostPort` in try/catch, return `null` + log warn on `IllegalArgumentException`, and all three stream call sites (`updateSelectorHandler`, `fetchUpstreamData`, `onDiscoveryUpstreamUpdated`) `.filter(Objects::nonNull)`. `mapToDiscoveryUpstreamData`'s null-url path now returns `null` instead of throwing ([d05fec8](https://github.com/apache/shenyu/commit/d05fec8182f31c1131831c520572aa9abee00216)).
> * **#5 non-numeric port `NumberFormatException`: partially fixed.** Blank/null is now a typed `IllegalArgumentException`. A non-numeric port (`example.com:abc`) still makes `normalizeUrl`'s `Integer.parseInt(parts[1])` throw `NumberFormatException`. In the transfer/stream paths that's now harmless (NFE is a subclass of IAE, caught and skipped), but the direct `normalizeUrl` call in `nativeCreateOrUpdate` still surfaces an untyped NFE → 500. Minor — consider validating the port token in `parseHostPort` (throw typed IAE) so the service path fails cleanly too. Not a blocker.
> * **#6 IPv6 coverage stops at admin/common (`ShenyuResolverHelper`/`PrxInfoUtil` still `split(":")`): not addressed.** Out of scope for this PR as flagged; please open a follow-up for the grpc/tars gateway-side consumers, or note in the PR description that this PR scopes to admin/common.
> 
> One thing to look at before merge: the `build` matrix is currently red on all OS jobs (windows + ubuntu 17/19/20/21), but it's **not a code failure** — the log shows `##[error]The operation was canceled.` ~7 min in, with every surefire suite passing (`Tests run: X, Failures: 0, Errors: 0`) and no compile errors. It looks like a job timeout/cancellation, not a regression from these commits. Please re-run CI; if it reproduces, the `build` workflow's `timeout-minutes` may need bumping (the reactor is large and shared runners are slow). Once `build` is green I'm happy to approve — no code blockers remain.

Hi,I have fixed the issues you mentioned. Please take a look when you have time.

Aias00:
The core bracket-IPv6 path is correct, and the layered defense (skip malformed URLs in `DiscoveryTransfer.mapToCommonUpstream` + `Objects::nonNull` filters in `UpstreamCheckService`) is good. A few things worth addressing:

**Non-bracketed IPv6 is silently misparsed.** `IpUtils.parseHostPort` falls back to `lastIndexOf(':')` for non-bracketed input. That works for `host:port` and IPv4, but for a non-bracketed IPv6 like `::1` (no port) it produces `host=":"`, `port="1"`, and `2001:db8::1:8080` produces `host="2001:db8::1"`, `port="8080"` (only coincidentally right when the last segment is the port). There's no "more than one colon and not bracketed → reject" guard, so a user who forgets brackets gets a silently broken upstream (garbage host) instead of a clear error. Since `normalizeUrl` then feeds that garbage host into `buildUrl`, you get `[:]`-style nonsense stored. Suggest: if a non-bracketed URL has >1 colon, throw "IPv6 addresses must be bracketed, e.g. [::1]:8080" (or parse it as IPv6+port). file: `IpUtils.parseHostPort`.

**`matchByHostAndPort` is O(N×K) DB load.** It's called per-upstream inside the `handleAdded`/`handleUpdated`/`handleDeleted` loops, and each call does `mapper.selectByDiscoveryHandlerId(...)` (full list for the handler). For a sync batch of K upstreams that's K full-list queries. Fine for small handlers, but it's a perf cliff for large deployments / bulk re-registration. Suggest loading the handler's existing upstreams once per batch and passing the in-memory list into the matcher.

**One malformed URL aborts the whole sync batch.** In `handleAdded`, `normalizeUrl` throws `IllegalArgumentException` but the only catch is `DuplicateKeyException`, so it propagates and aborts the ADDED loop. In `handleUpdated`/`handleDeleted` the `catch (Exception e) { LOG.error(...); throw e; }` logs then rethrows, also aborting the batch. For a discovery sync listener, one bad URL shouldn't kill the entire batch. Suggest catching `IllegalArgumentException` per-iteration, logging, and continuing.

Minor: `validateIPv6Address` calls `InetAddress.getByName` for bracketed input — fine for IPv6 literals (no DNS), but a bracketed non-IPv6-with-colons string triggers a DNS lookup before throwing. Slow rejection of invalid input, not a correctness issue.

Overall the direction is right; the three items above are the ones I'd want addressed before merge.

wy471x:
> The core bracket-IPv6 path is correct, and the layered defense (skip malformed URLs in `DiscoveryTransfer.mapToCommonUpstream` + `Objects::nonNull` filters in `UpstreamCheckService`) is good. A few things worth addressing:
> 
> **Non-bracketed IPv6 is silently misparsed.** `IpUtils.parseHostPort` falls back to `lastIndexOf(':')` for non-bracketed input. That works for `host:port` and IPv4, but for a non-bracketed IPv6 like `::1` (no port) it produces `host=":"`, `port="1"`, and `2001:db8::1:8080` produces `host="2001:db8::1"`, `port="8080"` (only coincidentally right when the last segment is the port). There's no "more than one colon and not bracketed → reject" guard, so a user who forgets brackets gets a silently broken upstream (garbage host) instead of a clear error. Since `normalizeUrl` then feeds that garbage host into `buildUrl`, you get `[:]`-style nonsense stored. Suggest: if a non-bracketed URL has >1 colon, throw "IPv6 addresses must be bracketed, e.g. [::1]:8080" (or parse it as IPv6+port). file: `IpUtils.parseHostPort`.
> 
> **`matchByHostAndPort` is O(N×K) DB load.** It's called per-upstream inside the `handleAdded`/`handleUpdated`/`handleDeleted` loops, and each call does `mapper.selectByDiscoveryHandlerId(...)` (full list for the handler). For a sync batch of K upstreams that's K full-list queries. Fine for small handlers, but it's a perf cliff for large deployments / bulk re-registration. Suggest loading the handler's existing upstreams once per batch and passing the in-memory list into the matcher.
> 
> **One malformed URL aborts the whole sync batch.** In `handleAdded`, `normalizeUrl` throws `IllegalArgumentException` but the only catch is `DuplicateKeyException`, so it propagates and aborts the ADDED loop. In `handleUpdated`/`handleDeleted` the `catch (Exception e) { LOG.error(...); throw e; }` logs then rethrows, also aborting the batch. For a discovery sync listener, one bad URL shouldn't kill the entire batch. Suggest catching `IllegalArgumentException` per-iteration, logging, and continuing.
> 
> Minor: `validateIPv6Address` calls `InetAddress.getByName` for bracketed input — fine for IPv6 literals (no DNS), but a bracketed non-IPv6-with-colons string triggers a DNS lookup before throwing. Slow rejection of invalid input, not a correctness issue.
> 
> Overall the direction is right; the three items above are the ones I'd want addressed before merge.

Thank you for the detailed code review and constructive suggestions for improvements.

Issues Fixed
                                                                                                                                                                                     
  1. Non-bracketed IPv6 silently misparsed (IpUtils.java)                                                                                                                            
  - parseHostPort used lastIndexOf(':') for non-bracketed input. An unbracketed IPv6 like ::1 produced garbage host : with port 1; 2001:db8::1:8080 coincidentally worked only when  
  the last segment was the port.                                                                                                                                                     
  - Fix: After extracting host via lastColon, check host.indexOf(':') >= 0 — if the host portion still contains colons, it's an unbracketed IPv6. Throw IllegalArgumentException with
   message: "IPv6 addresses must be bracketed, e.g. [::1]:8080".                                                                                                                     
                                                                                                                                                                                     
  2. matchByHostAndPort caused O(N×K) DB queries (CommonUpstreamUtils.java, DiscoveryDataChangedEventSyncListener.java)
  - Called per-upstream inside the handleAdded/Updated/Deleted loops, each call doing mapper.selectByDiscoveryHandlerId(...) — a full-list query. For K upstreams, that's K DB       
  round-trips.                                                                                                                                                                       
  - Fix: Added an overload matchByHostAndPort(List<DiscoveryUpstreamDO>, String) that takes a pre-loaded in-memory list. The sync listener loads the list lazily via a holder array —
   only on the first iteration that actually needs the fallback match, and reuses it for subsequent iterations. When all upstreams match exactly by URL, zero extra queries are made.
                                                                                                                                                                                     
  3. One malformed URL aborted the entire sync batch (DiscoveryDataChangedEventSyncListener.java)
  - handleAdded only caught DuplicateKeyException, so IllegalArgumentException from normalizeUrl propagated and killed the loop. handleUpdated/Deleted caught Exception but re-threw,
   also aborting.                                                                                                                                                                    
  - Fix: handleAdded now catches general Exception (logging and continuing). handleUpdated and handleDeleted removed the throw e — log the error and continue to the next upstream.

Aias00 (approved):
Reviewed #6456 — IPv6-aware upstream URL parsing.

## Verdict: APPROVE

This is a well-structured, well-tested fix. I previously requested changes on this PR (null-url NPE regression + robustness gaps from Copilot). All of those are now resolved in the code, and the changes are correct.

### Prior blocking concerns — resolved

- **Null/blank URL NPE in `nativeCreateOrUpdate` (was my #1 blocker, CI red).** `CommonUpstreamUtils.normalizeUrl` / `IpUtils.parseHostPort` now throw `IllegalArgumentException("Invalid upstream URL, null or blank")` on blank/null instead of NPE-ing deep inside `substring`/`Integer.parseInt`. `DiscoveryUpstreamServiceImplTest.testNativeCreate/testNativeUpdate` now set a real URL, and `DiscoveryDataChangedEventSyncListener` wraps every `handleAdded/Updated/Deleted` call in `try/catch (Exception)` so a malformed URL is logged, never crashes the sync loop.

- **Copilot: `parseHostPort` could throw on malformed input (`substring(1,-1)`, empty port).** Fixed. `parseHostPort` validates `closingBracket < 0` → IAE; defaults empty port to `"80"`; `validateNumericPort` enforces 1–65535 range; `validateIPv6Address` strips zone IDs before `InetAddress.getByName` (no cross-platform `UnknownHostException`). Comprehensive unit tests cover every branch (missing `]`, hostname-in-brackets, non-numeric port, out-of-range, IPv6 zone IDs, etc.).

- **Copilot: `checkUrl` let parse exceptions propagate.** Both the http(s) and bare branches now wrap parsing in `try/catch` and return `false` with a log. Good fail-closed behavior.

- **Copilot: `buildUrl` double-brackets an already-bracketed host.** Fixed — `buildUrl` detects `host.startsWith("[") && endsWith("]")` and uses it verbatim; otherwise brackets it. No `[[::1]]:8080` output.

- **Copilot / `DiscoveryTransfer` raw `NumberFormatException`.** `mapToCommonUpstream` and `mapToDiscoveryUpstreamData` now wrap `parseHostPort` in `try/catch (IllegalArgumentException)` and return `null` for malformed URLs (with a warning), and callers (`UpstreamCheckService`, `LoadServiceDocEntryImpl`, `DiscoveryDataChangedEventSyncListener`) filter nulls. No more unhandled NFE.

### Nice additions
- Backward-compat migration: `matchByHostAndPort` lets old non-normalized records (e.g. `192.168.1.1` vs `192.168.1.1:80`, or unbracketed `2001:db8::1:9090`) be matched and migrated to canonical form on ADD/UPDATE/DELETE. Tested in `DiscoveryUpstreamServiceImplTest` and `DiscoveryDataChangedEventSyncListenerTest`.

### One note (not a blocker)
The current CI run shows **all `build` jobs red**, but the failure is infra, not code: the `curl` download of `maven-mvnd-1.0.2-linux-amd64.zip` from `downloads.apache.org` returned a corrupt/non-zip file (`End-of-central-directory signature not found` → `unzip` exits 9) before any compilation. PR #6429's identical `build (17, ubuntu-latest)` job passed (13m35s) on the same infra, confirming this is a transient mvnd-download flake. **Please just re-run the workflow** once the mirror is reachable — the code itself compiles and the new unit tests are green locally by design.

Approving. Recommend a CI re-run after infra recovers; no code changes required from my side.

Aias00:

PR #6457, opened 2026-07-25
fix(alert): guard against null alert/dateCreated in email alert template

### LINKED ISSUE #6446: [BUG] Email alert reporting can fail with NPE for null or partial AlarmContent
### Current Behavior

`EmailAlertNotifyStrategy.buildAlertHtmlTemplate()` dereferences `alert` before checking whether it is null, and formats `dateCreated` without guarding against a null date.

Evidence:

`shenyu-alert/src/main/java/org/apache/shenyu/alert/strategy/EmailAlertNotifyStrategy.java:75-87`

```java
context.setVariable("content", alert.getContent());
SimpleDateFormat simpleDateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
Date alertTime = alert.getDateCreated();
if (Objects.isNull(alert)) {
    alertTime = new Date();
}
String alarmTime = simpleDateFormat.format(alertTime);
```

`AlertReportController` accepts the request body and dispatches it directly:

`shenyu-admin/src/main/java/org/apache/shenyu/admin/controller/AlertReportController.java:47-48`

```java
public ShenyuAdminResult reportAlert(@Valid @RequestBody final AlarmContent alarmContent) {
    alertDispatchService.dispatchAlert(alarmContent);
```

A null/partial `AlarmContent`, or one with missing `dateCreated`, can cause a `NullPointerException` in the email alert strategy and make `/alert/report` fail with an internal error instead of a controlled validation failure or default timestamp.

### Expected Behavior

Alert reporting should validate missing alert content up front, and the email strategy should only dereference `alert` after the null check. Missing `dateCreated` should either be rejected as invalid input or default to the current time safely.

### Steps To Reproduce

1. Configure an email alert receiver.
2. POST `/alert/report` with a body that omits required alert fields or omits `dateCreated`.
3. The email template path dereferences `alert`/`alert.getDateCreated()` before validating the values.

### Suggested Fix

Move the null check before `alert.getContent()` and `alert.getDateCreated()`. Add bean validation constraints or explicit validation in `AlertReportController`/dispatch service for required `AlarmContent` fields.

wy471x:
Move null check before alert dereferencing in EmailAlertNotifyStrategy to prevent NullPointerException. Add @NotBlank validation on AlertReportRequest title and content fields. Add field by field mapping from AlertReportRequest to AlertContent in the AlertReportController. Add unit tests for these fixes.

Make sure that:

## Summary of Changes                                                                                
                      
  1. AlertReportRequest.java (new, shenyu-admin) — Controller-layer request DTO with @NotBlank on title and content. Validation only applies to external HTTP requests to
  /alert/report, not to internal AlarmContent users.

  2. AlertReportController.java — Changed from @Valid @RequestBody AlarmContent to @Valid @RequestBody AlertReportRequest, with explicit field-by-field mapping into AlarmContent
  before dispatch.

  3. EmailAlertNotifyStrategy.java — Guards dateCreated == null with a fallback to new Date().

  4. AlarmContentTest.java (new) — Builder and getter/setter test for AlarmContent.

  5. EmailAlertNotifyStrategyTest.java (new) — Tests that buildAlertHtmlTemplate does not throw on null dateCreated or null content.

  6. AlertReportControllerTest.java (new) — MockMvc tests covering blank/null title → 400, blank/null content → 400, and valid request → 200 with correct field mapping to
  dispatchAlert.

close [#6446](https://github.com/apache/shenyu/issues/6446)


Aias00:
I found one regression risk outside the email-template null guard.

Adding `@NotBlank` to `AlarmContent.title` and `AlarmContent.content` changes runtime behavior because `AlertReportController` already receives `@Valid @RequestBody AlarmContent`. After this change, null/blank `title` or `content` is rejected with a validation error before alert dispatch.

That can drop existing alerts. For example, the gateway error flow can build an alert from an exception message; an `IllegalArgumentException` can have a null message, and `AlarmSender` copies that value into `AlarmContent.content`. Before this PR the alert could still reach dispatch and the email strategy could render it defensively. With the new shared DTO constraints, the request is rejected at `/alert/report` before any notify strategy sees it.

Could we keep the email `alert == null` / `dateCreated == null` guard, but remove the new `@NotBlank` constraints and the added `jakarta.validation-api` dependency from `shenyu-common`? If stricter validation is desired for external report requests, it would be safer to put it on a controller-layer request DTO and map into `AlarmContent` explicitly.

wy471x:
> I found one regression risk outside the email-template null guard.
> 
> Adding `@NotBlank` to `AlarmContent.title` and `AlarmContent.content` changes runtime behavior because `AlertReportController` already receives `@Valid @RequestBody AlarmContent`. After this change, null/blank `title` or `content` is rejected with a validation error before alert dispatch.
> 
> That can drop existing alerts. For example, the gateway error flow can build an alert from an exception message; an `IllegalArgumentException` can have a null message, and `AlarmSender` copies that value into `AlarmContent.content`. Before this PR the alert could still reach dispatch and the email strategy could render it defensively. With the new shared DTO constraints, the request is rejected at `/alert/report` before any notify strategy sees it.
> 
> Could we keep the email `alert == null` / `dateCreated == null` guard, but remove the new `@NotBlank` constraints and the added `jakarta.validation-api` dependency from `shenyu-common`? If stricter validation is desired for external report requests, it would be safer to put it on a controller-layer request DTO and map into `AlarmContent` explicitly.

@Aias00 Hi, I have fixed the issues you mentioned. Please take a look when you have time.

Aias00 (commented):
Reviewed #6457. The core fix is sound: `EmailAlertNotifyStrategy.buildAlertHtmlTemplate` previously dereferenced `alert.getContent()`/`alert.getDateCreated()` *before* the dead `Objects.isNull(alert)` guard; the new ordering with `dateCreated == null → new Date()` is correct and is reachable via the HTTP report path (a caller omitting `dateCreated`). `EmailAlertNotifyStrategyTest.testNullDateCreatedShouldNotThrowNpe` backs that path. Moving `@NotBlank` from the shared `AlarmContent` DTO to a controller-layer `AlertReportRequest` is the right call — it avoids silently dropping gateway-generated alerts (which bypass the controller via `AlarmSender` → `AlertDispatchService.dispatchAlert`) when content is null.

Non-blocking issues worth addressing before/after merge:

1. **PR description is stale.** The body still claims `@NotBlank` was added to `AlarmContent`, that `jakarta.validation-api` was added to `shenyu-common/pom.xml`, and that `AlarmContentTest` verifies `@NotBlank`. None of that is true in the final diff — `AlarmContent` is unmodified, the only pom change is a trailing-whitespace trim, and `AlarmContentTest` only tests builder/setters. Please rewrite the body to match the merged approach (controller-layer `AlertReportRequest`).

2. **No test backs the new validation or the request→AlarmContent mapping.** `AlertReportController.reportAlert` now manually copies 7 fields and relies on `@NotBlank`, but there is no `AlertReportControllerTest` (the repo has many `*ControllerTest` examples). A `MockMvc` test asserting (a) blank title/body → 400 with the validation message and (b) a valid request maps through to `dispatchAlert` with the right fields would close the gap.

3. **`shenyu-common/pom.xml` whitespace-only hunk** — drop it; it serves no purpose in the final approach.

4. Minor: the `alert == null` branch in `buildAlertHtmlTemplate` is unreachable (`AlertDispatchServiceImpl.DispatchTask` already guards `Objects.nonNull(alert)` before invoking handlers), while `DingTalkRobotAlertNotifyStrategy.send()` still dereferences `alert.getTitle()` unguarded. Either apply the guard consistently or drop the unreachable `alert == null` part and keep only the real `dateCreated == null` fix.

CI note: `it` / `shenyu-integrated-test-combination` failed (other `it` jobs cancelled); `ci` and `e2e` are green. Please confirm the combination failure is unrelated/flaky before merging.

wy471x:
> Reviewed #6457. The core fix is sound: `EmailAlertNotifyStrategy.buildAlertHtmlTemplate` previously dereferenced `alert.getContent()`/`alert.getDateCreated()` _before_ the dead `Objects.isNull(alert)` guard; the new ordering with `dateCreated == null → new Date()` is correct and is reachable via the HTTP report path (a caller omitting `dateCreated`). `EmailAlertNotifyStrategyTest.testNullDateCreatedShouldNotThrowNpe` backs that path. Moving `@NotBlank` from the shared `AlarmContent` DTO to a controller-layer `AlertReportRequest` is the right call — it avoids silently dropping gateway-generated alerts (which bypass the controller via `AlarmSender` → `AlertDispatchService.dispatchAlert`) when content is null.
> 
> Non-blocking issues worth addressing before/after merge:
> 
> 1. **PR description is stale.** The body still claims `@NotBlank` was added to `AlarmContent`, that `jakarta.validation-api` was added to `shenyu-common/pom.xml`, and that `AlarmContentTest` verifies `@NotBlank`. None of that is true in the final diff — `AlarmContent` is unmodified, the only pom change is a trailing-whitespace trim, and `AlarmContentTest` only tests builder/setters. Please rewrite the body to match the merged approach (controller-layer `AlertReportRequest`).
> 2. **No test backs the new validation or the request→AlarmContent mapping.** `AlertReportController.reportAlert` now manually copies 7 fields and relies on `@NotBlank`, but there is no `AlertReportControllerTest` (the repo has many `*ControllerTest` examples). A `MockMvc` test asserting (a) blank title/body → 400 with the validation message and (b) a valid request maps through to `dispatchAlert` with the right fields would close the gap.
> 3. **`shenyu-common/pom.xml` whitespace-only hunk** — drop it; it serves no purpose in the final approach.
> 4. Minor: the `alert == null` branch in `buildAlertHtmlTemplate` is unreachable (`AlertDispatchServiceImpl.DispatchTask` already guards `Objects.nonNull(alert)` before invoking handlers), while `DingTalkRobotAlertNotifyStrategy.send()` still dereferences `alert.getTitle()` unguarded. Either apply the guard consistently or drop the unreachable `alert == null` part and keep only the real `dateCreated == null` fix.
> 
> CI note: `it` / `shenyu-integrated-test-combination` failed (other `it` jobs cancelled); `ci` and `e2e` are green. Please confirm the combination failure is unrelated/flaky before merging.

I have fixed all the issues you mentioned, as well as the CI problems. I have already verified that they pass in my own fork repository. Thank you for the code review.

Aias00:
Good fix — the old null guard was dead code: `Date alertTime = alert.getDateCreated(); if (Objects.isNull(alert)) { alertTime = new Date(); }` checked `alert` (the parameter) for null *after* already calling `alert.getDateCreated()`, so a null `alert` would NPE before reaching the check, and a null `getDateCreated()` on a non-null `alert` was never replaced. The replacement `Objects.isNull(alert.getDateCreated()) ? new Date() : alert.getDateCreated()` correctly guards the right thing.

Moving validation into a dedicated `AlertReportRequest` DTO with `@NotBlank` on `title`/`content` is the right call — it constrains the HTTP boundary without affecting internal callers that build `AlarmContent` directly. Tests cover the 400 paths (blank/null title + content) and the NPE guard for null `dateCreated`, plus a valid-dispatch assertion via `ArgumentCaptor`.

Two minor notes:
- `shenyu-common/pom.xml` has a cosmetic whitespace-only change (a blank line removed before `</dependencies>`) that looks unintentional — fine to drop or keep, just flagging.
- The `/alert/report` endpoint's auth posture is unchanged by this PR (validation improved, but anyone who could post before still can). Out of scope here, just noting it's a pre-existing consideration.

Aias00 (approved):
## Summary

This PR hardens the `/alert/report` endpoint: it validates the incoming payload and fixes a NullPointerException in the email alert template rendering.

### What is correct
- `AlertReportController.reportAlert` now accepts a dedicated `AlertReportRequest` DTO (with `@NotBlank` on `title` and `content`) and maps it field-by-field onto the internal `AlarmContent` via its builder. This keeps validation concerns on the public API boundary while leaving internal `AlarmContent` callers untouched.
- `EmailAlertNotifyStrategy.buildAlertHtmlTemplate` had a real bug: it dereferenced `alert.getContent()` and `alert.getDateCreated()` and *then* checked `Objects.isNull(alert)` — a null `alert` would already have thrown, and a merely null `dateCreated` was never guarded. It is now `Objects.isNull(alert.getDateCreated()) ? new Date() : alert.getDateCreated()`, which correctly falls back when the date is missing.
- The controller uses `@Valid @RequestBody`, and admin already depends on `spring-boot-starter-validation`, so blank/empty title or content is rejected with HTTP 400.

### Verification
- `AlertReportControllerTest` asserts 400 for blank title, null title, blank content, and null content, and asserts a valid payload dispatches an `AlarmContent` with the expected fields.
- `EmailAlertNotifyStrategyTest` (new) covers null `dateCreated`, valid alert, and null `content` — all render without throwing.
- `AlarmContentTest` (new) covers the builder and getters/setters.

### Non-blocking suggestions
- `AlertReportRequest` only maps the fields it declares; any extra `AlarmContent` fields previously accepted by the old `AlarmContent`-typed body are now dropped (Jackson ignores unknown properties). Please confirm no alert producer relies on sending additional `AlarmContent` fields through this endpoint.
- `level` (byte) has no validation; a missing `level` defaults to `0`. That matches prior behavior, so not a regression, but worth a sanity check on downstream level handling.

### Conclusion
Approved. Both the input validation and the NPE fix are correct and well covered by new tests.

```bash
gh pr review 6457 --repo apache/shenyu --approve --body-file review_6457.md
```

dengliming:
@wy471x Please resolve conflicts.

Aias00 (changes requested):

Aias00 (approved):
Approving — two genuine NPE fixes plus properly scoped request validation.

**1. `EmailAlertNotifyStrategy#buildAlertHtmlTemplate`.** The old guard was inverted and useless:

```java
Date alertTime = alert.getDateCreated();   // dereferences alert
if (Objects.isNull(alert)) {               // ...then checks whether it was null
    alertTime = new Date();
}
```

Two problems: the null check happens after the dereference (so a null `alert` NPEs before the guard can help), and the guard tests the wrong thing anyway — the field that can legitimately be null is `dateCreated`, whose null then reaches `simpleDateFormat.format(null)` and throws. `Objects.isNull(alert.getDateCreated()) ? new Date() : alert.getDateCreated()` is correct. `testNullDateCreatedShouldNotThrowNpe` pins it.

**2. Introducing `AlertReportRequest` instead of annotating `AlarmContent`.** This is the right call: `AlarmContent` lives in `shenyu-common` and is constructed directly by internal producers (the alert pipeline), not only by this endpoint — putting `@NotBlank` on it would have made internal callers subject to HTTP-layer validation. The DTO keeps the constraint at the API boundary and the controller maps it onto `AlarmContent` explicitly. The wire format is unchanged, so existing clients are unaffected.

Tests are good: four 400 cases (blank/null title, blank/null content) plus a happy path that captures the dispatched `AlarmContent` and verifies every field including `namespaceId` and `labels` — which confirms the mapping isn't dropping anything.

Two follow-ups (non-blocking):

1. `buildAlertHtmlTemplate` still calls `alert.getContent()` and `alert.getDateCreated()` unguarded, so a null `alert` would still NPE. `send(...)` appears to guarantee non-null, but if that's an invariant worth enforcing, an `Objects.requireNonNull` (or an early return) would make it explicit.
2. The controller tests assert `status().isBadRequest()` via `MockMvcBuilders.standaloneSetup`. ShenYu's admin layer has global exception handlers that conventionally return HTTP 200 with an error message body, so the real response for an invalid payload may differ from 400. The standalone setup doesn't include those handlers, so the test is valid for what it covers — just worth confirming what the admin UI actually receives.

PR #6528, opened 2026-07-31
fix(admin): persist upstream status changes into selector handle

### LINKED ISSUE #6522: [BUG] RPC client registration does not persist upstream status changes in selector handles
### Search before asking

### Apache ShenYu Component

shenyu-admin

### What happened

The Dubbo, gRPC, and Tars client registration services detect upstream status changes, but they do not apply those changes to the `existList` that is serialized back into the selector handle.

For example, `ShenyuClientRegisterGrpcServiceImpl.buildHandle(...)` does this:

```java
List<GrpcUpstream> existList = GsonUtils.getInstance().fromCurrentList(selectorDO.getHandle(), GrpcUpstream.class);
...
List<GrpcUpstream> diffStatusList = addList.stream().filter(upstream -> !upstream.isStatus()
        || existList.stream().anyMatch(e -> e.equals(upstream) && e.isStatus() != upstream.isStatus())).collect(Collectors.toList());
if (CollectionUtils.isNotEmpty(diffStatusList)) {
    canAddList.addAll(diffStatusList);
}
...
return GsonUtils.getInstance().toJson(CollectionUtils.isEmpty(existList) ? canAddList : existList);
```

The same pattern exists in the Dubbo and Tars registration services.

The upstream `equals(...)` methods compare identity fields such as host/protocol/url/namespace, but they do not include `status`. Therefore a status-only change is detected in `diffStatusList`, but the matching object already in `existList` keeps its old status. Since the method serializes `existList` when it is not empty, the selector handle written back to DB still contains the stale status.

This affects status changes such as a single-node deleted/offline event where the new upstream is marked `status=false` before diffing.

### Expected behavior

When an existing upstream is re-registered with a different status, the matching item inside `existList` should be updated before the handle is serialized. The persisted selector handle should reflect the same status change that the registration flow detected.

### How to reproduce

1. Register a Dubbo, gRPC, or Tars upstream so the selector handle contains the upstream with `status=true`.
2. Send a registration event for the same upstream with `EventType.DELETED`, causing the newly built upstream to be marked `status=false`.
3. `diffStatusList` detects the changed status and adds it to `canAddList`.
4. Because `existList` is not updated, the method returns `GsonUtils.toJson(existList)` with the original `status=true` value.
5. The DB selector handle remains stale, so admin display/export/restart replay can use the old upstream status.

### Debug logs

_No response_

### Environment

Current `master` branch.

Duansg:
### What changed

Close #6522

### Problem

`buildHandle` in the divide / dubbo / grpc / tars client register services detects status-only changes into `diffStatusList`, but never applies them to `existList` — the upstream `equals()` methods compare identity fields only (host, protocol, url,
namespaceId), so a status-changed upstream is still treated as an existing one and keeps its old status. `existList` is then serialized back into the selector handle, so the change is lost.

Reproduce: register an upstream with `status=true`, then send a DELETED event for it. The change is detected but the DB selector handle still holds `status=true`.

This is always hit with the default `shenyu.upstream.check=false`, where the handle is written purely from `buildHandle`'s return value. With health check enabled the path is masked by `UpstreamCheckService.updateSelectorHandler`.

Note: the issue lists dubbo/grpc/tars; `ShenyuClientRegisterDivideServiceImpl` shares the same code and is fixed as well.

### Fix

Add `syncUpstreamStatus(existList, diffStatusList)` to `AbstractShenyuClientRegisterServiceImpl` and call it from the four implementations, writing the new status back onto the matching entry in `existList`. No change to `equals()`/`hashCode()` — those are relied on by `UPSTREAM_MAP`, `ZOMBIE_SET` and the `removeAll` paths.

Going back online is symmetric: the client re-registers with `status=true`, `diffStatusList` matches again and `existList` is restored.

### Tests

Added `testBuildHandleWithStatusChanged` to the four register service tests: a handle holding two live upstreams plus a DELETED event for one of them must yield `status=false` for that node and `status=true` for the other, with the list size unchanged. All four fail without the production change.

`mvn test -pl shenyu-admin` (4 classes, 28 cases) and `mvn checkstyle:check` pass.

### DIFF
--- a/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/register/AbstractShenyuClientRegisterServiceImpl.java
+++ b/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/register/AbstractShenyuClientRegisterServiceImpl.java
@@ -361,6 +361,24 @@ public abstract class AbstractShenyuClientRegisterServiceImpl extends FallbackSh
     }
 
     /**
+     * Sync the changed status back to the exist list, so that the status change
+     * can be persisted into the selector handle.
+     *
+     * <p>The upstream {@code equals} methods only compare the identity fields, so an upstream whose
+     * status just changed is still regarded as an existing one and keeps its old status in the exist list.
+     *
+     * @param existList      the upstream list parsed from the selector handle
+     * @param diffStatusList the upstream list whose status has changed
+     * @param <T>            the upstream type
+     */
+    protected <T extends CommonUpstream> void syncUpstreamStatus(final List<T> existList, final List<T> diffStatusList) {
+        for (T changed : diffStatusList) {
+            existList.stream().filter(exist -> exist.equals(changed))
+                    .forEach(exist -> exist.setStatus(changed.isStatus()));
+        }
+    }
+
+    /**
      * Build context path default rule dto rule dto.
      *
      * @param selectorId  the selector id
--- a/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/register/ShenyuClientRegisterDivideServiceImpl.java
+++ b/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/register/ShenyuClientRegisterDivideServiceImpl.java
@@ -92,6 +92,7 @@ public class ShenyuClientRegisterDivideServiceImpl extends AbstractContextPathRe
                     || existList.stream().anyMatch(e -> e.equals(upstream) && e.isStatus() != upstream.isStatus())).collect(Collectors.toList());
             if (CollectionUtils.isNotEmpty(diffStatusList)) {
                 canAddList.addAll(diffStatusList);
+                syncUpstreamStatus(existList, diffStatusList);
             }
         }
 
--- a/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/register/ShenyuClientRegisterDubboServiceImpl.java
+++ b/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/register/ShenyuClientRegisterDubboServiceImpl.java
@@ -84,6 +84,7 @@ public class ShenyuClientRegisterDubboServiceImpl extends AbstractShenyuClientRe
                     || existList.stream().anyMatch(e -> e.equals(upstream) && e.isStatus() != upstream.isStatus())).collect(Collectors.toList());
             if (CollectionUtils.isNotEmpty(diffStatusList)) {
                 canAddList.addAll(diffStatusList);
+                syncUpstreamStatus(existList, diffStatusList);
             }
         }
 
--- a/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/register/ShenyuClientRegisterGrpcServiceImpl.java
+++ b/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/register/ShenyuClientRegisterGrpcServiceImpl.java
@@ -107,6 +107,7 @@ public class ShenyuClientRegisterGrpcServiceImpl extends AbstractShenyuClientReg
                     || existList.stream().anyMatch(e -> e.equals(upstream) && e.isStatus() != upstream.isStatus())).collect(Collectors.toList());
             if (CollectionUtils.isNotEmpty(diffStatusList)) {
                 canAddList.addAll(diffStatusList);
+                syncUpstreamStatus(existList, diffStatusList);
             }
         }
         if (doSubmit(selectorDO.getId(), canAddList)) {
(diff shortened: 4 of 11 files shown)


Aias00 (commented):
Reviewed #6528 — fix(admin): persist upstream status changes into selector handle.

Verified the fix and the four new tests — they are real regression tests (they fail without `syncUpstreamStatus` because `existList[0].status` stays `true`). The root-cause analysis is correct: `CommonUpstream.equals()` (omits `status`) means a status-changed upstream still matches its old entry in `existList`, so the old status was serialized back into the handle. The new helper writes the new status onto the matching entry before serialization, which is the right minimal fix. The symmetric `false -> true` (re-register) path also works via the same predicate. CI is green.

One scope question worth a conscious maintainer call:

- **`ShenyuClientRegisterWebSocketServiceImpl.buildHandle` (lines 78-96) has the same — actually worse — bug and is not touched by this PR.** It never computes `diffStatusList` at all, only `diffList` for new upstreams, so a DELETED/offline status change is lost from the WebSocket selector handle exactly as in #6522. The new `syncUpstreamStatus` helper alone wouldn't fix it (it needs the `diffStatusList` detection block first), so it's a slightly larger change. Suggest either folding it in here for completeness or opening a follow-up issue so it isn't lost.

Minor nits (not blocking):
- The PR body claims the "going back online" symmetric path works, but no test covers `status=false -> true`. One extra test (seed the handle with `status=false`, re-register, assert `true`) would close the loop.
- Pre-existing: `isEventDeleted = uriList.size() == 1 && EventType.DELETED` in Dubbo/Grpc/Tars means batch DELETED events don't get `status=false` set, so `syncUpstreamStatus` won't flag them either. Divide is unaffected (per-upstream `eventType` in the builder). Worth a separate follow-up.

Duansg:
@Aias00  Thanks for the detailed review. I have updated the patch with:

- false -> true recovery tests for Divide, Dubbo, gRPC, and Tars;
- the missing diffStatusList detection and syncUpstreamStatus logic for WebSocket;
- a WebSocket false -> true regression test.

I intentionally did not include the batch DELETED change after tracing the complete shutdown flow. It is not safe to fix independently:

1. For discovery-driven plugins such as Divide, gRPC, and WebSocket, runtime availability is determined by the discovery-upstream row rather than the selector handle. Simply routing DELETED through registerURI() would write status=false into the handle, while doDiscoveryLocal() would leave an existing discovery row at status=0, or recreate it with status=0 after deletion. The gateway could therefore continue routing traffic to the closed instance.

2. The same JVM shutdown currently emits both signals: the priority-2 hook sends OFFLINE, while the priority-1 hook sends DELETED. Their existing semantics are different, so handling DELETED as a status update while OFFLINE removes data would produce competing state transitions during the same shutdown.

3. Dubbo and Tars also have different runtime consumption paths, so a complete fix needs a per-protocol review of the authoritative upstream state. Because of this, the batch DELETED behavior requires a separate follow-up that first defines the relationship between DELETED and OFFLINE, and then updates both selector and discovery state consistently. I kept this PR focused on correcting status reconciliation inside buildHandle and covering the symmetric recovery path.

Aias00:
Clean, well-targeted fix — `syncUpstreamStatus` correctly uses the identity-only `equals()` to locate the existing upstream and then writes the changed status back into `existList` before it's serialized into the selector handle, which is exactly the gap that lost status changes (#6522). The WebSocket service refactor is a nice consistency win too — it previously had no `diffStatusList` handling at all, so it was silently dropping status changes that Divide/Dubbo/Grpc/Tars already handled; now all five behave the same way, and the return-path refactor (`toJson(isEmpty(existList) ? canAddList : existList)`) preserves the old empty/non-empty semantics.

Test coverage is strong — two scenarios per protocol (online→offline via `EventType.DELETED`, and offline→online recovery), each asserting only the intended upstream's status flips.

Minor: `syncUpstreamStatus` is O(n×m) (stream over `existList` per changed upstream), but upstream lists are small so this is fine; just noting it's quadratic if lists ever grow. No action needed.

Aias00 (approved):
## Summary

This PR fixes issue #6522: upstream status changes (online/offline) detected during client registration were stored in `diffStatusList` but never written back into the selector handle, so the persisted handle kept stale statuses.

### What is correct
- New `AbstractShenyuClientRegisterServiceImpl.syncUpstreamStatus(existList, diffStatusList)` mutates the matching existing upstream in `existList` to the changed status (matched by `equals`, which compares only identity fields). Because `buildHandle` returns the (mutated) `existList`, the corrected status is persisted into the selector handle.
- The call is wired into `divide`, `dubbo`, `grpc`, `tars`, and `websocket` `buildHandle` implementations when a `diffStatusList` is present.
- The websocket path is also refactored to compute `diffStatusList` and call `syncUpstreamStatus`, and to return the serialized `existList` (which now reflects the synced statuses) instead of the stale pre-mutation `handleAdd`.

### Verification
- Tests `testBuildHandleWithStatusChanged` and `testBuildHandleWithStatusRecovered` for divide/dubbo/grpc/tars assert that, after a status flip (offline→online and online→offline), the returned handle contains the updated status for the changed upstream and leaves the other one untouched.
- `ShenyuClientRegisterWebSocketServiceImplTest.testBuildHandleWithStatusRecovered` covers the websocket recovery path.

### Non-blocking suggestions
- The websocket suite only covers status *recovery* (offline→online), not the offline direction. Adding the opposite case would mirror the other four services, though the shared `syncUpstreamStatus` logic is already exercised by them.
- `syncUpstreamStatus` only updates `status`; if other mutable fields (e.g. weight) can change for an "existing" upstream without changing identity, those would still be missed. Today only status is tracked via `diffStatusList`, so this is fine.

### Conclusion
Approved. The fix correctly propagates status changes into the persisted handle across all five RPC register services, with solid regression tests.

```bash
gh pr review 6528 --repo apache/shenyu --approve --body-file review_6528.md
```

Aias00 (approved):
Approved as PMC (Aias00). Passing CI, mergeable, small targeted fix. Review only - not merged.

PR #6529, opened 2026-08-01
fix: Non-default HTTP retry backoff strategies cannot be selected

### LINKED ISSUE #6464: [BUG] Non-default HTTP retry backoff strategies cannot be selected

### Current Behavior

The HTTP client plugin appears to support multiple retry backoff strategies (`default`, `fixed`, `exponential`, `custom`), but current master cannot select any non-default strategy through built-in code.

`AbstractHttpClientPlugin` reads the strategy type from an exchange attribute:

```java
String retryStrategyType = (String) Optional
    .ofNullable(exchange.getAttribute(Constants.HTTP_RETRY_BACK_OFF_SPEC))
    .orElse(HttpRetryBackoffSpecEnum.getDefault());
```

However `Constants.HTTP_RETRY_BACK_OFF_SPEC` is currently defined as the literal string `"default"`, so it is used as the exchange attribute key. I could not find any current code path that writes an exchange attribute with that key. As a result, the code always falls back to `HttpRetryBackoffSpecEnum.getDefault()` and the `fixed` / `exponential` / `custom` branches are effectively unreachable from built-in configuration.

There is also a hard failure if anything does force the `custom` branch: `CustomRetryStrategy#execute(...)` currently returns `null`, and `AbstractHttpClientPlugin` immediately calls methods on the returned `Mono`, causing a `NullPointerException`.

### Expected Behavior

Non-default retry backoff strategies should either be configurable and stored in the exchange under a real attribute key, or the dead strategy branches should be removed/disabled until they are supported.

The `custom` strategy should not return `null`; if it is unsupported, it should fail explicitly or be excluded from selection.

### Steps To Reproduce

1. Configure an HTTP route with retry enabled.
2. Try to select a non-default backoff strategy such as `fixed` or `exponential` through the available rule/configuration path.
3. The HTTP client still uses the default strategy because no built-in code writes `exchange.setAttribute(Constants.HTTP_RETRY_BACK_OFF_SPEC, "fixed")` or similar.
4. If an extension manually writes the attribute value `custom`, `CustomRetryStrategy#execute(...)` returns `null` and the request fails with NPE.

### Environment

```markdown
```

### Debug logs

_No response_

### Anything else?

- `shenyu-common/src/main/java/org/apache/shenyu/common/constant/Constants.java`
- `shenyu-plugin/shenyu-plugin-httpclient/src/main/java/org/apache/shenyu/plugin/httpclient/AbstractHttpClientPlugin.java`
- `shenyu-plugin/shenyu-plugin-httpclient/src/main/java/org/apache/shenyu/plugin/httpclient/CustomRetryStrategy.java`

hengyuss:
Fixes #6464 

Make sure that:


Aias00 (commented):
Reviewed #6529 — fixes #6464. The root-cause analysis is correct and the fix is sound: `Constants.HTTP_RETRY_BACK_OFF_SPEC` was `"default"`, used as the exchange attribute *key*, so nothing ever wrote under it and `AbstractHttpClientPlugin` always fell back to `DefaultRetryStrategy`, leaving `fixed`/`exponential`/`custom` unreachable. Renaming the key to `"httpRetryBackOffSpec"`, adding `DivideRuleHandle.retryBackOffSpec`, populating it in `DividePlugin`, switching the `switch` to `HttpRetryBackoffSpecEnum`, and replacing `CustomRetryStrategy`'s `return null` with an explicit `Mono.error` all address the reported defects. Backward compatibility is fine: existing rule-handle JSON lacks the new field, Gson leaves it at the field initializer (`"default"`), so existing configs keep routing to `DefaultRetryStrategy`. DB IDs are collision-free and consistent across all 6 init dialects and 5 upgrade scripts. The constant-value change only affects an in-memory attribute key (no persisted JSON key changes; only 3 Java consumers).

Non-blocking, but worth addressing before merge:

1. (should_fix) No regression test for the actual bug. `DivideRuleHandleTest.testGetterSetter` covers every sibling field except `retryBackOffSpec`; `DividePluginTest` doesn't assert `Constants.HTTP_RETRY_BACK_OFF_SPEC` is placed into the exchange from `ruleHandle.getRetryBackOffSpec()`; `RetryStrategyTest` exercises strategies in isolation but not the selection switch in `AbstractHttpClientPlugin.execute`. The whole point of the PR — that a configured non-default spec is actually selected — is untested, so #6464 can regress silently. Suggested additions: (a) a `DividePluginTest` case asserting the attribute is set from the rule handle (mirroring the `HTTP_RETRY`/`RETRY_STRATEGY` puts); (b) `HttpRetryBackoffSpecEnumTest` covering `acquireByName` for each name + null + unknown → `DEFAULT_BACKOFF` (every other enum in the package has one); (c) extend `testGetterSetter` to set/assert `retryBackOffSpec`.

2. (nit) The new `retryBackOffSpec` `plugin_handle` row uses `sort=0`, identical to `retryStrategy`, so the two fields' admin-UI order is non-deterministic. Give it a distinct sort (e.g. 1).

3. (nit / open question) `CustomRetryStrategy` now throws `UnsupportedOperationException("Please implement CustomRetryStrategy via SPI")`. Returning an explicit error is a clear improvement over `return null`, but I couldn't find an SPI/registry for this strategy in the diff or under `shenyu-plugin-httpclient`. If no extension point exists, consider hiding `custom` from the `RETRY_BACKOFF_SPEC` dict until one does, since selecting it today always yields a 503.

Note: PR is in `BEHIND` merge state (needs a rebase onto master before merge).

Aias00:
Solid, well-tested fix — the string→enum switch in `AbstractHttpClientPlugin` is cleaner, `acquireByName` gracefully falls back to `DEFAULT_BACKOFF` for null/empty/unknown (covered by `HttpRetryBackoffSpecEnumTest`), and `DivideRuleHandle.equals/hashCode/toString` are correctly updated (plus a nice `toString` comma fix). Removing `CustomRetryStrategy` is safe — it was a stub whose `execute(...)` returned `null`, and the only caller (`AbstractHttpClientPlugin:94`) is rewritten in this PR; no other references remain. Backward compat looks fine: existing rules without `retryBackOffSpec` deserialize to the field default `"default"`, which matches the pre-PR behavior (the attribute was previously null → `Optional.orElse(getDefault())` → `DefaultRetryStrategy`).

One minor inconsistency: the SQL scripts add a `custom` dict entry (`'custom','custom','custom',3,0`, `enabled=0`) across all dialects, but `HttpRetryBackoffSpecEnum` no longer has `CUSTOM_BACKOFF` (removed in this PR). So `custom` is now a dead/disabled dict value — `acquireByName("custom")` falls through to `DEFAULT_BACKOFF`. Since the entry is `enabled=0` it won't surface in the admin UI, so there's no runtime impact, but you could either drop the `custom` dict rows or keep the enum value to avoid the mismatch. Just flagging.

Aias00 (dismissed):
## Summary

This PR makes the HTTP retry backoff strategy selectable via `DivideRuleHandle.retryBackOffSpec` (default / fixed / exponential) instead of the previously hard-wired, non-selectable behavior.

### What is correct
- `DivideRuleHandle` gains a `retryBackOffSpec` field (default `HttpRetryBackoffSpecEnum.getDefault()`), and `equals`/`hashCode`/`toString` are updated consistently so handle serialization round-trips correctly.
- `DividePlugin.doExecute` puts `HTTP_RETRY_BACK_OFF_SPEC` into the exchange attributes (defaulting to `getDefault()` when empty), and `AbstractHttpClientPlugin.execute` now resolves the strategy via `HttpRetryBackoffSpecEnum.acquireByName(...)` instead of a brittle string `switch`. Unknown/empty values fall back to `DEFAULT_BACKOFF`.
- The dead `CustomRetryStrategy` class (whose `execute` returned `null`) and the `CUSTOM_BACKOFF` enum value are removed. `AbstractHttpClientPlugin` no longer references them.
- DB changes are complete: `plugin_handle` + `shenyu_dict` rows for `retryBackOffSpec` are added to mysql/ob/og/oracle/pg init scripts **and** to the `2.7.0-upgrade-2.7.1-*` upgrade scripts, plus the h2 test schema.
- Tests added: `HttpRetryBackoffSpecEnumTest`, `DivideRuleHandleTest` getter/setter, and `DividePluginTest` asserting both the explicit value and the default attribute are propagated.

### Verification
- A repo-wide search for `CustomRetryStrategy` / `CUSTOM_BACKOFF` on `master` returns only the two files this PR itself modifies/deletes (`CustomRetryStrategy.java`, `AbstractHttpClientPlugin.java`, `HttpRetryBackoffSpecEnum.java`). No other module references the removed symbols, so the deletion does not leave dangling compiles.
- `Constants.HTTP_RETRY_BACK_OFF_SPEC` was renamed from the misleading literal `"default"` to `"httpRetryBackOffSpec"`; both producer (`DividePlugin`) and consumer (`AbstractHttpClientPlugin`) use the same constant, and the attribute is a per-request exchange key (never persisted), so there is no external/serialized contract break.

### Non-blocking suggestions
- The `2.7.0-upgrade-2.7.1-*.sql` inserts are plain `INSERT`s (not `INSERT IGNORE`). This matches the surrounding entries in those scripts, but if an operator re-runs an upgrade script it would now duplicate-key. Worth a follow-up to make the new rows idempotent, consistent with how other ShenYu upgrades eventually converge.
- Please confirm the admin metadata-sync path exposes `retryBackOffSpec` as a configurable field for the divide plugin so operators can actually select it (the schema is in place; this is just the admin UI/metadata wiring).

### Conclusion
Approved. The implementation is coherent, the removal is safe, and the DB coverage is thorough across all supported dialects.

```bash
gh pr review 6529 --repo apache/shenyu --approve --body-file review_6529.md
```

dengliming (changes requested):
`loadBalancerFactoryMockedStatic` is being closed twice here: once explicitly after the mocked call path, and then again after the `assertThrows` branch. That double-close triggers Mockito's static mock lifecycle error and is the direct cause of the failing CI job.

Could we either remove the second `close()` or wrap this mock in try-with-resources so it is guaranteed to be closed exactly once?

dengliming (approved):

PR #6530, opened 2026-08-01
fix: Invalid discoveryHandlerId can persist broken discovery upstream rows

### LINKED ISSUE #6466: [BUG] Invalid discoveryHandlerId can persist broken discovery upstream rows

### Current Behavior

`/discovery-upstream` accepts `DiscoveryUpstreamDTO.discoveryHandlerId` with only `@NotBlank` validation. It does not verify that the referenced discovery handler exists.

The create path inserts the row first and then refreshes discovery data:

```java
fetchAll(discoveryUpstreamDTO.getDiscoveryHandlerId());
```

`fetchAll(...)` then dereferences the handler returned by `discoveryHandlerMapper.selectById(discoveryHandlerId)` without a null check:

```java
DiscoveryHandlerDO discoveryHandlerDO = discoveryHandlerMapper.selectById(discoveryHandlerId);
...
SelectorDO selectorDO = selectorMapper.selectByDiscoveryHandlerId(discoveryHandlerDO.getId());
...
DiscoveryDO discoveryDO = discoveryMapper.selectById(discoveryHandlerDO.getDiscoveryId());
```

The `discovery_upstream` schema has a unique key on `(discovery_handler_id, upstream_url)`, but no foreign key to `discovery_handler`. Therefore a request with a non-existent `discoveryHandlerId` can insert an orphan upstream row and then fail with a server error during refresh.

### Expected Behavior

`discoveryHandlerId` should be validated before insert/update, or `fetchAll(...)` should fail gracefully before data is persisted.

If the handler does not exist, the API should return a validation error and should not leave a broken `discovery_upstream` row behind.

### Steps To Reproduce

1. Call `POST /discovery-upstream` with a valid `namespaceId`, valid URL/status/weight fields, but a `discoveryHandlerId` that does not exist.
2. The DTO passes validation because `discoveryHandlerId` is nonblank.
3. `DiscoveryUpstreamServiceImpl#create(...)` inserts the row.
4. `fetchAll(...)` dereferences a null `DiscoveryHandlerDO`, returning a 500 and leaving inconsistent data.

### Environment

```markdown
```

### Debug logs

_No response_

### Anything else?

- `shenyu-admin/src/main/java/org/apache/shenyu/admin/model/dto/DiscoveryUpstreamDTO.java`
- `shenyu-admin/src/main/java/org/apache/shenyu/admin/controller/DiscoveryUpstreamController.java`
- `shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/DiscoveryUpstreamServiceImpl.java`
- `db/init/mysql/schema.sql`

hengyuss:
Invalid discoveryHandlerId can persist broken discovery upstream rows
Fixes: #6466 

Make sure that:


Aias00 (commented):
Reviewed #6530. The fix correctly closes the reported path in #6466 (POST `/discovery-upstream` with a non-existent `discoveryHandlerId` → orphan row + NPE in `fetchAll`). It reuses the existing `@Existed`/`DiscoveryHandlerMapper.existed` validator the repo already uses for the `id` and `namespaceId` fields and for the PUT path variable, so it's consistent and correct for the HTTP create/update paths. No blocker from me.

Two non-blocking items:

1. **No regression test (should_fix).** Issue #6466 ships a 4-step repro and the PR adds no test. The H2 `discovery_handler` table (`schema.sql:1249`), `AbstractSpringIntegrationTest` (`@ActiveProfiles("h2")`), and `DiscoveryUpstreamMapperTest.existed()` already provide the harness. Please add a controller/validation test that POSTs a `DiscoveryUpstreamDTO` with a non-existent `discoveryHandlerId` and asserts a 400 (plus a happy-path 200 with an existing handler) — otherwise a future revert/refactor of this annotation regresses silently. The PR checklist also leaves "submit test cases" unchecked.

2. **Update endpoint behavior (nit).** `PUT /discovery-upstream/{discoveryHandlerId}` now `@Existed`-validates the `discoveryHandlerId` inside each body element, but `DiscoveryUpstreamServiceImpl.updateBatch` (`:115`) overwrites it with the path id and ignores the body value. If the admin frontend sends a placeholder/mismatched id in the PUT body, this becomes a new 400. Please confirm the frontend sends the matching path id in each body element; if so, no action needed.

Out of scope (follow-up, not for this PR): the import/programmatic paths (`DiscoveryUpstreamServiceImpl.importData` at `:255`/`:293`, `nativeCreateOrUpdate` at `:123`, fed by Gson in `ConfigsServiceImpl:440` and `DiscoveryUpstreamDataConfigsExportImportHandler:59`) bypass `@Valid` and can still insert orphan rows. They don't call `fetchAll`, so they dodge the NPE in #6466 — hence not blocking here — but a service-level existence check would fully match the issue title's intent.

Aias00:
Clean defensive fix — adding `@Existed(provider = DiscoveryHandlerMapper.class)` on `discoveryHandlerId` rejects invalid handler ids at the validation boundary instead of persisting broken rows (the #6466 root cause). Two things worth confirming:

Is the validation actually exercised on every write path that can persist a `DiscoveryUpstreamDTO`? `@Existed` is a bean-validation annotation, so it only fires when the DTO passes through a `@Valid` controller method (or a programmatic `Validator` call). The config-import path in `DiscoveryUpstreamServiceImpl.importData(...)` takes DTOs and inserts them directly — if that path doesn't go through bean validation, the annotation won't catch broken imports there, and the `getOrDefault(id, id)` guard from #6532 remains the only protection on that flow. Worth a note on which paths are covered.

No test (the body's "submit test cases" checkbox is unchecked). A test passing a non-existent `discoveryHandlerId` through a `@Valid` endpoint and asserting a validation failure would pin the behavior.

Aias00 (approved):
## Summary

This PR prevents broken discovery-upstream rows caused by an invalid `discoveryHandlerId` by validating the id against `DiscoveryHandlerMapper` at the API boundary.

### What is correct
- `DiscoveryUpstreamDTO.discoveryHandlerId` now carries `@Existed(provider = DiscoveryHandlerMapper.class, message = "discoveryHandlerId is not existed")`, in addition to the existing `@NotBlank`.
- The `DiscoveryUpstreamController` already applies `@Valid @RequestBody` on `createDiscoveryUpstream` (line 55), `createDiscoveryUpstreamList` (line 67) and `updateDiscoveryUpstream` (line 87), so the `@Existed` constraint is actually enforced — verified in the source.
- `DiscoveryHandlerMapper.existed(Serializable id)` exists, so the validation provider resolves correctly.
- A new `DiscoveryUpstreamControllerTest` asserts that a non-existent handler id is rejected with the expected message, and that valid ids succeed.

### Verification
- Controller endpoints confirmed to use `@Valid`; the test mocks `SpringBeanUtils.getBean(DiscoveryHandlerMapper.class)` and `discoveryHandlerMapper.existed(...)` and expects the `"discoveryHandlerId is not existed"` message — the wiring is exercised end-to-end.

### Non-blocking suggestions
- Consider extending the validation to the `batch` create path as well (it already has `@Valid`, so it is covered) and to any import endpoints that build `DiscoveryUpstreamDTO` directly, to keep the guard consistent.

### Conclusion
Approved. This closes the gap where an invalid handler id could persist a row that can never be resolved.

```bash
gh pr review 6530 --repo apache/shenyu --approve --body-file review_6530.md
```

PR #6532, opened 2026-08-01
fix: Discovery_upstream_import_checks_duplicates_before_remapping_handler_ids

### LINKED ISSUE #6465: [BUG] Discovery upstream import checks duplicates before remapping handler ids

### Current Behavior

Namespace import remaps exported discovery handler ids to target-namespace handler ids in `DiscoveryServiceImpl#importData(...)`:

```java
discoveryHandlerIdMapping.put(discoveryHandlerDO.getId(), discoveryHandlerId);
discoveryHandlerDO.setId(discoveryHandlerId);
```

But `DiscoveryUpstreamServiceImpl#importData(...)` checks duplicate upstream URLs before applying that mapping:

```java
String discoveryHandlerId = discoveryUpstreamDTO.getDiscoveryHandlerId();
Set<String> existsUpstreamUrlSet = discoveryHandlerUpstreamMap
    .getOrDefault(discoveryHandlerId, Lists.newArrayList())
    .stream()
    .map(DiscoveryUpstreamDO::getUpstreamUrl)
    .collect(Collectors.toSet());

if (existsUpstreamUrlSet.contains(url)) {
    ...
    continue;
}

discoveryUpstreamDTO.setDiscoveryHandlerId(discoveryHandlerIdMapping.get(discoveryUpstreamDTO.getDiscoveryHandlerId()));
```

`discoveryHandlerUpstreamMap` is grouped by target database handler ids, but the lookup uses the exported/source handler id. So an existing upstream under the remapped target handler is not detected. The insert can then hit the database unique key `(discovery_handler_id, upstream_url)`.

Because this import method is not transactional, a duplicate-key error can also leave earlier imported upstream rows committed while later rows fail.

### Expected Behavior

The import flow should remap `discoveryHandlerId` before duplicate lookup and insert, then check duplicates against the target handler id.

Imports should either skip/report duplicate upstreams consistently or run in a transaction so partial imports are not left behind after an insert failure.

### Steps To Reproduce

1. Export/import discovery data into a namespace where the corresponding target discovery handler already has one of the upstream URLs.
2. During import, `DiscoveryServiceImpl` maps the exported handler id to a new target handler id.
3. `DiscoveryUpstreamServiceImpl#importData(...)` checks duplicates using the old exported handler id, so it does not see the existing target row.
4. The insert uses the remapped target handler id and can violate `UNIQUE(discovery_handler_id, upstream_url)`.

### Environment

```markdown
```

### Debug logs

_No response_

### Anything else?

- `shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/DiscoveryServiceImpl.java`
- `shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/DiscoveryUpstreamServiceImpl.java`
- `db/init/mysql/schema.sql`

hengyuss:
Fixes: #6465 

Make sure that:

### DIFF
--- a/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/DiscoveryUpstreamServiceImpl.java
+++ b/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/DiscoveryUpstreamServiceImpl.java
@@ -290,6 +290,7 @@ public class DiscoveryUpstreamServiceImpl implements DiscoveryUpstreamService {
     }
     
     @Override
+    @Transactional(rollbackFor = Exception.class)
     public ConfigImportResult importData(final String namespace, final List<DiscoveryUpstreamDTO> discoveryUpstreamList, final ConfigsImportContext context) {
         if (CollectionUtils.isEmpty(discoveryUpstreamList)) {
             return ConfigImportResult.success();
@@ -317,7 +318,7 @@ public class DiscoveryUpstreamServiceImpl implements DiscoveryUpstreamService {
             }
             discoveryUpstreamDTO.setNamespaceId(namespace);
             discoveryUpstreamDTO.setId(null);
-            discoveryUpstreamDTO.setDiscoveryHandlerId(discoveryHandlerIdMapping.get(discoveryUpstreamDTO.getDiscoveryHandlerId()));
+            discoveryUpstreamDTO.setDiscoveryHandlerId(discoveryHandlerIdMapping.getOrDefault(discoveryUpstreamDTO.getDiscoveryHandlerId(), discoveryUpstreamDTO.getDiscoveryHandlerId()));
             DiscoveryUpstreamDO discoveryUpstreamDO = DiscoveryUpstreamDO.buildDiscoveryUpstreamDO(discoveryUpstreamDTO);
             discoveryUpstreamMapper.insert(discoveryUpstreamDO);
             successCount++;
--- a/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/DiscoveryUpstreamServiceImpl.java
+++ b/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/DiscoveryUpstreamServiceImpl.java
@@ -305,7 +305,7 @@ public class DiscoveryUpstreamServiceImpl implements DiscoveryUpstreamService {
             String discoveryHandlerId = discoveryUpstreamDTO.getDiscoveryHandlerId();
             String url = discoveryUpstreamDTO.getUrl();
             Set<String> existsUpstreamUrlSet = discoveryHandlerUpstreamMap
-                    .getOrDefault(discoveryHandlerId, Lists.newArrayList())
+                    .getOrDefault(discoveryHandlerIdMapping.getOrDefault(discoveryHandlerId, discoveryHandlerId), Lists.newArrayList())
                     .stream()
                     .map(DiscoveryUpstreamDO::getUpstreamUrl)
                     .collect(Collectors.toSet());
--- a/shenyu-admin/src/test/java/org/apache/shenyu/admin/service/DiscoveryUpstreamServiceTest.java
+++ b/shenyu-admin/src/test/java/org/apache/shenyu/admin/service/DiscoveryUpstreamServiceTest.java
@@ -36,6 +36,7 @@ import org.apache.shenyu.admin.model.entity.ProxySelectorDO;
 import org.apache.shenyu.admin.model.entity.SelectorDO;
 import org.apache.shenyu.admin.model.result.ConfigImportResult;
 import org.apache.shenyu.admin.model.vo.DiscoveryUpstreamVO;
+import org.apache.shenyu.admin.service.configs.ConfigsImportContext;
 import org.apache.shenyu.admin.service.impl.DiscoveryUpstreamServiceImpl;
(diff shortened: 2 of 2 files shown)


Aias00 (commented):
Reviewing #6532 (fixes #6465) — one-line change to `DiscoveryUpstreamServiceImpl#importData(String, List, ConfigsImportContext)`.

**The fix is correct.** `discoveryHandlerUpstreamMap` is keyed by target/DB `discoveryHandlerId` (grouped from `selectByNamespaceId`), but the DTO carries the *source* handler id, and `discoveryHandlerIdMapping` (populated in `DiscoveryServiceImpl#importData`, `put(oldId, newId)`) is the old→new map. The old code looked up the source id in a DB-id-keyed map, so the duplicate check always returned an empty set and the insert then collided with `UNIQUE(discovery_handler_id, upstream_url)`. Using `discoveryHandlerIdMapping.getOrDefault(id, id)` to remap before the lookup is the right fix. Import order is also sound: `ConfigsExportImportEnum` runs Discovery (9) before DiscoveryUpstream (10), so the mapping is populated in time. No blockers.

Two things I'd like addressed before merge:

1. **No test for the fixed overload.** `DiscoveryUpstreamServiceTest#testImportData` only exercises `importData(List)`, not the `(namespace, list, context)` method this PR actually changes — and the PR checklist leaves the test box unchecked. Please add a unit test that stubs `selectByNamespaceId` with an existing upstream under the *remapped* handler id and asserts the DTO (carrying the *source* id + duplicate url) is skipped with `successCount=0`. That's the precise regression being fixed and it's currently unguarded.

2. **Namespace overload is still not `@Transactional`.** Issue #6465's "Expected Behavior" explicitly asks for transactional safety so partial imports aren't left behind on insert failure. The `importData(List)` overload has `@Transactional(rollbackFor = Exception.class)`; the namespace overload this PR touches does not. Adding the annotation would close the issue's secondary requirement and is consistent with the sibling overload.

Nits (non-blocking, pre-existing):
- `discoveryHandlerUpstreamMap` is built once before the loop and not updated after each insert, so two DTOs in the same batch with the same (remapped handler id, url) will: first inserts OK, second bypasses the stale in-memory check and hits the DB unique key. Same pattern exists in the List overload, so not a regression — but worth a follow-up to fold the import batch into the dedup set.
- Minor asymmetry: L308 uses `getOrDefault(id, id)` (falls back to source id) while L320 uses `.get(id)` (returns null on miss). If a source handler id were ever absent from the mapping, the dup-check would run against the source id while the insert would null the field. I couldn't find a reachable path that produces such an upstream (the export always includes the handler), but a hand-edited import JSON could, so making the two lookups consistent (or documenting the invariant) would be safer.

Aias00:
Both fixes look correct and well-scoped. The duplicate-check key fix is right — `discoveryHandlerUpstreamMap` is keyed by `DiscoveryUpstreamDO::getDiscoveryHandlerId` (L303, i.e. the stored/remapped id), so looking it up via `discoveryHandlerIdMapping.getOrDefault(id, id)` is the correct way to find existing upstreams for the same handler. The `getOrDefault(id, id)` on the remap (L320) is a real data-integrity fix — `.get(...)` returning null on an unmapped id previously wrote rows with a null `discoveryHandlerId`. Adding `@Transactional(rollbackFor = Exception.class)` is a reasonable defensive add.

One coverage gap: the fallback path for the second fix (an unmapped handler id kept as-is via `getOrDefault(id, id)`) isn't tested. The new test only uses a mapped id (`old_handler_id`→`new_handler_id`). A second case where `discoveryHandlerId` is *not* in `discoveryHandlerIdMapping` — asserting the inserted row's `discoveryHandlerId` equals the original — would pin that behavior and prevent a future regression to `.get(...)`.

Minor (out of scope for #6465): the 2-arg `importData` (~L250-290) has the same multi-insert loop without `@Transactional`, so a mid-loop failure leaves partial inserts. Not addressed here, but the asymmetry is worth a note.

Aias00 (approved):
## Summary

This PR fixes duplicate detection during discovery-upstream import when handler ids are remapped by `ConfigsImportContext.discoveryHandlerIdMapping`.

### What is correct
- The existing-url lookup now keys off the **remapped** handler id: `discoveryHandlerIdMapping.getOrDefault(discoveryHandlerId, discoveryHandlerId)` instead of the raw `discoveryHandlerId`. This makes the duplicate check consistent with the id that is actually persisted, so an incoming upstream whose URL already exists under the target (mapped) handler is correctly treated as a duplicate.
- The insert no longer produces a `null` handler id: `setDiscoveryHandlerId(... .getOrDefault(..., original))` replaces the previous `get(...)` which could return `null` for unmapped handlers.
- `@Transactional(rollbackFor = Exception.class)` on the `importData(namespace, list, context)` overload gives the multi-row import atomicity.
- Tests `testImportDataWithNamespaceAndContext` and `testImportDataWithNamespaceUnmappedHandlerId` cover both the mapped and unmapped paths and assert the duplicate is rejected (`successCount == 0`).

### Verification
- `ConfigImportResult` is asserted for both success and duplicate scenarios; the mock `selectByNamespaceId` returns rows keyed by the remapped id, exercising exactly the fixed path.

### Non-blocking suggestions
- The single-arg `importData(List)` overload is still non-transactional. Consider annotating it too for consistency, or have it delegate to the 3-arg overload so both share the same atomicity guarantee.

### Conclusion
Approved. The change is a clear correctness fix with covering tests and low risk.

```bash
gh pr review 6532 --repo apache/shenyu --approve --body-file review_6532.md
```

PR #6533, opened 2026-08-01
 fix: return ERROR_TOKEN when basic auth credentials are missing

### LINKED ISSUE #6474: [BUG] Basic Auth plugin throws NPE when credentials are missing
### Current Behavior

The Basic Auth plugin can throw `NullPointerException` when a protected request does not provide credentials.

`BasicAuthPlugin.doExecute()` reads the credential from the `Authorization` header or URI user info. If neither exists, `authorization` can be `null`, but it is still passed to the configured authentication strategy:

```java
authenticationStrategy.authenticate(basicAuthRuleHandle, authorization)
```

The default strategy then dereferences the value directly:

```java
return authentication.equals(((DefaultBasicAuthRuleHandle) basicAuthRuleHandle).getAuthorization());
```

### Expected Behavior

Missing Basic Auth credentials should fail authentication and return the normal `ERROR_TOKEN` response instead of throwing an internal exception.

### Steps to Reproduce

1. Enable the Basic Auth plugin on a route with the default strategy.
2. Send a request to that route without an `Authorization` header and without URI user info.
3. The default strategy dereferences `authentication` and can throw `NullPointerException`.

### Code Location

- `shenyu-plugin/shenyu-plugin-security/shenyu-plugin-basic-auth/src/main/java/org/apache/shenyu/plugin/basic/auth/BasicAuthPlugin.java`
- `shenyu-plugin/shenyu-plugin-security/shenyu-plugin-basic-auth/src/main/java/org/apache/shenyu/plugin/basic/auth/strategy/DefaultBasicAuthAuthenticationStrategy.java`

lll-peanut:
Fixes #6474

  ## What's changed
  - BasicAuthPlugin.doExecute now short-circuits with ERROR_TOKEN when the
    Authorization header and URI user info are both missing, instead of passing
    null to the authentication strategy.
  - DefaultBasicAuthAuthenticationStrategy.authenticate now uses Objects.equals
    so a null credential fails authentication instead of throwing NPE.

  ## Why
  A request without credentials produced `authorization == null`, and the
  default strategy dereferenced it (`authentication.equals(...)`), causing a
  NullPointerException (HTTP 500) instead of the expected 401 ERROR_TOKEN.

  ## Tests
  - BasicAuthPluginTest.testDoExecuteWithoutAuthorization: request without
    Authorization header / user info must return Illegal authorization and not
    invoke the chain (failed with NPE before the fix).
  - DefaultBasicAuthAuthenticationStrategyTest.testAuthenticateWithNullAuthentication:
    authenticate(null) must return false, not throw.
  - `mvn -pl shenyu-plugin/shenyu-plugin-security/shenyu-plugin-basic-auth test`
    → 15 tests passed.

  ## Manual verification
  - no credentials        → {"code":401,"message":"Illegal authorization"}
  - Authorization: test:test123 → request forwarded to upstream (200)
  - wrong credentials     → 401 Illegal authorization

Make sure that:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -130,10 +130,10 @@ jobs:
             ./mvnw.cmd -B clean test -Prelease
           elif mvnd --version > /dev/null 2>&1; then
             echo "Using mvnd for build"
-            mvnd -B clean test -Prelease
+            mvnd -B clean test -Prelease -DskipRemoteResources=true
           else
             echo "Falling back to maven wrapper"
-            ./mvnw -B clean test -Prelease
+            ./mvnw -B clean test -Prelease -DskipRemoteResources=true
           fi
       - uses: codecov/codecov-action@v1
         with:
--- a/pom.xml
+++ b/pom.xml
@@ -84,6 +84,7 @@
         <java.version>17</java.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <skipTests>false</skipTests>
+        <skipRemoteResources>false</skipRemoteResources>
         <!-- maven plugin version start -->
         <exec-maven-plugin.version>1.6.0</exec-maven-plugin.version>
         <jacoco-maven-plugin.version>0.8.12</jacoco-maven-plugin.version>
@@ -670,6 +671,13 @@
 
     <build>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-remote-resources-plugin</artifactId>
+                <configuration>
+                    <skip>${skipRemoteResources}</skip>
+                </configuration>
+            </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-source-plugin</artifactId>
--- a/shenyu-admin/src/main/java/org/apache/shenyu/admin/model/page/PageParameter.java
+++ b/shenyu-admin/src/main/java/org/apache/shenyu/admin/model/page/PageParameter.java
@@ -56,7 +56,7 @@ public class PageParameter implements Serializable {
      */
     public PageParameter(final Integer currentPage, final Integer pageSize) {
         this.currentPage = Objects.isNull(currentPage) || currentPage <= 0 ? 1 : currentPage;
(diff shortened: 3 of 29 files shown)


Aias00 (approved):
## Review: fix: return ERROR_TOKEN when basic auth credentials are missing

**Verdict: Approve (ship it).** No blocking issues — the fix is correct, minimal, fail-closed, and well-tested.

### Correctness
- Before: a request with no `Authorization` header and no URI userInfo produced `authorization == null` (`StringUtils.defaultString(null, null)`), which then hit `authentication.equals(...)` in `DefaultBasicAuthAuthenticationStrategy.authenticate` → `NullPointerException` (HTTP 500) instead of the expected 401.
- After: `BasicAuthPlugin.doExecute` short-circuits on `StringUtils.isBlank(authorization)` → `ShenyuResultEnum.ERROR_TOKEN` (code 401, "Illegal authorization", confirmed in `ShenyuResultEnum.java:43`), and `DefaultBasicAuthAuthenticationStrategy.authenticate` now uses `Objects.equals` so a `null` credential returns `false` rather than NPE. The two changes are complementary defense-in-depth (the plugin guard also protects custom SPI strategies from a null-credential NPE), not redundant.
- No fail-open / auth-bypass is introduced: `chain.execute` is unreachable without a successful `authenticate`. Confirmed `authenticate` is only invoked when `basicAuthRuleHandle` is non-null (`Objects.nonNull(authenticationStrategy)` derives from the handle via `Optional`), so the `null`-handle cast path is not reachable here.

### Tests
- `BasicAuthPluginTest.testDoExecuteWithoutAuthorization` is a genuine regression test: pre-fix it throws NPE synchronously from `doExecute` (failing `verifyComplete()`); post-fix it asserts the body contains "Illegal authorization" and `chain` is never invoked. Good.
- `DefaultBasicAuthAuthenticationStrategyTest.testAuthenticateWithNullAuthentication` directly covers `authenticate(..., null) → false`.

### Nits (optional)
1. `BasicAuthPluginTest.testDoExecuteWithoutAuthorization` reassigns the shared instance field `exchange` instead of a local variable. Harmless given `@BeforeEach`, but a local `ServerWebExchange noAuthExchange = ...` would be clearer.
2. Consider adding the empty-header case (`Authorization: ""`) — it exercises the `defaultString("", null) → "" → isBlank` branch of the new guard, distinct from the pure-null path.

CI: all build / e2e / it / it-k8s / CodeQL checks pass. No license-header or generated-artifact issues.

PR #6534, opened 2026-08-01
fix: handle null discoveryUpstreams in proxy selector update

### LINKED ISSUE #6517: [BUG] Proxy selector update throws NPE when discoveryUpstreams is omitted
### Search before asking

### Apache ShenYu Component

shenyu-admin

### What happened

`ProxySelectorAddDTO.discoveryUpstreams` is optional in the DTO; it has no `@NotNull` or `@NotEmpty` validation:

```java
private List<DiscoveryUpstream> discoveryUpstreams;
```

The create path handles this safely:

```java
if (!CollectionUtils.isEmpty(proxySelectorAddDTO.getDiscoveryUpstreams())) {
    proxySelectorAddDTO.getDiscoveryUpstreams().forEach(...);
    ...
}
```

But the update path deletes existing upstream rows first and then directly iterates the request list without a null check:

```java
int result = discoveryUpstreamMapper.deleteByDiscoveryHandlerId(discoveryHandlerId);
LOG.info("delete discovery upstreams, count is: {}", result);
proxySelectorAddDTO.getDiscoveryUpstreams().forEach(discoveryUpstream -> {
    DiscoveryUpstreamDO discoveryUpstreamDO = DiscoveryUpstreamDO.builder()
            ...
            .build();
});
```

If the client omits `discoveryUpstreams` or sends it as `null`, the update request throws `NullPointerException` after deleting the existing upstream rows inside the transaction. The transaction should roll back, but the API still returns a server error instead of treating the missing list consistently with the create path or rejecting it through validation.

### Expected behavior

Proxy selector update should either:

- validate `discoveryUpstreams` as required and return a clear validation error, or
- handle null/empty lists consistently with the create path.

It should not throw `NullPointerException` from `getDiscoveryUpstreams().forEach(...)`.

### How to reproduce

1. Create a proxy selector with a discovery handler.
2. Send a `PUT /proxy-selector/{id}` update request whose JSON body omits `discoveryUpstreams` or sets it to `null`.
3. The controller accepts the DTO validation because `discoveryUpstreams` is not annotated as required.
4. `ProxySelectorServiceImpl.update(...)` calls `proxySelectorAddDTO.getDiscoveryUpstreams().forEach(...)` and throws `NullPointerException`.

### Debug logs

_No response_

### Environment

Current `master` branch.

lll-peanut:
Fixes #6517

## What's changed
  - ProxySelectorServiceImpl.update no longer calls
    `discoveryUpstreams.forEach(...)` when the list is null/omitted.
  - The delete-and-reinsert of discovery upstreams (and its log statement) is
    now guarded by a non-empty check, consistent with the create path
    (`addUpstreamList`). A null/empty list means upstreams are left untouched.

  ## Why
  `discoveryUpstreams` is optional in ProxySelectorAddDTO, but the update path
  deleted existing upstream rows and then iterated the request list without a
  null check, throwing NullPointerException and returning a server error
  instead of a consistent result.

  ## Tests
  - Added ProxySelectorServiceTest.testUpdateWithNullDiscoveryUpstreams:
    update with null discoveryUpstreams must succeed and must not call
    deleteByDiscoveryHandlerId. It failed with
    `NullPointerException: ... getDiscoveryUpstreams() is null` before the fix.
  - `mvn -pl shenyu-admin test -Dtest=ProxySelectorServiceTest` → 6 tests passed.

  ## Manual verification
  - Create a proxy selector with one upstream, then PUT /proxy-selector/{id}
    with the JSON body omitting discoveryUpstreams.
  - Before fix: HTTP 500 + NPE in admin log.
  - After fix: HTTP 200 "update success", existing upstreams are preserved.

Make sure that:

### DIFF
--- a/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/ProxySelectorServiceImpl.java
+++ b/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/ProxySelectorServiceImpl.java
@@ -368,28 +368,30 @@ public class ProxySelectorServiceImpl implements ProxySelectorService {
         discoveryDO.setProps(discovery.getProps());
         discoveryMapper.updateSelective(discoveryDO);
         // update discovery upstream list
-        int result = discoveryUpstreamMapper.deleteByDiscoveryHandlerId(discoveryHandlerId);
-        LOG.info("delete discovery upstreams, count is: {}", result);
-        proxySelectorAddDTO.getDiscoveryUpstreams().forEach(discoveryUpstream -> {
-            DiscoveryUpstreamDO discoveryUpstreamDO = DiscoveryUpstreamDO.builder()
-                    .id(UUIDUtils.getInstance().generateShortUuid())
-                    .discoveryHandlerId(discoveryHandlerId)
-                    .namespaceId(discoveryDO.getNamespaceId())
-                    .protocol(discoveryUpstream.getProtocol())
-                    .url(discoveryUpstream.getUrl())
-                    .status(discoveryUpstream.getStatus())
-                    .weight(discoveryUpstream.getWeight())
-                    .props(discoveryUpstream.getProps())
-                    .dateCreated(Optional.ofNullable(discoveryUpstream.getStartupTime()).map(t -> new Timestamp(Long.parseLong(t))).orElse(currentTime))
-                    .dateUpdated(Optional.ofNullable(discoveryUpstream.getStartupTime()).map(t -> new Timestamp(Long.parseLong(t))).orElse(currentTime))
-                    .build();
-            discoveryUpstreamMapper.insert(discoveryUpstreamDO);
-        });
+        if (!CollectionUtils.isEmpty(proxySelectorAddDTO.getDiscoveryUpstreams())) {
+            int result = discoveryUpstreamMapper.deleteByDiscoveryHandlerId(discoveryHandlerId);
+            LOG.info("delete discovery upstreams, count is: {}", result);
+            proxySelectorAddDTO.getDiscoveryUpstreams().forEach(discoveryUpstream -> {
+                DiscoveryUpstreamDO discoveryUpstreamDO = DiscoveryUpstreamDO.builder()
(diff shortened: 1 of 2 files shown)


Aias00 (approved):
Reviewed #6534 — fix: handle null discoveryUpstreams in proxy selector update.

The fix is correct and well-scoped. `ProxySelectorServiceImpl.update` previously called `deleteByDiscoveryHandlerId` and then `getDiscoveryUpstreams().forEach(...)` unconditionally; a null/omitted `discoveryUpstreams` (an optional field per `ProxySelectorAddDTO`) wiped rows inside the transaction and then NPE'd. Moving the delete+insert block inside `if (!CollectionUtils.isEmpty(...))` mirrors the existing create path (`addUpstreamList`, lines 255-278) and resolves issue #6517. On the null/empty path the existing DB rows are now re-fetched and re-synced via `changeUpstream`, which is a safe no-op for `LocalDiscoveryProcessor`/`AbstractDiscoveryProcessor` (they just publish a sync event from the supplied list). Log statements were relocated inside the guard, which is an improvement (the previous "insert count" line was effectively dead on the null path).

No blockers. Two nits:

- The new `testUpdateWithNullDiscoveryUpstreams` only exercises the `null` case; an empty-list variant (`setDiscoveryUpstreams(Collections.emptyList())`) would cover the other side of `CollectionUtils.isEmpty`, though both collapse to the same path.
- The test asserts `deleteByDiscoveryHandlerId` is never called but does not assert `changeUpstream` is invoked with the preserved rows, so the "existing upstreams are re-synced" contract isn't pinned. A `verify(discoveryProcessor).changeUpstream(any(), any())` would lock that.

Out-of-scope note (not blocking this PR): `proxySelectorAddDTO.getDiscovery()` is still dereferenced without a null check a few lines above (365-368), so a client omitting both `discovery` and `discoveryUpstreams` would still NPE one line earlier. Worth a follow-up hardening issue.

CI is green (ci/e2e/it/it-k8s/CodeQL all SUCCESS); mergeState is BLOCKED only pending review.

PR #6535, opened 2026-08-01
fix: validate parent tag existence when creating child tag

### LINKED ISSUE #6520: [BUG] Creating a child tag with a missing parentTagId throws NPE
### Search before asking

### Apache ShenYu Component

shenyu-admin

### What happened

`TagServiceImpl.createInner(...)` loads the parent tag when `parentTagId` is not the root id:

```java
if (!tagDTO.getParentTagId().equals(AdminConstants.TAG_ROOT_PARENT_ID)) {
    TagDO tagDO = tagMapper.selectByPrimaryKey(tagDTO.getParentTagId());
    ext = buildExtParamByParentTag(tagDO);
} else {
    ext = GsonUtils.getInstance().toJson(Optional.ofNullable(tagExt).orElse(new TagDO.TagExt()));
}
```

`buildExtParamByParentTag(...)` then dereferences the parent tag immediately:

```java
private String buildExtParamByParentTag(final TagDO parentTagDO) {
    String ext = "";
    if (parentTagDO.getId().equals(AdminConstants.TAG_ROOT_PARENT_ID)) {
        ...
    }
```

If the request supplies a non-root `parentTagId` that does not exist, `tagMapper.selectByPrimaryKey(...)` returns `null`, and the admin API fails with `NullPointerException` instead of a validation error.

### Expected behavior

Creating or updating a child tag should validate that `parentTagId` exists before building inherited tag metadata. A missing parent tag should return a clear parameter/validation error rather than an internal 500.

### How to reproduce

1. Call the tag create API with a non-root `parentTagId` that is not present in the tag table.
2. `TagServiceImpl.createInner(...)` calls `tagMapper.selectByPrimaryKey(parentTagId)` and receives `null`.
3. `buildExtParamByParentTag(null)` calls `parentTagDO.getId()` and throws `NullPointerException`.

### Debug logs

_No response_

### Environment

Current `master` branch.

lll-peanut:
Fixes #6520

  ## What's changed
  - TagServiceImpl.createInner now validates the parent tag exists when
    parentTagId is not the root id. A missing parent tag throws
    ValidFailException("parent tag is not found") instead of passing null to
    buildExtParamByParentTag.

  ## Why
  Creating a child tag with a non-existent parentTagId made
  tagMapper.selectByPrimaryKey return null, then
  buildExtParamByParentTag(null) called parentTagDO.getId() and threw
  NullPointerException (internal 500) instead of a clear validation error.

  ## Tests
  - Added TagServiceTest.testCreateWithNonExistentParentTag: create with a
    missing parentTagId must throw ValidFailException. It failed with
    `NullPointerException: ... "parentTagDO" is null` before the fix.
  - `mvn -pl shenyu-admin test -Dtest=TagServiceTest` → 7 tests passed.

  ## Manual verification
  - POST /tag with a parentTagId that does not exist.
  - Before fix: NPE in admin log, generic error response.
  - After fix: response message "parent tag is not found", no NPE.

Make sure that:

### DIFF
--- a/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/TagServiceImpl.java
+++ b/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/TagServiceImpl.java
@@ -72,6 +72,7 @@ public class TagServiceImpl implements TagService {
         String ext = "";
         if (!tagDTO.getParentTagId().equals(AdminConstants.TAG_ROOT_PARENT_ID)) {
             TagDO tagDO = tagMapper.selectByPrimaryKey(tagDTO.getParentTagId());
+            Assert.notNull(tagDO, "parent tag is not found");
             ext = buildExtParamByParentTag(tagDO);
         } else {
             ext = GsonUtils.getInstance().toJson(Optional.ofNullable(tagExt).orElse(new TagDO.TagExt()));
--- a/shenyu-admin/src/test/java/org/apache/shenyu/admin/service/TagServiceTest.java
+++ b/shenyu-admin/src/test/java/org/apache/shenyu/admin/service/TagServiceTest.java
@@ -18,6 +18,7 @@
 package org.apache.shenyu.admin.service;
 
 import com.google.common.collect.Lists;
+import org.apache.shenyu.admin.exception.ValidFailException;
 import org.apache.shenyu.admin.mapper.TagMapper;
 import org.apache.shenyu.admin.model.dto.TagDTO;
 import org.apache.shenyu.admin.model.entity.TagDO;
@@ -38,6 +39,7 @@ import java.util.List;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.BDDMockito.given;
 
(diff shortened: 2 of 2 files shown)


Aias00 (approved):
Reviewed #6535 — fix for #6520. Tight and correct.

**Change scope (2 files, +9 lines):** `TagServiceImpl.createInner` now calls `Assert.notNull(tagDO, "parent tag is not found")` right after `tagMapper.selectByPrimaryKey(parentTagId)`, so a child-tag create with a non-existent `parentTagId` throws `ValidFailException` instead of NPE-ing inside `buildExtParamByParentTag(null)` (which dereferences `parentTagDO.getId()` at line 203). The root-parent path (`else` branch) is untouched, and the only other caller of `buildExtParamByParentTag` — `recurseUpdateTag` — feeds it `allData.get(id)` where `id` always comes from `allDataMap` keys, so it can't regress. `createRootTag`'s only production caller (`PullSwaggerDocServiceImpl`) passes `TAG_ROOT_PARENT_ID`, so it takes the root branch and is unaffected.

**Exception routing verified:** `ValidFailException` → `ShenyuAdminException` → `ShenyuException`, caught by `@ExceptionHandler(ShenyuException.class)` in `ExceptionHandlers`, returning `ShenyuAdminResult.error("parent tag is not found")`. Replaces the previous generic NPE response — strictly better for callers.

**Test:** `testCreateWithNonExistentParentTag` stubs `selectByPrimaryKey("456")` → null and asserts `ValidFailException`. On master (without the guard) this path throws NPE, so the test genuinely fails before the fix and passes after — a real regression test, not a tautological one.

**Two nits (non-blocking):**
1. The new test asserts only the exception type. Consider also asserting `getMessage().equals("parent tag is not found")` to lock the user-facing contract the PR body describes.
2. Wording: the sibling guard at line 88 is `"the updated tag is not found"`; the new one is `"parent tag is not found"` (missing leading article). Trivial parallelism nit only.

No blockers. CI green, mergeState BLOCKED only pending review.

PR #6821, opened 2026-08-03
fix: apply configured WAF statusCode to HTTP response instead of hardcoded 403

### LINKED ISSUE #6477: [BUG] WAF reject rules ignore the configured HTTP status code
### Current Behavior

WAF reject rules expose a configurable `statusCode`, but the HTTP transport status is always set to `403 Forbidden`.

`WafHandle` includes `statusCode` and defaults it to `403`, but `WafPlugin` hardcodes the response status:

```java
exchange.getResponse().setStatusCode(HttpStatus.FORBIDDEN);
Object error = ShenyuResultWrap.error(exchange, Integer.parseInt(wafHandle.getStatusCode()), Constants.REJECT_MSG, null);
```

So a rule configured with another status code only changes the body code, while the actual HTTP status remains 403.

### Expected Behavior

The configured WAF `statusCode` should either be applied to the HTTP response status as well, or the configuration should be removed/renamed so operators do not expect it to control the transport status.

### Impact

Operators cannot return custom reject statuses from WAF rules, and clients can observe inconsistent HTTP status and response body code.

### Code Location

- `shenyu-common/src/main/java/org/apache/shenyu/common/dto/convert/rule/WafHandle.java`
- `shenyu-plugin/shenyu-plugin-waf/src/main/java/org/apache/shenyu/plugin/waf/WafPlugin.java`

wy471x:
apply configured WAF statusCode to HTTP response instead of hardcoded 403

<!--
-->
Make sure that:

## Summary
### Problem: 
WafPlugin hardcoded exchange.getResponse().setStatusCode(HttpStatus.FORBIDDEN), ignoring the configurable statusCode from WafHandle. This meant setting a custom          
  statusCode (e.g., 404) only changed the response body code, while the actual HTTP transport status remained 403.
                                                                                                                                                                                     
### Changes:                                                                                                                                                                                                                                                                                                                                                       
  - WafPlugin.java:67-71 — Parse wafHandle.getStatusCode() once, apply it via setRawStatusCode(statusCode) instead of setStatusCode(HttpStatus.FORBIDDEN), and reuse the same int    
  value in ShenyuResultWrap.error(), keeping HTTP status and body code consistent.                                                                                                   
  - WafPluginTest.java — Updated testWafPluginReject to use a valid statusCode: "403" and added assertEquals(403, exchange.getResponse().getRawStatusCode()). Added                  
  testWafPluginRejectWithCustomStatusCode that configures statusCode: "404" and asserts the HTTP response status is actually 404.

close [#6477](https://github.com/apache/shenyu/issues/6477)


Aias00:
Good catch — hardcoding `HttpStatus.FORBIDDEN` ignored the configured `statusCode` on the wire, and `setRawStatusCode(int)` is the right way to honor custom codes. One behavioral change worth a guard before merge:

**Null/non-numeric `statusCode` now produces a 500 instead of a 403 (WAF no longer fails closed).** `WafHandle.statusCode` is a plain String with no default initializer (`WafHandle.java:37`), so a reject rule whose JSON omits `statusCode` deserializes to `null`. In the old code, `setStatusCode(FORBIDDEN)` ran *before* `Integer.parseInt(wafHandle.getStatusCode())` for the body — so even when the parse threw, 403 was already on the response. In the new code (`WafPlugin.java:67`), `int statusCode = Integer.parseInt(wafHandle.getStatusCode())` parses *first*, so a null/non-numeric value throws before any status is set and propagates to the global error handler (likely 500). For a WAF, a misconfigured reject rule should fail closed (403), not surface as a 500. Suggested fix:

```java
int statusCode = Optional.ofNullable(wafHandle.getStatusCode())
        .filter(NumberUtils::isCreatable)
        .map(Integer::parseInt)
        .orElse(HttpStatus.FORBIDDEN.value());
```

Also: the two new tests cover 403 and 404 (valid codes) but not the null/empty/non-numeric `statusCode` case — exactly the path that now 500s. A test for that case would pin the intended fail-closed behavior.

Minor: `setRawStatusCode` accepts any int, so a user configuring `statusCode:"200"` makes a WAF reject look like a 200 success to the client. Restoring configurability is the PR's intent, so this is the operator's responsibility, but a 4xx/5xx range guard would prevent foot-shooting if desired.

Aias00 (approved):
Thanks @wy471x!

This correctly fixes #6477. The old code hard-coded `setStatusCode(HttpStatus.FORBIDDEN)` while putting the configured value only into the response body, so a custom `statusCode` (e.g. 404) never reached the transport layer. Parsing the value once and applying it via `setRawStatusCode` while reusing the same int in `ShenyuResultWrap.error` keeps the HTTP status and the body code consistent.

The test updates are appropriate: `testWafPluginReject` now uses a valid 403 and asserts on `getRawStatusCode()`, and the new `testWafPluginRejectWithCustomStatusCode` proves the configured 404 actually becomes the HTTP status — this test fails on the old code.

Non-blocking suggestion for a follow-up: `Integer.parseInt(wafHandle.getStatusCode())` is unguarded (pre-existing risk, since the old body path parsed it too), and configs that historically stored non-numeric or out-of-range values like "0" would now hit the transport status path as well. Wrapping the parse in a try/catch with a 403 fallback would make the REJECT path fully defensive:

```java
int statusCode;
try {
    statusCode = Integer.parseInt(wafHandle.getStatusCode());
} catch (NumberFormatException e) {
    statusCode = HttpStatus.FORBIDDEN.value();
}
```

Not worth blocking this fix on. LGTM, approving.

PR #6892, opened 2026-08-04
fix: prevent ArrayIndexOutOfBoundsException in K8s ingress protocol annotation parsing

### LINKED ISSUE #6485: [BUG] Kubernetes upstream protocol annotation can break Divide and Dubbo ingress parsing
### Current Behavior

Kubernetes ingress parsing can throw when the `shenyu.apache.org/upstreams-protocol` annotation has fewer entries than the number of endpoint addresses.

In `DivideIngressParser`, the parser indexes the protocol array once per endpoint address without checking the length:

```java
upstream.setProtocol(Objects.isNull(protocol) ? "http://" : protocol[i++]);
```

In `DubboIngressParser`, the parser also requires the annotation to exist and then increments `i` twice in one expression:

```java
String[] protocols = annotations.get(IngressConstants.UPSTREAMS_PROTOCOL_ANNOTATION_KEY).split(",");
...
.protocol(Objects.isNull(protocols[i++]) ? "dubbo://" : protocols[i++])
```

This can skip protocol entries and can throw `ArrayIndexOutOfBoundsException` even when the annotation length matches the endpoint count. If the annotation is missing in the Dubbo parser, `annotations.get(...).split(",")` can also throw.

### Expected Behavior

The parsers should tolerate a missing or shorter protocol annotation by using the plugin default protocol for endpoints without an explicit protocol, matching the safer behavior used by the websocket parser.

### Steps to Reproduce

1. Create a Kubernetes ingress that references a service with multiple endpoint addresses.
2. Set `shenyu.apache.org/upstreams-protocol` to a single value, or omit it for Dubbo.
3. Reconcile the ingress.
4. Divide/Dubbo parsing can throw while building upstreams.

### Code Location

- `shenyu-kubernetes-controller/src/main/java/org/apache/shenyu/k8s/parser/DivideIngressParser.java`
- `shenyu-kubernetes-controller/src/main/java/org/apache/shenyu/k8s/parser/DubboIngressParser.java`
- `shenyu-kubernetes-controller/src/main/java/org/apache/shenyu/k8s/parser/WebSocketParser.java`
  - websocket parsing already guards `i < protocols.length`, which is the safer behavior.

wy471x:
- DivideIngressParser: add bounds check when indexing protocol array, fall back to "http://" when annotation has fewer entries than endpoint addresses
- DubboIngressParser: fix NPE when annotation is missing, fix double-increment bug where protocols[i++] was evaluated twice in one expression
- Add unit tests covering missing, exact, fewer, mixed, and empty annotation scenarios

<!--
-->
Make sure that:

## Summary

### Bug Fixes                                                                                                                                                                          
                                                                                                                                                                                     
  DivideIngressParser.java — The protocol array was indexed with protocol[i++] without any bounds check. When the shenyu.apache.org/upstreams-protocol annotation had fewer entries  
  than endpoint addresses, it threw ArrayIndexOutOfBoundsException. Fixed by switching to an indexed for-loop with a bounds guard (i >= protocol.length) that falls back to          
  "http://".                                                                                                                                                                         
                                                            
  DubboIngressParser.java — Two bugs fixed:                                                                                                                                          
  1. annotations.get(...).split(",") would NPE when the annotation was missing. Added null/containsKey guard, defaulting to "dubbo://".                                              
  2. protocols[i++] was evaluated twice in the same ternary expression (Objects.isNull(protocols[i++]) ? "dubbo://" : protocols[i++]), double-incrementing i and causing both skipped
   entries and AIOOBE. Replaced with a single bounds-checked protocols[i].                                                                                                           
                                                                                                                                                                                     
### Unit Tests                                                
                                                                                                                                                                                     
  - DivideIngressParserTest.java — 5 tests: annotation missing, exact match, fewer entries than addresses, mixed protocols, empty annotation.                                        
  - DubboIngressParserTest.java — 5 tests covering the same scenarios plus the previously-NPE case.

close [#6485](https://github.com/apache/shenyu/issues/6485)


Aias00:
Clean fix — the double-increment in `Objects.isNull(protocols[i++]) ? "dubbo://" : protocols[i++]` was a classic bug (the condition's `i++` always advances, and the true branch advances again, skipping elements), correctly resolved by computing `upstreamProtocol` once before the ternary. The AIOOBE bounds check and the annotation null-guard are both correct.

I did a variant search across the k8s parsers: `WebSocketParser.java:170-192` already has the correct guard pattern (null-check + bounds-check + single evaluation), so this PR brings `DivideIngressParser` and `DubboIngressParser` in line with that reference implementation. No other parser carries the `[i++]` bug — `SofaParser`/`GrpcParser` don't use the annotation-based protocol array, so they're unaffected.

One minor pre-existing edge case (consistent across all parsers incl. `WebSocketParser`, so not a regression): when the annotation value is an empty string, `"".split(",")` returns `[""]`, so the first upstream gets `protocol = ""` rather than the `"http://"`/`"dubbo://"` fallback — the bounds check only covers index-out-of-range, not empty entries. `testEmptyProtocolAnnotation` only asserts `assertNotNull`, sidestepping this. A `StringUtils.isBlank(protocols[i])` check would make empty entries fall back to the default too. Fine to leave, just flagging for completeness.

Aias00 (approved):
Thanks @wy471x!

Both fixes are correct and close real defects:

- **DivideIngressParser**: `protocol[i++]` with no bounds check threw `ArrayIndexOutOfBoundsException` whenever `shenyu.apache.org/upstreams-protocol` had fewer entries than endpoint addresses. The indexed loop with the `i >= protocol.length` fallback to `http://` is the right behavior, and incrementing `i` per address (rather than only when a protocol was assigned) keeps the annotation entries aligned with their addresses.
- **DubboIngressParser**: the old `Objects.isNull(protocols[i++]) ? "dubbo://" : protocols[i++]` both double-incremented `i` (skipping entries and eventually going out of bounds) and NPE'd when the annotation was missing. The single bounds-checked `protocols[i]` plus the `annotations` null/`containsKey` guard fix all three problems at once, with `dubbo://` as the sensible default.

Tests cover missing/exact/fewer/mixed/empty annotation scenarios for both parsers and assert on the actual upstream protocol values, which is exactly the right level of verification.

⚠️ One coordination note (non-blocking): this PR and #6896 both add `shenyu-kubernetes-controller/src/test/java/org/apache/shenyu/k8s/parser/DubboIngressParserTest.java` with different test sets, so whichever merges second will need a rebase that merges the two test classes together (this PR's file also doesn't include the parseIngressRule null-safety tests from #6896). Worth reconciling before either lands.

LGTM, approving.

PR #6894, opened 2026-08-05
fix: set MQTT sessionPresent from cleanSession flag and stored session state (#6889)

### LINKED ISSUE #6889: [BUG] MCP tool-call timeout does not cancel the in-flight plugin chain
## Description
The CONNACK session-present flag is hard-coded `true` for every connection, including the very first connect and clients with the clean-session flag set. Combined with the complete absence of session-state persistence (`Disconnect.cleanChannel` only touches `ChannelRepository`; `Disconnect.java:39` carries `// todo Clean session`), the broker actively lies about stored session state.

## Location
- `shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/Connect.java:72`

## Impact
Violates MQTT-3.2.2-6/7. A client reconnecting with `cleanSession=1` is told a session was present (must be `0`); a first-time client is told the same.

## Suggested fix
Set `sessionPresent` from the connect-flags `cleanSession` bit and stored session state (false when `cleanSession=1` or no prior session).

## Related existing
None.

lll-peanut:
fix: https://github.com/apache/shenyu/issues/6889

  - Connect no longer hard-codes sessionPresent(true); it is derived from the cleanSession flag and stored per-client session state
  - add SessionRepository/MqttSession and resume stored subscriptions on reconnect
  - add unit tests covering first connect, reconnect, cleanSession discard, and subscription resume

DISCONNECT is now dispatched by MqttFactory, so clean-session sessions are removed on graceful disconnect; the disconnected channel is also removed from SubscribeRepository while persistent session state is kept for resume.

Make sure that:


Aias00:
Good fix — the hardcoded `sessionPresent(true)` was a clear spec violation (MQTT-3.2.2-6/7), and the new derivation `!cleanSession && session != null` is correct, including the ordering (sessionPresent computed before the cleanSession overwrite). Two things worth addressing:

**Subscription resume loses QoS.** `Connect.java` (~L91) calls `SubscribeRepository.add(new ArrayList<>(session.getTopics()), Collections.singletonList(ctx.channel()))`, which goes through the `add(List<String>, List<Channel>)` overload (`SubscribeRepository.java:43`) — a topic→channel mapping only. It bypasses the QoS-aware `add(Channel, List<MqttTopicSubscription>)` overload (L56). `MqttSession.topics` is a `Set<String>`, so the original QoS levels (and duplicate subscriptions with differing QoS) are collapsed; resumed subscriptions get an implicit default QoS rather than the client's original. Could `MqttSession` store `List<MqttTopicSubscription>` (or `Map<String, MqttQoS>`) and resume via the QoS-aware overload?

**Clean-session state leaks on graceful DISCONNECT (dead-code cleanup).** As the PR body notes, `Disconnect` isn't dispatched by `MqttFactory`, so the cleanup added in `Disconnect.java:42-52` is unreachable today. Consequence: a clean-session client that disconnects gracefully leaves its `MqttSession` in `SessionRepository` until the same clientId next connects clean (`Connect.java` discards it lazily); clients that never reconnect leak indefinitely. Acceptable as a follow-up, but please either wire `Disconnect` dispatch in this PR or open a tracking issue and reference it in the PR/code so the leak isn't lost.

Minor (pre-existing, now more reachable): `SubscribeRepository.add(List,List)` (L43-49) has a check-then-act race — `get(s)` returns `getOrDefault(topic, new CopyOnWriteArrayList<>())`, so for an absent topic two concurrent resumes can each create a fresh list and one `put` overwrites the other's channel. Not introduced here, just noting the resume path makes it reachable.

lll-peanut:
> Good fix — the hardcoded `sessionPresent(true)` was a clear spec violation (MQTT-3.2.2-6/7), and the new derivation `!cleanSession && session != null` is correct, including the ordering (sessionPresent computed before the cleanSession overwrite). Two things worth addressing:
> 
> **Subscription resume loses QoS.** `Connect.java` (~L91) calls `SubscribeRepository.add(new ArrayList<>(session.getTopics()), Collections.singletonList(ctx.channel()))`, which goes through the `add(List<String>, List<Channel>)` overload (`SubscribeRepository.java:43`) — a topic→channel mapping only. It bypasses the QoS-aware `add(Channel, List<MqttTopicSubscription>)` overload (L56). `MqttSession.topics` is a `Set<String>`, so the original QoS levels (and duplicate subscriptions with differing QoS) are collapsed; resumed subscriptions get an implicit default QoS rather than the client's original. Could `MqttSession` store `List<MqttTopicSubscription>` (or `Map<String, MqttQoS>`) and resume via the QoS-aware overload?
> 
> **Clean-session state leaks on graceful DISCONNECT (dead-code cleanup).** As the PR body notes, `Disconnect` isn't dispatched by `MqttFactory`, so the cleanup added in `Disconnect.java:42-52` is unreachable today. Consequence: a clean-session client that disconnects gracefully leaves its `MqttSession` in `SessionRepository` until the same clientId next connects clean (`Connect.java` discards it lazily); clients that never reconnect leak indefinitely. Acceptable as a follow-up, but please either wire `Disconnect` dispatch in this PR or open a tracking issue and reference it in the PR/code so the leak isn't lost.
> 
> Minor (pre-existing, now more reachable): `SubscribeRepository.add(List,List)` (L43-49) has a check-then-act race — `get(s)` returns `getOrDefault(topic, new CopyOnWriteArrayList<>())`, so for an absent topic two concurrent resumes can each create a fresh list and one `put` overwrites the other's channel. Not introduced here, just noting the resume path makes it reachable.

Thanks for the review. I addressed the three points:

  1. MqttSession now stores subscriptions as topic -> MqttQoS, and Connect resumes stored subscriptions through the QoS-aware SubscribeRepository.add(Channel,
  List<MqttTopicSubscription>) path. I also added tests for QoS retention and QoS replacement on resubscribe.

  2. DISCONNECT is now dispatched by MqttFactory. Clean-session DISCONNECT removes the stored session, and DISCONNECT also unregisters the current channel from SubscribeRepository
  while keeping persistent session state for later resume.

  3. SubscribeRepository.add now uses computeIfAbsent instead of the previous get/default-list/put check-then-act sequence, including the QoS-aware overload.

  I also added a guard return after rejecting unsupported protocol versions to avoid continuing through authentication/session setup after sending the CONNACK rejection.

Aias00 (dismissed):
Thanks @lll-peanut!

This properly fixes #6889. The session-present handling now follows MQTT-3.2.2-6/7: `sessionPresent` is only true for a `cleanSession=0` connect when the server holds a stored session for the clientId, instead of the old hard-coded `sessionPresent(true)`. The surrounding session semantics look right too:

- `cleanSession=true` discards any previously stored session; `cleanSession=false` creates a persistent one when absent.
- Resuming stored subscriptions (topic + QoS) onto the new channel via `SubscribeRepository` matches the spec's expectation that a persistent session's subscriptions survive reconnects.
- Wiring `DISCONNECT` through `MqttFactory` was the missing dispatch — clean-session state is discarded on graceful disconnect while persistent sessions are kept, and the channel is removed from `SubscribeRepository` in both cases.
- `Subscribe`/`Unsubscribe` now keep the session's topic→QoS map in sync, including QoS replacement on re-subscribe.
- The `computeIfAbsent` change in `SubscribeRepository.add` also fixes a real race: the old get-then-put could NPE or lose channels under concurrent subscribes.
- Bonus fix: the missing `return` after refusing an unsupported protocol version previously let CONNECT processing continue — good catch, and the new test pins that behavior down.

Tests cover first connect, reconnect, clean-session discard, subscription resume, QoS retention/replacement, and DISCONNECT dispatch, using Awaitility appropriately for the async repository operations. The single failing check (`k8s-examples-http`) is unrelated to `shenyu-protocol-mqtt`.

One note for follow-up (non-blocking): a client that connects with `cleanSession=true` and then drops the connection without DISCONNECT leaves its clean session in `SessionRepository` until the next connect replaces it — harmless given connect-time replacement, but worth tracking if session state grows.

LGTM, approving.

Aias00:

Aias00 (approved):
## Review: set MQTT `sessionPresent` from `cleanSession` flag and stored session

**Verdict: APPROVED** (with two non-blocking coordination notes)

### What this PR does
Implements correct MQTT session lifecycle semantics, fixing the previously hard-coded/incorrect `sessionPresent` behaviour:

- **`Connect.java`** — `sessionPresent` is now derived as `!cleanSession && storedSession != null` (MQTT-3.2.2-6/7). On a persistent-session reconnect it re-registers the stored subscriptions onto the new channel.
- **`Disconnect.java`** — clean-session disconnect now discards the stored session and unregisters the channel; persistent-session disconnect keeps the session and only unregisters the channel.
- **`MqttFactory.java`** — adds the **previously missing `DISCONNECT` dispatch case** (`messageType.disconnect(ctx)`), so `DISCONNECT` actually closes the channel and runs cleanup instead of being a silent no-op. (`MqttFactoryTest.disconnectShouldCloseChannel` now asserts the channel is inactive — previously it asserted the opposite.)
- **New `MqttSession` / `SessionRepository`** — persistent session state (clean-session flag + per-topic QoS subscriptions).
- **`Subscribe` / `Unsubscribe`** — track topics in the `MqttSession`.
- **`SubscribeRepository`** — `removeChannel` + `computeIfAbsent` fix so unsubscribe/cleanup no longer loses entries.

### Tests
`ConnectSessionPresentTest`, `DisconnectDispatchTest`, `SubscribeQosTest` are thorough and spec-driven (first-connect vs reconnect vs clean-session discard, QoS retention / re-subscribe replacement, channel re-registration). Existing `PublishTest` / `MqttFactoryTest` updated for the new singletons. Good coverage.

### Notes (non-blocking)
1. **Merge coordination with the active MQTT PR cluster.** This touches `MqttFactory`, `Connect`, `Disconnect`, `Subscribe`, `SubscribeRepository` — the same files edited by #6906 (wildcard matching), #6902 (LWT), #6923 (SUBACK QoS), #6913 (delivery QoS), #6977 (ByteBuf release), #6939 (shutdown NPE). Please rebase/sequence these so they don't conflict; ideally land the session-lifecycle foundation (#6894) before the feature PRs build on top of it.
2. **CI:** the only failing job is `it-k8s/build (shenyu-integrated-test-k8s-ingress-http)`, which is an unrelated k8s ingress integration-test infra job. The Java `pr_build`, `build`, and license-header checks are green.

Thanks for the clean, well-tested fix — this brings ShenYu's MQTT handling back in line with the spec.

lll-peanut:
> Heads-up (PMC Aias00): your PR's CI failures are in shared infrastructure checks (`k8s-examples-http`, `e2e-case-logging-rocketmq`, `build-docker-images`) that are currently failing across many unrelated PRs. This looks like a CI-environment issue, not a problem with your change. Please re-run the failed checks; the code change itself appears fine. No code changes needed for these specific failures.

Thanks for the approval and the coordination notes! I agree that #6894 should preferably be merged before the overlapping MQTT PRs. I’ll keep an eye on potential conflicts and rebase if needed. The remaining failed k8s check appears unrelated, so I’ll rerun it if necessary.

PR #6895, opened 2026-08-05
fix: ServiceProcessor imports Spring @Service instead of Dubbo @Service — old-style Dubbo beans never get beanPath

### LINKED ISSUE #6682: [BUG] ServiceProcessor imports Spring `@Service` instead of Dubbo `@Service` — old-style Dubbo beans never get beanPath
- Severity: High
- Location:
`shenyu-client/shenyu-client-dubbo/shenyu-client-apache-dubbo/src/main/java/org/apache/shenyu/client/apache/dubbo/processor/extractor/ServiceProcessor.java:25` (import), `:40-42` (process); `shenyu-client/shenyu-client-dubbo/shenyu-client-apache-dubbo/src/main/java/org/apache/shenyu/client/apache/dubbo/ApacheDubboApiBeansExtractor.java:21,33` (extractor uses Dubbo's `@Service`)
- 
Description:
`ApacheDubboApiBeansExtractor` correctly imports `org.apache.dubbo.config.annotation.Service` (Dubbo's deprecated `@Service`) and registers it as a supported annotation. But `ServiceProcessor` imports `org.springframework.stereotype.Service` (Spring's generic `@Service`) and types itself `ApiAnnotationProcessor<Service>` with `matchAnnotation()` returning Spring's `Service.class`. A bean annotated with Dubbo's `@Service` (or `@ShenyuService` which meta-annotates it) does NOT carry Spring's `@Service`, so `ApiAnnotationProcessor.getAnnotation(apiBean)` always returns null, and `process(apiBean, annotation)` is never invoked. The beanPath is never set (stays `""` default).
- 
Impact:
In the ApiBean-based extraction path, old-style `@Service`/`@ShenyuService` Dubbo beans produce metadata with missing path components (only contextPath + methodName, no service path). The gateway routes to wrong paths.
- 
Suggested fix:
Change `ServiceProcessor` import from `org.springframework.stereotype.Service` to `org.apache.dubbo.config.annotation.Service`. Consider using `annotation.path()` instead of `annotation.value()` to match `DubboServiceProcessor`.
- 
Confidence: High
- Related existing: none — #5815 (annotation default values), #5943 (method config) are different; this is a wrong-import bug.

---

hengyuss:
Fixes: #6682 

Make sure that:


Aias00:
Thanks for this fix — the root cause (matching Spring's `@Service` instead of Dubbo's) is correctly addressed, and the new `getRpcExt` mirrors `DubboServiceProcessor`. A few observations from reading the diff and surrounding code.

**`rpcExt` builder diverges from `ApacheDubboServiceBeanListener.buildRpcExt`** (`ApacheDubboServiceBeanListener.java:225-252`). The listener's builder sets `.serialization(serviceBean.getSerialization())` (L236) and populates per-method `DubboRpcMethodExt` (L239-250); the new `getRpcExt` in `ServiceProcessor` (and the existing `DubboServiceProcessor:86-99`) sets neither. For old-style `@Service` users this means the registered `rpcExt` omits `serialization` and per-method loadbalance/retries/timeout/sent. If downstream (`AbstractDubboMetaDataHandler` / `ApacheDubboConfigCache`) selects the serializer from `rpcExt.serialization`, these users silently fall back to the default. Could you confirm whether the omission is intentional, or extract a shared `DubboRpcExtBuilders.from(ServiceBean)` helper so all three builders stay aligned?

**No test covers the fix path.** The PR body's "submit test cases" checkbox is unchecked and no test files changed. The highest-value missing test: a `ServiceProcessorTest` proving (a) a Dubbo `@Service`-annotated bean whose instance is a `ServiceBean` gets `beanPath = annotation.path()` and a populated `rpcExt`, and (b) a non-`ServiceBean` instance gets the `"{}"` fallback. That guards the exact regression this PR fixes.

Minor: `serviceBean.getProtocol().getName()` (new `getRpcExt`) can NPE if `getProtocol()` is null — same pre-existing pattern as `DubboServiceProcessor:88` / `ApacheDubboServiceBeanListener:227`, so not a regression, just noting the line is reachable with an unconfigured `ServiceBean`.

Aias00 (dismissed):
Thanks @hengyuss!

This correctly fixes #6682. `ServiceProcessor` was matching Dubbo's old-style `@Service` annotation but importing Spring's `org.springframework.stereotype.Service`, so the annotation callback never fired for old-style Dubbo beans and `beanPath` was never set. Switching to `org.apache.dubbo.config.annotation.Service` fixes the wiring, and reading `annotation.path()` (rather than `value()`) is right since `path()` is the canonical attribute on Dubbo's `@Service`/`@DubboService`.

The `DubboRpcExtBuilders` consolidation is a good cleanup on top:

- One shared builder now produces identical `rpcExt` for `@DubboService` and old-style `@Service` beans; `DubboServiceProcessor` previously emitted an `rpcExt` without `serialization` and without method-level config, so this also brings the processor path to parity with the `ApacheDubboServiceBeanListener` path.
- The added `Objects.nonNull(serviceBean.getProtocol())` guard fixes a latent NPE the old inline code had.
- Adding `process(ApiBean.ApiDefinition)` to `ServiceProcessor` means method-level registrations also carry `rpcExt`, matching `DubboServiceProcessor`.
- The `provided`-scope `dubbo` dependency in `shenyu-client-dubbo-common` is safe: the class is only reachable from the apache-dubbo client module, so alibaba-dubbo consumers never load it.

Full CI matrix is green (including the apache-dubbo integrated tests and checkstyle/license checks), and the new unit tests cover the ServiceBean / non-ServiceBean / null-protocol / method-config branches. LGTM, approving.

dengliming (approved):

PR #6308, opened 2026-03-29
[type:refactor] remove sofa support

Aias00:
## Summary
- remove sofa modules, examples, integrated tests, and spring boot starters
- clean remaining sofa references from upgrade scripts, release docs, and the tracked admin static bundle
- keep only the upgrade SQL delete statements that purge historical sofa data and remove the obsolete SOFA-specific negative test

## Verification
- `mvn -pl shenyu-common -Dtest=RpcTypeEnumTest test`
- `git diff --check`
- `rg -n --hidden -S "\bsofa\b|Sofa|SOFA" /Users/aias/Work/github/shenyu/.worktrees/remove-sofa --glob '!**/.git/**' --glob '!**/.git' --glob '!**/target/**' --glob '!**/node_modules/**' --glob '!RELEASE-NOTES.md'

PR #6896, opened 2026-08-05
fix: add null-safety checks for labels and services in K8s RPC ingress parsers

### LINKED ISSUE #6486: [BUG] Kubernetes RPC ingress parsers can throw NPE when labels or services are missing
### Current Behavior

The Kubernetes RPC ingress parsers assume ingress labels and label-referenced services are always present.

For Dubbo, gRPC, and Sofa route parsing, the code iterates `labels.keySet()` and then immediately dereferences the service returned by `serviceLister.namespace(namespace).get(labels.get(label))`:

```java
for (String label : labels.keySet()) {
    Map<String, String> metadataAnnotations = serviceLister.namespace(namespace)
            .get(labels.get(label))
            .getMetadata()
            .getAnnotations();
    ...
}
```

If an ingress has rules but no labels, `labels` can be `null`. If a label points to a service name that is not currently present in the local cache, `serviceLister...get(...)` can return `null`. Both cases can throw `NullPointerException` during reconciliation.

### Expected Behavior

The parsers should handle missing labels or missing referenced services gracefully, for example by skipping that metadata/rule entry and allowing a later reconcile after the service cache is populated.

### Steps to Reproduce

1. Create a Dubbo, gRPC, or Sofa ingress with `spec.rules` but without metadata labels expected by the parser, or with a label value that references a service not present in the controller cache.
2. Let the Shenyu Kubernetes controller reconcile it.
3. The parser can throw while iterating labels or reading service metadata annotations.

### Code Location

- `shenyu-kubernetes-controller/src/main/java/org/apache/shenyu/k8s/parser/DubboIngressParser.java`
- `shenyu-kubernetes-controller/src/main/java/org/apache/shenyu/k8s/parser/GrpcParser.java`
- `shenyu-kubernetes-controller/src/main/java/org/apache/shenyu/k8s/parser/SofaParser.java`

wy471x:

<!--
-->
Make sure that:

## Summary
### Bug Fix (3 files)                                                                                                                                                                  
   
  DubboIngressParser.java, GrpcParser.java, SofaParser.java — in each parseIngressRule method, added null-safety guards before iterating over labels:                                
                                                            
  1. Null/empty labels check — if labels is null or empty, skip the label iteration with continue instead of throwing NullPointerException.                                          
  2. Missing service check — before calling .getMetadata().getAnnotations(), check if serviceLister.namespace(namespace).get(...) returned null (service not yet in cache). Log a
  warning and continue, allowing a later reconcile to pick it up once the service cache is populated.                                                                                
                                                            
### Unit Tests (3 new files, 9 tests)                                                                                                                                                  
                                                            
  - DubboIngressParserTest.java — 3 tests                                                                                                                                            
  - GrpcParserTest.java — 3 tests                           
  - SofaParserTest.java — 3 tests   

close [#6486](https://github.com/apache/shenyu/issues/6486)

### DIFF
--- a/shenyu-kubernetes-controller/src/main/java/org/apache/shenyu/k8s/parser/DubboIngressParser.java
+++ b/shenyu-kubernetes-controller/src/main/java/org/apache/shenyu/k8s/parser/DubboIngressParser.java
@@ -238,8 +238,16 @@ public class DubboIngressParser implements K8sResourceParser<V1Ingress> {
                     SelectorData selectorData = createSelectorData(pathPath, conditionList, upstreamList);
                     List<RuleData> ruleDataList = new ArrayList<>();
                     List<MetaData> metaDataList = new ArrayList<>();
+                    if (Objects.isNull(labels) || labels.isEmpty()) {
+                        continue;
+                    }
                     for (String label : labels.keySet()) {
-                        Map<String, String> metadataAnnotations = serviceLister.namespace(namespace).get(labels.get(label)).getMetadata().getAnnotations();
+                        V1Service service = serviceLister.namespace(namespace).get(labels.get(label));
+                        if (Objects.isNull(service) || Objects.isNull(service.getMetadata())) {
+                            LOG.warn("Service {} not found in namespace {}", labels.get(label), namespace);
+                            continue;
+                        }
+                        Map<String, String> metadataAnnotations = service.getMetadata().getAnnotations();
                         DubboRuleHandle ruleHandle = createDubboRuleHandle(annotations);
                         List<ConditionData> ruleConditionList = getRuleConditionList(metadataAnnotations);
                         RuleData ruleData = createRuleData(metadataAnnotations, ruleHandle, ruleConditionList);
--- a/shenyu-kubernetes-controller/src/main/java/org/apache/shenyu/k8s/parser/GrpcParser.java
+++ b/shenyu-kubernetes-controller/src/main/java/org/apache/shenyu/k8s/parser/GrpcParser.java
@@ -252,8 +252,16 @@ public class GrpcParser implements K8sResourceParser<V1Ingress> {
                     SelectorData selectorData = createSelectorData(pathPath, conditionList, grpcUpstreamList);
                     List<RuleData> ruleDataList = new ArrayList<>();
                     List<MetaData> metaDataList = new ArrayList<>();
+                    if (Objects.isNull(labels) || labels.isEmpty()) {
+                        continue;
+                    }
                     for (String label : labels.keySet()) {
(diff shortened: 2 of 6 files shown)


Aias00 (approved):
Thanks @wy471x!

The null-safety guards are correct and well-placed:

- `ingress.getMetadata().getLabels()` is passed straight into `parseIngressRule` and can be `null` for an ingress without labels, so the old `labels.keySet()` was a guaranteed NPE; the null/empty guard fixes that. Skipping the whole path entry (rather than registering an `IngressConfiguration` with empty rules/metadata) is the right outcome — a selector with no rules and no service metadata would be meaningless.
- Guarding the `serviceLister` lookup is equally important: the service may legitimately not be in the informer cache yet during startup/reconcile races, and warn-and-continue lets a later reconcile pick it up instead of crashing the parse loop.
- The extra `service.getMetadata()` null check closes the remaining dereference chain.

The three test files cover the null-labels, missing-service, and happy-path cases for each parser, and the mocked `Lister`/`Indexer` setup matches how the production code queries the cache. LGTM, approving.

dengliming:
@wy471x Thanks for the fix. This PR currently has merge conflicts and needs to be rebased/merged with the latest target branch before it can be reviewed and merged. Could you please resolve the conflicts and update the PR?

Aias00 (changes requested):

Aias00 (approved):
Approving — consistent null-safety fix across all three parsers.

The chained expression

```java
serviceLister.namespace(namespace).get(labels.get(label)).getMetadata().getAnnotations();
```

has three unguarded links. Any of them being null — an Ingress with no selector `labels`, a label naming a Service the lister hasn't cached (or that was just deleted), or a Service whose `getMetadata()` is null — throws an NPE inside the reconcile loop. Since this runs on every Ingress event, a single unresolvable reference can wedge the controller's parsing path.

The fix is applied identically in `DubboIngressParser`, `GrpcParser` and `SofaParser`:

```java
V1Service service = serviceLister.namespace(namespace).get(labels.get(label));
if (Objects.isNull(service) || Objects.isNull(service.getMetadata())) {
    LOG.warn("Service {} not found in namespace {}", labels.get(label), namespace);
    continue;
}
```

Three things make this the right shape:
- The `labels == null || labels.isEmpty()` guard is hoisted above the loop, so an Ingress without selector labels is skipped cleanly rather than entering the iteration at all.
- `continue` (not a bare `if`) means a missing Service skips just that label and the remaining labels still get parsed — a partial failure doesn't discard the whole rule.
- The `LOG.warn` names both the Service and the namespace, which is exactly what an operator needs to diagnose a stale selector.

Consistency across the three parsers matters here because they duplicate this logic; fixing only one would leave the same bug latent in the others.

One note for coordination: `SofaParser` is touched here while #6308 proposes removing sofa support entirely. Whichever lands second will need a rebase — not a problem with this change, just flagging the overlap.

PR #6897, opened 2026-08-06
fix: enforce request body size limit during buffering in AiProxyPlugin

### LINKED ISSUE #6837: [BUG] AiProxyPlugin validates body size AFTER unbounded DataBufferUtils.join — OOM before 413
## Description
`AiProxyPlugin.doExecute` does `DataBufferUtils.join(exchange.getRequest().getBody())` (no `maxSize` overload) which aggregates the entire inbound body into a single in-memory `DataBuffer` **before** the size check at line 106 runs. The 5 MB guard fires only after the full payload is already resident in heap.

## Location
- `shenyu-plugin/shenyu-plugin-ai/shenyu-plugin-ai-proxy/src/main/java/org/apache/shenyu/plugin/ai/proxy/enhanced/AiProxyPlugin.java:103-110`

## Impact
A malicious/buggy client streaming a multi-GB body triggers full buffering → `OutOfMemoryError` before the 413 is ever produced. Trivially-achievable gateway DoS on any selector with the AI-proxy plugin enabled. (Mitigating factors: an upstream reverse proxy or Netty `HttpObjectAggregator` maxContentLength, if configured, could prevent this — but the plugin code itself has no defense.)

## Suggested fix
Use the `DataBufferUtils.join` variant with a `maxSize` and emit 413 on `DataBufferLimitException`, or stream-count bytes and short-circuit before buffering.

## Related existing
Distinct from AI-TOKENLIMITER-BODYWRITER-OOM (#6758, different plugin, response side).

wy471x:
Use DataBufferUtils.join with maxSize to prevent OOM from large request bodies. Previously the full body was buffered into memory before the size check ran, allowing a multi-GB payload to exhaust heap. Now DataBufferLimitException is raised during buffering and returned as 413.

Make sure that:

## Summary
- Replaced DataBufferUtils.join(exchange.getRequest().getBody()) (no max size) with DataBufferUtils.join(exchange.getRequest().getBody(), MAX_REQUEST_BODY_SIZE_BYTES). This makes   
  Spring's DataBufferUtils enforce the 5 MB limit during buffering — a DataBufferLimitException is thrown immediately when the limit is exceeded, rather than buffering the entire   
  multi-GB body into heap first. Added .onErrorResume(DataBufferLimitException.class, ...) to catch that exception and return HTTP 413. Removed the now-redundant post-buffer size   
  check.
- Added testRequestBodyExceedsMaxSize which mocks DataBufferUtils.join to return Mono.error(new DataBufferLimitException(...)) and asserts the response status is 413                
  PAYLOAD_TOO_LARGE.

close [#6837](https://github.com/apache/shenyu/issues/6837)

### DIFF
--- a/shenyu-plugin/shenyu-plugin-ai/shenyu-plugin-ai-proxy/src/main/java/org/apache/shenyu/plugin/ai/proxy/enhanced/AiProxyPlugin.java
+++ b/shenyu-plugin/shenyu-plugin-ai/shenyu-plugin-ai-proxy/src/main/java/org/apache/shenyu/plugin/ai/proxy/enhanced/AiProxyPlugin.java
@@ -41,6 +41,7 @@ import org.springframework.ai.chat.client.ChatClient;
 import org.springframework.ai.chat.model.ChatModel;
 import org.springframework.ai.chat.model.ChatResponse;
 import org.springframework.core.io.buffer.DataBuffer;
+import org.springframework.core.io.buffer.DataBufferLimitException;
 import org.springframework.core.io.buffer.DataBufferUtils;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpStatus;
@@ -63,7 +64,7 @@ public class AiProxyPlugin extends AbstractShenyuPlugin {
     /**
      * Maximum request body size: 5MB.
      */
-    private static final long MAX_REQUEST_BODY_SIZE_BYTES = 5 * 1024 * 1024L;
+    private static final int MAX_REQUEST_BODY_SIZE_BYTES = 5 * 1024 * 1024;
 
     private final AiModelFactoryRegistry aiModelFactoryRegistry;
 
@@ -100,18 +101,8 @@ public class AiProxyPlugin extends AbstractShenyuPlugin {
                         CacheKeyUtils.INST.getKey(
                                 selector.getId(), Constants.DEFAULT_RULE));
 
-        return DataBufferUtils.join(exchange.getRequest().getBody())
+        return DataBufferUtils.join(exchange.getRequest().getBody(), MAX_REQUEST_BODY_SIZE_BYTES)
                 .flatMap(dataBuffer -> {
-                    // Validate actual body size after reading, not just Content-Length header
-                    final int actualSize = dataBuffer.readableByteCount();
-                    if (actualSize > MAX_REQUEST_BODY_SIZE_BYTES) {
-                        DataBufferUtils.release(dataBuffer);
-                        LOG.warn("[AiProxy] Request body size {} exceeds maximum allowed size {}", 
-                                actualSize, MAX_REQUEST_BODY_SIZE_BYTES);
-                        exchange.getResponse().setStatusCode(HttpStatus.PAYLOAD_TOO_LARGE);
-                        return exchange.getResponse().setComplete();
-                    }
-                    
                     final String requestBody = dataBuffer.toString(StandardCharsets.UTF_8);
                     DataBufferUtils.release(dataBuffer);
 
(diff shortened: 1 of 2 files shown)


shenyu-plugin/shenyu-plugin-ai/shenyu-plugin-ai-proxy/src/main/java/org/apache/shenyu/plugin/ai/proxy/enhanced/AiProxyPlugin.java
```diff
@@ -100,18 +101,8 @@ protected Mono<Void> doExecute(
                         CacheKeyUtils.INST.getKey(
                                 selector.getId(), Constants.DEFAULT_RULE));
 
-        return DataBufferUtils.join(exchange.getRequest().getBody())
+        return DataBufferUtils.join(exchange.getRequest().getBody(), MAX_REQUEST_BODY_SIZE_BYTES)
```
nikhiln64, line 104: ## Context
Nice catch, this is a real fix. Buffering the whole body with the single argument DataBufferUtils.join before checking readableByteCount meant the memory was already spent by the time the limit was enforced, and passing the maxSize into join so it aborts mid stream and onErrorResume maps it to 413 is the right shape.

▎ Example: The test mocks DataBufferUtils.join statically and matches the size argument with anyInt(), so it only proves the DataBufferLimitException to 413 mapping and never pins the limit the plugin is responsible for. A regression that weakened the limit, say passing Integer.MAX_VALUE or a mis scaled constant, would still pass green.

## Suggestion

An ArgumentCaptor or an eq(MAX_REQUEST_BODY_SIZE_BYTES) matcher so the test asserts join is called with exactly the 5MB constant would close that gap. If it is easy, one unmocked test that feeds a real body over the limit through doExecute and asserts 413 with no downstream call would make the guarantee end to end.

## NOTE
Not blocking, the production change looks correct as is.

Aias00 (approved):
## Review: #6897 — fix: enforce request body size limit during buffering in AiProxyPlugin

**Verdict: ✅ Approve**

Correct and important OOM-prevention fix for #6837. The before/after difference is exactly the right one.

### What's correct
- **The limit is now enforced *during* buffering, not after.** Old code called `DataBufferUtils.join(exchange.getRequest().getBody())` with **no** max size — the entire (possibly multi-GB) body was materialized into heap, *then* a post-read size check ran. That check was pointless against an OOM: the damage was already done. The new `DataBufferUtils.join(body, MAX_REQUEST_BODY_SIZE_BYTES)` makes Spring abort with `DataBufferLimitException` the moment the accumulated bytes cross 5 MB, and Spring releases the partial buffers — so heap stays bounded.
- **The `long` → `int` change is required, not cosmetic.** `DataBufferUtils.join(Publisher, int maxByteCount)` takes an `int` maxSize, so the constant had to become `int` to be passed without a cast. 5 MB (5_242_880) is nowhere near `Integer.MAX_VALUE`, so the narrower type is harmless. I confirmed the constant is only used in this one call now (the post-check that also referenced it was removed), so there's no dangling `long` usage.
- **Narrow, correct error handling.** `.onErrorResume(DataBufferLimitException.class, ...)` catches *only* that exception type and returns 413 `PAYLOAD_TOO_LARGE`; any other error still propagates. The response status is set before `setComplete()`, and the test asserts exactly that.
- **Actual-byte semantics preserved.** The removed comment noted "validate actual body size, not just Content-Length" — the `join` maxSize counts real bytes read, so that intent is retained (a lying `Content-Length` header no longer bypasses the check).
- **Test is meaningful.** `testRequestBodyExceedsMaxSize` mocks `DataBufferUtils.join(any(), anyInt())` to emit `Mono.error(DataBufferLimitException)`, runs `doExecute`, and asserts the response is 413 via `StepVerifier`. It pins the new error path.

### Non-blocking suggestions
1. **Confirm `mockito-inline` is on this module's test classpath.** `MockedStatic<DataBufferUtils>` requires the inline mock maker. You noted local tests passed, so it's almost certainly present — just flagging so a CI environment without it doesn't surprise you.
2. **Buffer release on the happy path is already correct** (`DataBufferUtils.release(dataBuffer)` after `toString()`), and on the limit path Spring releases the partial buffers, so no leak. Nothing to change; noting it for the record.
3. **Minor:** if you ever want a limit larger than ~2 GB you'd need a different mechanism (the `int` overload caps at `Integer.MAX_VALUE`), but 5 MB is far from that — not a concern today.

### Verdict
Approving. This properly closes the OOM window (limit checked mid-stream instead of after full buffering) with a correctly-scoped error handler and a test that exercises it.

PR #6898, opened 2026-08-08
fix: offload websocket reconnect from shared timer to dedicated executor

### LINKED ISSUE #6844: [BUG] ShenyuWebsocketClient.healthCheck reconnectBlocking blocks the shared single-thread wheel timer
## Description
`ShenyuWebsocketClient.healthCheck()` calls `this.reconnectBlocking()` — a synchronous TCP-connect + WS-handshake call — when `!isOpen()`. `healthCheck` runs on `WheelTimerFactory.getSharedTimer()`, whose `taskExecutor` is a single-thread `ThreadPoolExecutor(1,1)`. While one client's `reconnectBlocking()` blocks, ALL other timer tasks stall — including other websocket clients' health checks and the `WebsocketSyncDataService.masterCheck` round task. No backoff, no jitter, no async reconnect.

## Location
- `shenyu-sync-data-center/shenyu-sync-data-websocket/src/main/java/org/apache/shenyu/plugin/sync/data/websocket/client/ShenyuWebsocketClient.java:238-251` (healthCheck → reconnectBlocking), `:110,148` (runs on shared wheel timer); `HierarchicalWheelTimer` taskExecutor = `ThreadPoolExecutor(1,1)`

## Impact
With multiple websocket URLs (HA admin), if one admin is unreachable, `reconnectBlocking()` blocks the shared single-thread executor for the TCP timeout duration (potentially minutes). All other clients' health checks and the masterCheck task are delayed, causing missed heartbeats, delayed failover, and lost config updates across all sync connections.

## Suggested fix
Run `reconnectBlocking` on a dedicated executor (not the shared wheel timer), or use the async `reconnect()` variant with a dedicated connection-accepted callback; add exponential backoff + jitter.

## Related existing
Distinct from #6569 (long-polling single-thread scheduler) — this is the websocket client reconnect path.

wy471x:
## Summary
- `ShenyuWebsocketClient.healthCheck()` previously called `reconnectBlocking()` directly on the shared wheel timer's single-thread executor. When an admin was unreachable, the TCP connect timeout blocked all other timer tasks (other clients' health checks, `masterCheck`), causing missed heartbeats and delayed failover.
- Offloaded reconnect to a dedicated `CachedThreadPool` executor so the timer thread is never blocked.
- Added exponential backoff with jitter (1s–60s) and an `AtomicBoolean` guard to prevent concurrent reconnect attempts for the same client.

## Test plan
- [x] `calculateBackoff()` returns 0 for no failures, exponential growth for failures 1-10, capped at 60s, with jitter
- [x] `healthCheck()` resets backoff on healthy connection, does not double-submit reconnects
- [x] `doReconnect()` increments backoff on failure (capped at 10), resets `reconnecting` in finally, applies backoff sleep, preserves interrupt status
- [x] All 15 tests pass (11 new + 4 existing)

close [#6844](https://github.com/apache/shenyu/issues/6844)

### DIFF
--- a/shenyu-sync-data-center/shenyu-sync-data-websocket/src/main/java/org/apache/shenyu/plugin/sync/data/websocket/client/ShenyuWebsocketClient.java
+++ b/shenyu-sync-data-center/shenyu-sync-data-websocket/src/main/java/org/apache/shenyu/plugin/sync/data/websocket/client/ShenyuWebsocketClient.java
@@ -43,11 +43,18 @@ import org.java_websocket.handshake.ServerHandshake;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.shenyu.common.concurrent.ShenyuThreadFactory;
+
 import java.net.URI;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * The type shenyu websocket client.
@@ -58,22 +65,35 @@ public final class ShenyuWebsocketClient extends WebSocketClient {
      * logger.
      */
     private static final Logger LOG = LoggerFactory.getLogger(ShenyuWebsocketClient.class);
-    
+
+    private static final ExecutorService RECONNECT_EXECUTOR = Executors.newCachedThreadPool(
+            ShenyuThreadFactory.create("websocket-reconnect", true));
+
+    private static final long MIN_RECONNECT_BACKOFF_MS = TimeUnit.SECONDS.toMillis(1);
+
+    private static final long MAX_RECONNECT_BACKOFF_MS = TimeUnit.SECONDS.toMillis(60);
+
     private volatile boolean alreadySync = Boolean.FALSE;
-    
+
     private final WebsocketDataHandler websocketDataHandler;
-    
+
     private final Timer timer;
-    
+
     private TimerTask timerTask;
-    
+
     private String runningMode;
-    
+
     private String masterUrl;
-    
+
     private volatile boolean isConnectedToMaster;
-    
+
     private final String namespaceId;
+
+    private final AtomicBoolean reconnecting = new AtomicBoolean(false);
+
+    private volatile long lastReconnectAttemptTime;
+
+    private final AtomicInteger reconnectBackoff = new AtomicInteger(0);
 
     /**
      * Instantiates a new shenyu websocket client.
@@ -238,16 +258,49 @@ public final class ShenyuWebsocketClient extends WebSocketClient {
     private void healthCheck() {
         try {
             if (!this.isOpen()) {
-                this.reconnectBlocking();
+                if (this.reconnecting.compareAndSet(false, true)) {
+                    RECONNECT_EXECUTOR.submit(this::doReconnect);
+                }
             } else {
+                this.reconnectBackoff.set(0);
                 this.sendPing();
                 send(getInstanceInfo());
-//                send(DataEventTypeEnum.RUNNING_MODE.name());
                 LOG.debug("websocket send to [{}] ping message successful", this.getURI());
             }
         } catch (Exception e) {
             LOG.error("websocket connect is error :{}", e.getMessage());
         }
+    }
+
+    private void doReconnect() {
+        try {
+            long backoff = calculateBackoff();
+            long since = System.currentTimeMillis() - lastReconnectAttemptTime;
+            long waitMs = backoff - since;
+            if (waitMs > 0) {
+                Thread.sleep(waitMs);
+            }
+            lastReconnectAttemptTime = System.currentTimeMillis();
+            this.reconnectBlocking();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+        } catch (Exception e) {
+            reconnectBackoff.set(Math.min(reconnectBackoff.get() + 1, 10));
+            LOG.error("websocket reconnect server[{}] error", this.getURI(), e);
+        } finally {
+            this.reconnecting.set(false);
+        }
+    }
+
+    private long calculateBackoff() {
+        int failures = reconnectBackoff.get();
+        if (failures <= 0) {
+            return 0;
+        }
+        long base = Math.min(
+                MIN_RECONNECT_BACKOFF_MS * (1L << Math.min(failures - 1, 10)),
+                MAX_RECONNECT_BACKOFF_MS);
+        return base + (long) (base * 0.5 * ThreadLocalRandom.current().nextDouble());
     }
 
     private String getInstanceInfo() {
--- a/shenyu-sync-data-center/shenyu-sync-data-websocket/src/main/java/org/apache/shenyu/plugin/sync/data/websocket/client/ShenyuWebsocketClient.java
+++ b/shenyu-sync-data-center/shenyu-sync-data-websocket/src/main/java/org/apache/shenyu/plugin/sync/data/websocket/client/ShenyuWebsocketClient.java
@@ -43,16 +43,18 @@ import org.java_websocket.handshake.ServerHandshake;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.shenyu.common.concurrent.MemorySafeTaskQueue;
 import org.apache.shenyu.common.concurrent.ShenyuThreadFactory;
+import org.apache.shenyu.common.concurrent.ShenyuThreadPoolExecutor;
 
 import java.net.URI;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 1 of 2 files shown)


Aias00 (dismissed):
## Review: #6898 — fix: offload websocket reconnect from shared timer to dedicated executor

**Verdict: ✅ Approve (with a few non-blocking suggestions)**

The core fix is correct and important, and the backoff/guard logic is well-tested.

### What's correct
- **Real root cause fixed.** `healthCheck()` runs on the shared wheel-timer's single-thread executor. The old code called `reconnectBlocking()` (a *blocking* TCP connect) directly on that thread, so an unreachable admin would stall every other timer task (other clients' health checks, `masterCheck`) → missed heartbeats and delayed failover. Offloading to a dedicated `CachedThreadPool` removes the blockage. This is exactly the right fix for #6844.
- **`AtomicBoolean reconnecting` guard prevents double-submit.** `compareAndSet(false, true)` in `healthCheck`, reset in `doReconnect`'s `finally`. While a reconnect is sleeping/in-flight, subsequent `healthCheck` calls are skipped. Correct.
- **Backoff math is sound.** `calculateBackoff()`: 0 failures → 0; else `MIN * (1 << min(failures-1, 10))` capped at 60s, plus up to +50% jitter. I verified the test bounds: failures=1 → [1000,1500]ms, =2 → [2000,3000], =4 → [8000,12000], =10 → capped 60000 + jitter ≤ 90000. All consistent with the assertions.
- **`InterruptedException` is respected** — `Thread.currentThread().interrupt()` restores status (tested), and the generic `catch (Exception)` increments backoff (capped at 10) and logs. `finally` always clears `reconnecting`.
- **Backoff reset on healthy path.** `healthCheck` sets `reconnectBackoff` to 0 when the socket is open, so a recovered connection doesn't carry stale backoff.
- **Good test coverage** — 11 new tests covering backoff zero/growth/cap/jitter, no-double-submit, reset-on-open, backoff increment/cap/reset-on-failure, backoff sleep, and interrupt preservation.

### Suggestions (non-blocking)
1. **`RECONNECT_EXECUTOR` is a static, unbounded `newCachedThreadPool`.** For a gateway with several admin endpoints, a sustained reconnect storm could grow the thread count (reclaimed only after 60s idle). Consider either a *bounded* pool or reusing ShenYu's existing managed executor (`ShenyuThreadFactory` + a fixed/limited pool) so reconnect threads are accounted for like the rest of the system.
2. **Backoff is a floor, not strictly enforced during long connects.** `lastReconnectAttemptTime` is stamped *after* `reconnectBlocking()` returns. If a connect itself takes longer than the computed `backoff` (e.g. multi-second TCP timeout), the next attempt's `waitMs` goes negative and reconnects immediately. Acceptable as a floor, just flagging the semantics.
3. **`testDoReconnect*` rely on the real `reconnectBlocking()` throwing** against `ws://localhost:9090` (no server). This is fine in CI but is technically environment-sensitive (and the connect attempt adds a few seconds of latency to those tests). If you ever see flakiness, stub `reconnectBlocking()` instead of calling through to the parent.

### Verdict
Approving. The fix addresses a genuine liveness bug (timer-thread blockage), and the retry/backoff/guard implementation matches its tests. Address the executor-bounding point as a follow-up if you want tighter resource control.

wy471x:
> ## Review: #6898 — fix: offload websocket reconnect from shared timer to dedicated executor
> **Verdict: ✅ Approve (with a few non-blocking suggestions)**
> 
> The core fix is correct and important, and the backoff/guard logic is well-tested.
> 
> ### What's correct
> * **Real root cause fixed.** `healthCheck()` runs on the shared wheel-timer's single-thread executor. The old code called `reconnectBlocking()` (a _blocking_ TCP connect) directly on that thread, so an unreachable admin would stall every other timer task (other clients' health checks, `masterCheck`) → missed heartbeats and delayed failover. Offloading to a dedicated `CachedThreadPool` removes the blockage. This is exactly the right fix for [[BUG] ShenyuWebsocketClient.healthCheck reconnectBlocking blocks the shared single-thread wheel timer #6844](https://github.com/apache/shenyu/issues/6844).
> * **`AtomicBoolean reconnecting` guard prevents double-submit.** `compareAndSet(false, true)` in `healthCheck`, reset in `doReconnect`'s `finally`. While a reconnect is sleeping/in-flight, subsequent `healthCheck` calls are skipped. Correct.
> * **Backoff math is sound.** `calculateBackoff()`: 0 failures → 0; else `MIN * (1 << min(failures-1, 10))` capped at 60s, plus up to +50% jitter. I verified the test bounds: failures=1 → [1000,1500]ms, =2 → [2000,3000], =4 → [8000,12000], =10 → capped 60000 + jitter ≤ 90000. All consistent with the assertions.
> * **`InterruptedException` is respected** — `Thread.currentThread().interrupt()` restores status (tested), and the generic `catch (Exception)` increments backoff (capped at 10) and logs. `finally` always clears `reconnecting`.
> * **Backoff reset on healthy path.** `healthCheck` sets `reconnectBackoff` to 0 when the socket is open, so a recovered connection doesn't carry stale backoff.
> * **Good test coverage** — 11 new tests covering backoff zero/growth/cap/jitter, no-double-submit, reset-on-open, backoff increment/cap/reset-on-failure, backoff sleep, and interrupt preservation.
> 
> ### Suggestions (non-blocking)
> 1. **`RECONNECT_EXECUTOR` is a static, unbounded `newCachedThreadPool`.** For a gateway with several admin endpoints, a sustained reconnect storm could grow the thread count (reclaimed only after 60s idle). Consider either a _bounded_ pool or reusing ShenYu's existing managed executor (`ShenyuThreadFactory` + a fixed/limited pool) so reconnect threads are accounted for like the rest of the system.
> 2. **Backoff is a floor, not strictly enforced during long connects.** `lastReconnectAttemptTime` is stamped _after_ `reconnectBlocking()` returns. If a connect itself takes longer than the computed `backoff` (e.g. multi-second TCP timeout), the next attempt's `waitMs` goes negative and reconnects immediately. Acceptable as a floor, just flagging the semantics.
> 3. **`testDoReconnect*` rely on the real `reconnectBlocking()` throwing** against `ws://localhost:9090` (no server). This is fine in CI but is technically environment-sensitive (and the connect attempt adds a few seconds of latency to those tests). If you ever see flakiness, stub `reconnectBlocking()` instead of calling through to the parent.
> 
> ### Verdict
> Approving. The fix addresses a genuine liveness bug (timer-thread blockage), and the retry/backoff/guard implementation matches its tests. Address the executor-bounding point as a follow-up if you want tighter resource control.

Thank you for the code review on this PR.

Fix:
1. Bounded executor — Replaced the unbounded newCachedThreadPool with a repo-standard ShenyuThreadPoolExecutor (core 1, max 8, 60s keep-alive) using MemorySafeTaskQueue +         
  ShenyuThreadFactory + AbortPolicy, so reconnect threads are accounted for like the rest of the system.                                                                             
  2. Backoff semantics — lastReconnectAttemptTime is now stamped in a finally after reconnectBlocking() completes, so a slow TCP connect no longer consumes the backoff window and   
  the next attempt strictly waits the full backoff.                                                                                                                                  
  3. Test isolation — testDoReconnect* now stub reconnectBlocking() with doThrow(...) instead of making real socket connections to ws://localhost:9090.

Aias00 (approved):
## Review: #6898 — fix: offload websocket reconnect from shared timer to dedicated executor

**Verdict: ✅ Approve**

Re-reviewing after the new commits. The branch now contains a 2026-08-15 commit `fix: bound reconnect executor and enforce backoff in websocket client` that **directly addresses the non-blocking suggestions from my prior review**. The change is materially better than what I approved before, so this is a re-approval of an improved diff (the prior approval was reset to REVIEW_REQUIRED by the new commits anyway).

### How my prior feedback was addressed
1. **Unbounded pool → bounded repo-standard executor (the main one).** Replaced `Executors.newCachedThreadPool(...)` with:
   ```java
   new ShenyuThreadPoolExecutor(
       RECONNECT_EXECUTOR_CORE_POOL_SIZE,   // 1
       RECONNECT_EXECUTOR_MAX_POOL_SIZE,    // 8
       RECONNECT_EXECUTOR_KEEP_ALIVE_MS,    // 60s
       TimeUnit.MILLISECONDS,
       new MemorySafeTaskQueue<>(Constants.THE_256_MB),
       ShenyuThreadFactory.create("websocket-reconnect", true),
       new ThreadPoolExecutor.AbortPolicy());
   ```
   Verified `ShenyuThreadPoolExecutor`, `MemorySafeTaskQueue`, `Constants.THE_256_MB`, and `ShenyuThreadFactory` all live in `shenyu-common` (which this module depends on) and the constructor signature matches the call — **compiles cleanly**. This removes the unbounded-thread-growth risk I flagged.
2. **Backoff window no longer consumed by a slow connect.** `lastReconnectAttemptTime` is now stamped in the `finally` block **after** `reconnectBlocking()` returns (success or failure), instead of before the attempt. Previously a slow TCP connect could eat time out of the intended backoff window.
3. **Tests no longer attempt real sockets.** `reconnectBlocking()` is stubbed (`doThrow(...)`), so the reconnect tests are deterministic and fast.

### Correctness of the new logic
- **Backoff math checks out.** `calculateBackoff()` returns
  `MIN(MIN_BACKOFF * (1 << min(failures-1, 10)), MAX_BACKOFF=60s) + jitter[0, base*0.5)`. I traced the test assertions and they line up exactly:
  | failures | expected range | assertion |
  |---|---|---|
  | 0 | 0 | `== 0` ✓ |
  | 1 | [1000, 1500) ms | `[1000,1500]` ✓ |
  | 2 | [2000, 3000) ms | `[2000,3000]` ✓ |
  | 4 | [8000, 12000) ms | `[8000,12000]` ✓ |
  | 10 | [60000, 90000) ms | `[60000,90000]` ✓ |
  The `reconnectBackoff` counter is capped at `10` (`Math.min(..., 10)`), matching the `min(failures-1, 10)` cap in the shift — consistent.
- **No negative-sleep crash.** `doReconnect` computes `waitMs = backoff - since` and only calls `Thread.sleep(waitMs)` when `waitMs > 0`, so even if `since > backoff` (which can happen across ticks) it never hits `Thread.sleep` with a negative arg.
- **No double-reconnect / no stuck flag.** `healthCheck` guards submission with `reconnecting.compareAndSet(false, true)`; `doReconnect`'s outer `finally` always resets `reconnecting` to `false`. The `InterruptedException` path re-asserts the interrupt flag (`Thread.currentThread().interrupt()`) and still resets the flag — clean.
- **Backoff reset on healthy connection.** When `isOpen()` is true, `healthCheck` resets `reconnectBackoff` to 0, so a recovered link doesn't carry a stale penalty.

### Tests
The new suite is solid: zero-backoff, exponential growth, max cap, jitter variation, "don't double-submit while reconnecting", backoff reset on open, increment-on-failure, cap-at-10, actual backoff sleep duration, and interrupt-status preservation. That's comprehensive coverage of the tricky scheduling/lifecycle cases.

### Minor non-blocking note (no change needed to merge)
The executor is a single **static** instance shared by all `ShenyuWebsocketClient` instances (core 1, scaling to max 8). This is the same sharing model as the original cached pool, just now bounded — strictly better, and fine for the gateway's small number of websocket sync clients. Not a concern.

### Verdict
Approving again. The revision removes the unbounded-pool risk, fixes the backoff-timing bug, and adds strong tests. Ready to merge once CI is green.

Issue #6883, opened 2026-08-04
[BUG] RewritePlugin placeholder branch drops first character of path-variable value + StringIndexOutOfBounds when regex lacks {


Aias00:
## Description
`responseContentLength` is a boxed `Integer` (default null). For a byte-type response with no `Content-Length` header (chunked/streamed), it stays null. When desensitization is enabled, `desensitizeShenyuRequestLog` calls `.toString()` on the null boxed value → NPE. Same unguarded `.toString()` pattern on `status` (`:192`) and `upstreamResponseTime` (`:195`).

## Location
- `shenyu-plugin/shenyu-plugin-logging/shenyu-plugin-logging-common/src/main/java/org/apache/shenyu/plugin/logging/common/collector/AbstractLogCollector.java:189-190,192,194-195`

## Impact
NPE thrown inside `doFinally` on the request thread for desensitization-enabled + byte-type chunked responses; that log lost.

## Suggested fix
Null-guard before `.toString()` (e.g. `String.valueOf(logInfo.getResponseContentLength())`).

## Related existing
None.

zhang-arvin:
I will work on this.

PR #6899, opened 2026-08-08
fix: NPE in AbstractLogCollector.desensitizeShenyuRequestLog on null boxed Integer/Long

### LINKED ISSUE #6884: [BUG] AbstractLogCollector.desensitizeShenyuRequestLog NPE on null boxed Integer/Long (chunked byte responses)
## Description
`new ThreadPoolExecutor(threadCount, MAX_ALLOW_THREADS, 60000L, TimeUnit.MICROSECONDS, …)` — `60000L` with `TimeUnit.MICROSECONDS` = 60 ms, not the intended 60 seconds. Burst-spawned non-core threads die after only 60 ms of idle.

## Location
- `shenyu-plugin/shenyu-plugin-logging/shenyu-plugin-logging-tencent-cls/src/main/java/org/apache/shenyu/plugin/tencent/cls/client/TencentClsLogCollectClient.java:163`
- `shenyu-plugin/shenyu-plugin-logging/shenyu-plugin-logging-aliyun-sls/src/main/java/org/apache/shenyu/plugin/aliyun/sls/client/AliyunSlsLogCollectClient.java:184`

## Impact
Callback-executor threads cycle rapidly on any pause in traffic; wasted thread creation/teardown.

## Suggested fix
Use `TimeUnit.MILLISECONDS` (or `SECONDS` with `60L`).

## Related existing
None — distinct from #6794 (Logging pools AbortPolicy + unit bug) which covers the send pools in different clients.

nikhiln64:
Fixes #6884

desensitizeShenyuRequestLog unconditionally calls toString on three boxed numeric fields of ShenyuRequestLog. For byte type media appendResponse invokes logResponse with a null writer, so a chunked response with no Content-Length header reaches desensitize with responseContentLength still null and the collector throws an NPE inside the response doFinally, losing the access log for that request. status and upstreamResponseTime are nullable on the same class of paths.

This change guards each of the three fields with Objects.nonNull before desensitizing, matching the null tolerance the String fields already get from StringUtils.hasLength inside desensitizeForSingleWord. Added AbstractLogCollectorTest covering the null case (chunked byte response shape) and the populated case (values preserved when keywords do not match them).

Note for reviewers, the issue body on #6884 describes a different bug due to an off by one paste in the Aug 4 batch, the analysis matching this title is in the body of #6883, details in my comment on the issue.

Make sure that:

- [x] Your local test passed `./mvnw clean install -Dmaven.javadoc.skip=true` (module scoped build of shenyu-plugin-logging-common with tests).

### DIFF
--- a/shenyu-plugin/shenyu-plugin-logging/shenyu-plugin-logging-common/src/main/java/org/apache/shenyu/plugin/logging/common/collector/AbstractLogCollector.java
+++ b/shenyu-plugin/shenyu-plugin-logging/shenyu-plugin-logging-common/src/main/java/org/apache/shenyu/plugin/logging/common/collector/AbstractLogCollector.java
@@ -186,13 +186,19 @@ public abstract class AbstractLogCollector<T extends AbstractLogConsumeClient<?,
         logInfo.setTimeLocal(desensitizeForSingleWord(GenericLoggingConstant.TIME_LOCAL, logInfo.getTimeLocal(), keyWordMatch, desensitizedAlg));
         logInfo.setMethod(desensitizeForSingleWord(GenericLoggingConstant.METHOD, logInfo.getMethod(), keyWordMatch, desensitizedAlg));
         logInfo.setRequestUri(desensitizeForSingleWord(GenericLoggingConstant.REQUEST_URI, logInfo.getRequestUri(), keyWordMatch, desensitizedAlg));
-        logInfo.setResponseContentLength(Integer.valueOf(desensitizeForSingleWord(GenericLoggingConstant.RESPONSE_CONTENT_LENGTH,
-                logInfo.getResponseContentLength().toString(), keyWordMatch, desensitizedAlg)));
+        if (Objects.nonNull(logInfo.getResponseContentLength())) {
+            logInfo.setResponseContentLength(Integer.valueOf(desensitizeForSingleWord(GenericLoggingConstant.RESPONSE_CONTENT_LENGTH,
+                    logInfo.getResponseContentLength().toString(), keyWordMatch, desensitizedAlg)));
+        }
         logInfo.setRpcType(desensitizeForSingleWord(GenericLoggingConstant.RPC_TYPE, logInfo.getRpcType(), keyWordMatch, desensitizedAlg));
-        logInfo.setStatus(Integer.valueOf(desensitizeForSingleWord(GenericLoggingConstant.STATUS, logInfo.getStatus().toString(), keyWordMatch, desensitizedAlg)));
+        if (Objects.nonNull(logInfo.getStatus())) {
+            logInfo.setStatus(Integer.valueOf(desensitizeForSingleWord(GenericLoggingConstant.STATUS, logInfo.getStatus().toString(), keyWordMatch, desensitizedAlg)));
+        }
         logInfo.setUpstreamIp(desensitizeForSingleWord(GenericLoggingConstant.UP_STREAM_IP, logInfo.getUpstreamIp(), keyWordMatch, desensitizedAlg));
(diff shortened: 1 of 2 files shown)


Aias00 (approved):
## Review: #6899 — fix: NPE in AbstractLogCollector.desensitizeShenyuRequestLog on null boxed Integer/Long

**Verdict: ✅ Approve**

Clean, correct, well-tested null-safety fix for #6884.

### What's correct
- **The NPE was real.** I confirmed the field types in `ShenyuRequestLog`: `private Integer responseContentLength;`, `private Integer status;`, `private Long upstreamResponseTime;`. The old code unconditionally called `.toString()` on all three inside `desensitizeShenyuRequestLog`, so any request reaching desensitize with one of them null (the PR's example: a chunked *byte*-media response where `LoggingServerHttpResponse` passes a null writer and `status` isn't committed) threw inside the response `doFinally`, dropping the entire access log.
- **Guard is correct and minimal.** Each of the three fields is now wrapped in `Objects.nonNull(...)` before `toString()`/`desensitize`. I verified `java.util.Objects` is already imported in `AbstractLogCollector`, so this compiles cleanly. The String fields (`clientIp`, etc.) were already null-tolerant via `StringUtils.hasLength` inside `desensitizeForSingleWord`, so the asymmetry is resolved.
- **Tests cover both directions.** `testDesensitizeToleratesNullBoxedNumericFields` drives the exact failure shape (only `clientIp` set) and asserts no throw + the three numeric fields stay `null` + `clientIp` is still desensitized. `testDesensitizePreservesPopulatedNumericFields` confirms values survive when their keywords aren't in the match set.

### Non-blocking suggestions
1. The PR note about #6884 being an off-by-one paste and the real analysis living in #6883 is fine context, but it'd help a future reader if the PR body briefly stated the *actual* repro (chunked byte response + null writer → null `responseContentLength`/`status`) without requiring them to open the issue. Minor doc clarity.
2. If a keyword *does* match one of these numeric fields, `desensitizeForSingleWord` masks it as a String — that's intended desensitization behavior, just noting it's unchanged by this PR.

### Verdict
Approving. This is the right fix in the right place, with regression tests that exercise the null path. Merge as-is.

nikhiln64:
Thanks for the careful review Aias00, and for confirming the field types and the repro. The concrete failure shape you mentioned, a chunked byte response with a null writer leaving responseContentLength null, is already in the opening paragraph of the description so a future reader has it without opening the issue. I have brought the branch up to date with master, and the workflow runs are sitting in action_required waiting on a maintainer to approve them, so whenever you have a moment could you approve the run so the checks can go green for merge.

nikhiln64:
The two red checks here are the k8s ingress integration tests for apache-dubbo and websocket, and they fail inside the ingress controller reconciler with a NullPointerException on V1Endpoints.getSubsets because the endpoints object came back null, followed by Healthcheck Failed. That is the ephemeral k8s cluster not coming ready in time rather than anything in this change, which only touches shenyu-plugin-logging-common and its test and never goes near ingress. A fresh run of those two jobs should clear them. Thanks again for the review Aias00, and whenever you have a moment could you trigger those two ingress jobs again so this can go green for merge.

Aias00:
> The two red checks here are the k8s ingress integration tests for apache-dubbo and websocket, and they fail inside the ingress controller reconciler with a NullPointerException on V1Endpoints.getSubsets because the endpoints object came back null, followed by Healthcheck Failed. That is the ephemeral k8s cluster not coming ready in time rather than anything in this change, which only touches shenyu-plugin-logging-common and its test and never goes near ingress. A fresh run of those two jobs should clear them. Thanks again for the review Aias00, and whenever you have a moment could you trigger those two ingress jobs again so this can go green for merge.

done

Aias00:
LGTM. thx for your contribution, bro

PR #6900, opened 2026-08-08
fix: evict stale upstream countMap entries in LeastActiveLoadBalance 

### LINKED ISSUE #6891: [BUG] SelectorDataRefresh/RuleDataRefresh NPE on null data list (HTTP long-polling)
## Description
`LeastActiveLoadBalance.countMap` (a `ConcurrentHashMap<String, Long>`) accumulates an entry for every distinct `Upstream.buildDomain()` (protocol+url) ever seen. New domains are added. The map is read to find the minimum-count domain and the selected domain's count is incremented. There is NO removal path: when an upstream is removed from the upstream list (e.g. auto-scaling down, upstream reconfiguration), its `countMap` entry persists indefinitely. Unlike `RoundRobinLoadBalancer` which has a `recyclePeriod`-based cleanup, `LeastActiveLoadBalance` has no eviction logic at all. The filter at line 50 (`domainMap.containsKey(entry.getKey())`) prevents stale entries from being selected, but they still occupy memory and are iterated on every `doSelect` call.

## Location
- `shenyu-loadbalancer/src/main/java/org/apache/shenyu/loadbalancer/spi/LeastActiveLoadBalance.java:37,44-46,55`

## Impact
Slow memory leak proportional to the number of distinct upstream URLs observed over the gateway's lifetime. In environments with dynamic upstream churn (container orchestration, blue-green deploys, ephemeral instances), `countMap` grows without bound. Also, the `entrySet().stream()` scan over stale entries on every request adds O(n) overhead where n includes all historical domains.

## Suggested fix
After building `domainMap` in `doSelect`, remove `countMap` keys not present in `domainMap`: `countMap.keySet().retainAll(domainMap.keySet())` (before the min-selection), or add periodic eviction like `RoundRobinLoadBalancer`'s recycle logic.

## Related existing
None — distinct from "p2c inflight leak" (which covers `P2cLoadBalancer`'s `inflight` counter, a different stat field in a different balancer). The LeastActive `countMap` unbounded-growth issue is not mentioned in any baseline entry or scan2 doc.

juicewcode:
Fixes #6891
- Fix the memory leak in LeastActiveLoadBalance.countMap. Previously entries were only added and never removed, so an upstream's entry lingered forever after it was taken out of the list, causing unbounded memory growth proportional to historical upstream churn and O(n) scan overhead on every selection. Now stale entries are evicted via a time-based periodic sweep (gated by CAS + 60s recyclePeriod) using entrySet().removeIf() on lastUpdate age, rather than an immediate retainAll against the current domain set. This tolerates temporary domain absence and avoids O(n) overhead per selection call. Evicted domains may linger in countMap for up to 60s but are safely excluded from selection by the domainMap.containsKey() filter.

Make sure that:

### DIFF
--- a/shenyu-loadbalancer/src/main/java/org/apache/shenyu/loadbalancer/spi/LeastActiveLoadBalance.java
+++ b/shenyu-loadbalancer/src/main/java/org/apache/shenyu/loadbalancer/spi/LeastActiveLoadBalance.java
@@ -68,6 +68,7 @@ public class LeastActiveLoadBalance extends AbstractLoadBalancer {
             activeCount.increase();
         }
 
+        // A removed domain's entry lingers for up to recyclePeriod, safely excluded from selection meanwhile.
         if (!updateLock.get() && now - lastRecycle > recyclePeriod && updateLock.compareAndSet(false, true)) {
             try {
                 countMap.entrySet().removeIf(item -> now - item.getValue().getLastUpdate() > recyclePeriod);
--- a/shenyu-loadbalancer/src/main/java/org/apache/shenyu/loadbalancer/spi/LeastActiveLoadBalance.java
+++ b/shenyu-loadbalancer/src/main/java/org/apache/shenyu/loadbalancer/spi/LeastActiveLoadBalance.java
@@ -24,8 +24,11 @@ import org.apache.shenyu.spi.Join;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Collectors;
 
 /**
@@ -34,28 +37,93 @@ import java.util.stream.Collectors;
 @Join
 public class LeastActiveLoadBalance extends AbstractLoadBalancer {
 
-    private final Map<String, Long> countMap = new ConcurrentHashMap<>();
+    private final int recyclePeriod = 60000;
+
+    private final ConcurrentMap<String, ActiveCount> countMap = new ConcurrentHashMap<>(16);
+
+    private final AtomicBoolean updateLock = new AtomicBoolean();
+
+    private volatile long lastRecycle;
 
     @Override
     protected Upstream doSelect(final List<Upstream> upstreamList, final LoadBalanceData data) {
+        long now = System.currentTimeMillis();
         Map<String, Upstream> domainMap = upstreamList.stream()
                 .collect(Collectors.toConcurrentMap(Upstream::buildDomain, upstream -> upstream));
 
-        domainMap.keySet().stream()
-                .filter(key -> !countMap.containsKey(key))
-                .forEach(domain -> countMap.put(domain, Long.MIN_VALUE));
-
-        countMap.keySet().retainAll(domainMap.keySet());
+        domainMap.keySet().forEach(domain -> {
+            ActiveCount activeCount = countMap.computeIfAbsent(domain, key -> new ActiveCount(now));
+            activeCount.setLastUpdate(now);
+        });
 
         final String domain = countMap.entrySet().stream()
                 // Ensure that the filtered domain is included in the domainMap.
                 .filter(entry -> domainMap.containsKey(entry.getKey()))
-                .min(Comparator.comparingLong(Map.Entry::getValue))
+                .min(Comparator.comparingLong(entry -> entry.getValue().getCount()))
                 .map(Map.Entry::getKey)
                 .orElse(upstreamList.get(0).buildDomain());
 
-        countMap.computeIfPresent(domain, (key, activated) -> Optional.of(activated).orElse(Long.MIN_VALUE) + 1);
+        ActiveCount activeCount = countMap.get(domain);
+        if (Objects.nonNull(activeCount)) {
+            activeCount.increase();
+        }
+
+        if (!updateLock.get() && now - lastRecycle > recyclePeriod && updateLock.compareAndSet(false, true)) {
+            try {
+                countMap.entrySet().removeIf(item -> now - item.getValue().getLastUpdate() > recyclePeriod);
+                lastRecycle = now;
+            } finally {
+                updateLock.set(false);
+            }
+        }
         return domainMap.get(domain);
     }
-    
+
+    /**
+     * The type Active count.
+     */
+    protected static class ActiveCount {
+
+        private final AtomicLong count = new AtomicLong(Long.MIN_VALUE);
+
+        private volatile long lastUpdate;
+
+        ActiveCount(final long lastUpdate) {
+            this.lastUpdate = lastUpdate;
+        }
+
+        /**
+         * Increase count.
+         */
+        void increase() {
+            count.addAndGet(1);
+        }
+
+        /**
+         * Get count.
+         *
+         * @return the count
+         */
+        long getCount() {
+            return count.get();
+        }
+
+        /**
+         * Gets last update.
+         *
+         * @return the last update
+         */
+        long getLastUpdate() {
+            return lastUpdate;
+        }
+
+        /**
+         * Sets last update.
+         *
+         * @param lastUpdate the last update
+         */
+        void setLastUpdate(final long lastUpdate) {
+            this.lastUpdate = lastUpdate;
+        }
+    }
 }
--- a/shenyu-loadbalancer/src/main/java/org/apache/shenyu/loadbalancer/spi/LeastActiveLoadBalance.java
+++ b/shenyu-loadbalancer/src/main/java/org/apache/shenyu/loadbalancer/spi/LeastActiveLoadBalance.java
@@ -45,6 +45,8 @@ public class LeastActiveLoadBalance extends AbstractLoadBalancer {
                 .filter(key -> !countMap.containsKey(key))
                 .forEach(domain -> countMap.put(domain, Long.MIN_VALUE));
 
+        countMap.keySet().retainAll(domainMap.keySet());
+
         final String domain = countMap.entrySet().stream()
                 // Ensure that the filtered domain is included in the domainMap.
                 .filter(entry -> domainMap.containsKey(entry.getKey()))
--- a/shenyu-loadbalancer/src/test/java/org/apache/shenyu/loadbalancer/spi/LeastActiveLoadBalanceTest.java
+++ b/shenyu-loadbalancer/src/test/java/org/apache/shenyu/loadbalancer/spi/LeastActiveLoadBalanceTest.java
@@ -22,8 +22,10 @@ import org.apache.shenyu.loadbalancer.entity.Upstream;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
+import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 
 /**
  * The type least activity load balance test.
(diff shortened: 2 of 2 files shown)


shenyu-loadbalancer/src/main/java/org/apache/shenyu/loadbalancer/spi/LeastActiveLoadBalance.java
```diff
@@ -45,6 +45,8 @@ protected Upstream doSelect(final List<Upstream> upstreamList, final LoadBalance
                 .filter(key -> !countMap.containsKey(key))
                 .forEach(domain -> countMap.put(domain, Long.MIN_VALUE));
 
+        countMap.keySet().retainAll(domainMap.keySet());
```
nikhiln64: ## Context
Thanks for tracking down the leak, the `countMap` really did grow without bound before this. One thing I would want to sort out before it lands. `LeastActiveLoadBalance` is a singleton (a bare @Join defaults to isSingleton true, and ExtensionLoader caches one instance per algorithm), so `countMap` at `LeastActiveLoadBalance.java:37` is a single JVM wide map shared by every selector and rule that uses leastActive, keyed only by the upstream domain with no selector namespacing. The retainAll you have added runs on every `doSelect` against just the current request's upstream set, so two selectors with different upstream lists end up deleting each other's entries.

> Example: If selector A serves {A1, A2} and selector B serves {B1, B2}, a request through A removes B1 and B2 from the map, and the next request through B removes A1 and A2 and re-adds B1 and B2 at `Long.MIN_VALUE`, discarding the counts B had accumulated. Under interleaved traffic that reset happens on nearly every request, so least active selection collapses toward near fixed selection and the map churns rather than just leaking. 

There is also a smaller lost update where a computeIfPresent increment no ops if another thread's retainAll just removed that key.

## Suggestion
At least gating the cleanup on the selector's own size mismatch rather than an unconditional `retainAll`, would fix the leak without touching other selector.

### NOTE
Happy to be corrected if leastActive is only ever meant to serve one selector at a time, but nothing in the SPI seems to enforce that and RoundRobin assumes the opposite.

nikhiln64 (changes requested):

shenyu-loadbalancer/src/main/java/org/apache/shenyu/loadbalancer/spi/LeastActiveLoadBalance.java
juicewcode: Thanks for the thorough review — you're right, and I've verified each point against the code.
     
  **Verified.** `LeastActiveLoadBalance` is a bare `@Join`, so `isSingleton` defaults to `true` and
  `ExtensionLoader.getJoin("leastActive")` returns one cached instance shared by every selector/rule
  using leastActive. At the call site, `DividePlugin` fetches the upstream list per selector
  (`UpstreamCacheManager.findUpstreamListBySelectorId(selector.getId())`), so the same singleton is
  invoked with different upstream sets. My unconditional `retainAll(domainMap.keySet())` therefore
  really does evict another selector's live entries on every `doSelect`, and the lost `computeIfPresent`
  increment is real — the original fix traded the leak for cross-selector count churn.

  **On the size-mismatch gate.** I agree with the goal, but I think comparing sizes runs into a problem
  here: `countMap` is a single map shared by all selectors and has no selector namespacing, so
  `countMap.size()` is the union of everyone's domains and is not comparable to any single selector's
  `upstreamList.size()`. In a multi-selector deployment the two sizes almost never match, so the gate
  would stay open and the cross-selector deletion would still happen on nearly every request. A size
  comparison also can't detect a member swap that keeps the list length unchanged (e.g. A1 replaced by
  B1, size still 2), so stale entries could linger. In short, without namespacing, "the selector's own
  size" isn't something we can actually measure — deriving one from the list would itself break whenever
  a selector's first upstream changes.

  **Proposed change.** I'd like to keep the single shared map and make eviction time-based instead,
  like `RoundRobinLoadBalancer`'s recycle logic but driven by a timestamp rather than a size comparison:

  ```java
  @Join
  public class LeastActiveLoadBalance extends AbstractLoadBalancer {

      private final int recyclePeriod = 60000;

      private final ConcurrentMap<String, ActiveCount> countMap = new ConcurrentHashMap<>(16);

      private final AtomicBoolean updateLock = new AtomicBoolean();

      private volatile long lastRecycle;

      @Override
      protected Upstream doSelect(final List<Upstream> upstreamList, final LoadBalanceData data) {
          long now = System.currentTimeMillis();
          Map<String, Upstream> domainMap = upstreamList.stream()
                  .collect(Collectors.toConcurrentMap(Upstream::buildDomain, upstream -> upstream));

          domainMap.keySet().forEach(domain -> {
              ActiveCount activeCount = countMap.computeIfAbsent(domain, key -> new ActiveCount(now));
              activeCount.setLastUpdate(now);
          });

          final String domain = countMap.entrySet().stream()
                  .filter(entry -> domainMap.containsKey(entry.getKey()))
                  .min(Comparator.comparingLong(entry -> entry.getValue().getCount()))
                  .map(Map.Entry::getKey)
                  .orElse(upstreamList.get(0).buildDomain());

          ActiveCount activeCount = countMap.get(domain);
          if (Objects.nonNull(activeCount)) {
              activeCount.increase();
          }

          if (!updateLock.get() && now - lastRecycle > recyclePeriod && updateLock.compareAndSet(false, true)) {
              try {
                  countMap.entrySet().removeIf(item -> now - item.getValue().getLastUpdate() > recyclePeriod);
                  lastRecycle = now;
              } finally {
                  updateLock.set(false);
              }
          }
          return domainMap.get(domain);
      }

      protected static class ActiveCount {

          private final AtomicLong count = new AtomicLong(Long.MIN_VALUE);

          private volatile long lastUpdate;

          ActiveCount(final long lastUpdate) {
              this.lastUpdate = lastUpdate;
          }

          void increase() {
              count.addAndGet(1);
          }

          long getCount() {
              return count.get();
          }

          long getLastUpdate() {
              return lastUpdate;
          }

          void setLastUpdate(final long lastUpdate) {
              this.lastUpdate = lastUpdate;
          }
      }
  }
```
  Key points:
  - Live entries are refreshed on every request, so they are never stale and can't be evicted by
  another selector's cleanup — no cross-selector deletion.
  - Only entries whose lastUpdate is older than RECYCLE_PERIOD are evicted, i.e. exactly the
  upstreams that were removed from the lists — the leak is bounded to ~`RECYCLE_PERIOD`.
  - Eviction is throttled to at most once per RECYCLE_PERIOD and guarded by an update lock, so steady
  state has no per-request cleanup overhead and no concurrent eviction.

  I'll update the tests to cover stale-entry eviction and the multi-selector isolation case. Happy to
  adjust RECYCLE_PERIOD or the eviction gating if you'd prefer a different approach — please let me
  know if you see any issues.

nikhiln64: This looks like the right direction, and you are right that the size gate does not work here. On a single shared map `countMap.size()` is the union of every selector's domains, so it is not comparable to one selector's `upstreamList.size()`, and it cannot see a same length member swap, so my suggestion would have left the cross selector deletion in place. The time based recycle avoids that cleanly. Refreshing `lastUpdate` for every live domain on each request means an entry can only be evicted after it has genuinely gone quiet for `recyclePeriod`, so one selector's cleanup can no longer drop another selector's live entries and the leak is bounded to about `recyclePeriod`. The updateLock and the once per period throttle keep the cleanup off the hot path, and the selected domain is always one you refreshed this call, so the countMap.get(domain) with the `Objects.nonNull` guard cannot be caught by a concurrent removeIf. The concurrency reads as sound to me.

The one thing I would think about is the `Long.MIN_VALUE` seed on a revived entry. If an upstream goes quiet longer than `recyclePeriod` it gets recycled, and when traffic returns its `ActiveCount` is recreated at `Long.MIN_VALUE`. Because selection picks the minimum count, that upstream is then chosen on almost every request until its counter climbs back to its peers, and since the count is cumulative and never decremented that catch up can take as many requests as the peers have served in total, so a node that just came back can absorb nearly all the traffic for a long stretch rather than a brief burst. A brand new upstream already has this property today, but recycling means an idle then active upstream hits it too. It might be worth seeding a new or revived entry to the current minimum of its live peers rather than `Long.MIN_VALUE`, so it rejoins at parity. Not a blocker, more a behavior to decide on consciously.

On the tests, the stale eviction and multi selector isolation cases you mentioned are the ones I would want to see. One more worth adding is the revive case: let an entry age past recyclePeriod so it is recycled, then send traffic again and assert the distribution does not collapse onto the revived node. Thanks for turning this around so quickly.

Aias00 (dismissed):
## Review: #6900 — fix: evict stale upstream countMap entries in LeastActiveLoadBalance

**Verdict: ✅ Approve (with one documentation correction — the description doesn't match the code)**

The underlying fix is correct and the test suite is genuinely strong (including cross-selector isolation). One thing to fix before merge: the PR body.

### What's correct
- **The memory-leak fix is real.** Previously `countMap` (a `ConcurrentHashMap<String, Long>`) only ever grew: every observed domain was added and never removed, so it leaked in proportion to upstream churn. The new `ActiveCount` wrapper (`count` + `lastUpdate`) plus the recycle path bounds it.
- **Eviction is time-based and concurrency-safe.** In `doSelect`, every domain present in the current `upstreamList` gets `lastUpdate = now`. A periodic cleanup (gated by `updateLock` CAS + `now - lastRecycle > recyclePeriod(60s)`) does `countMap.entrySet().removeIf(item -> now - item.getValue().getLastUpdate() > recyclePeriod)`. `removeIf` on a `ConcurrentHashMap` entrySet is safe (weakly-consistent iterator), and the `AtomicBoolean` CAS ensures only one thread runs the sweep — no double-sweep, no CME.
- **Selection semantics preserved.** New domains initialize `count` to `Long.MIN_VALUE` (so a freshly-added upstream is preferred first, exactly as before), and the min is still taken only over entries whose domain is in the current `domainMap`, so evicted/stale entries never affect routing.
- **Tests are excellent.** `testRemoveStaleCountMapEntries` covers the basic eviction; `testCountMapNotAffectOtherSelector` and `testCountMapNotAffectOtherSelectorWhenFirstRemoved` verify that removing an upstream from one selector does **not** evict a live domain that only belongs to a different selector sharing the same load-balancer instance. That multi-selector isolation is the subtle part and it's covered.

### Issue to fix (non-blocking for approval, please correct the description)
- **The PR body is inaccurate about the mechanism.** It says entries are evicted "via `countMap.keySet().retainAll(domainMap.keySet())` before selection." The actual code uses a **time-based** sweep (`removeIf` on `lastUpdate` age > 60s), not an immediate `retainAll` against the current domain set. The implemented approach is actually *better* (it tolerates a domain being temporarily absent and avoids an O(n) `retainAll` on every call), but the description should be updated so reviewers/maintainers aren't confused about what shipped.
- **Minor behavioral note (acceptable):** because eviction is time-based, a permanently-removed domain's entry lingers up to ~60s before being swept. During that window it's correctly excluded from selection (the `domainMap.containsKey` filter), so there's no correctness impact — just a bounded memory tail. Worth a one-line comment in `doSelect` explaining the 60s grace window.

### Verdict
Approving. This genuinely closes the unbounded-growth leak, keeps least-active selection correct, and the tests prove the tricky cross-selector case. Just align the PR description with the time-based implementation.

juicewcode:
@Aias00 Thanks for the review and approval! 

- Updated the PR description to accurately reflect the time-based eviction mechanism (CAS + 60s recyclePeriod via `removeIf` on `lastUpdate` age, not immediate `retainAll`).
- Added an inline comment in `doSelect` explaining the 60s grace window.

Aias00 (approved):
## Review: #6900 — fix: evict stale upstream countMap entries in LeastActiveLoadBalance

**Verdict: ✅ Approve**

Re-reviewing after the new commits (the branch now also contains a merge from master and a docs commit). The code is **unchanged from my prior approval** except for one addition — a clarifying inline comment — and that change is exactly in line with my previous review.

### What changed since last review
- A new commit `docs: Add inline comment explaining 60s grace window in doSelect` adds this line in `LeastActiveLoadBalance.doSelect`:
  ```java
  // A removed domain's entry lingers for up to recyclePeriod, safely excluded from selection meanwhile.
  ```
  This directly addresses the minor note I left in my earlier review (the 60s grace window wasn't explained in-code). Good — the comment is accurate: entries of permanently-removed domains correctly remain excluded from selection during the window (via the `domainMap.containsKey` filter), so there is no routing impact, just a bounded memory tail. No further action needed on that point.
- The merge-from-master commits did not touch the load-balancer logic.

### Still correct (unchanged from prior pass)
- **The memory-leak fix is real.** `countMap` previously only grew (`ConcurrentHashMap<String, Long>`), leaking with upstream churn. The `ActiveCount` wrapper (`count` + `lastUpdate`) plus the time-based recycle path now bounds it.
- **Eviction is time-based and concurrency-safe.** Every domain present in the current `upstreamList` gets `lastUpdate = now`. A periodic sweep gated by `updateLock` CAS + `now - lastRecycle > recyclePeriod(60s)` does `countMap.entrySet().removeIf(item -> now - item.getValue().getLastUpdate() > recyclePeriod)`. `removeIf` on a `ConcurrentHashMap` entrySet uses a weakly-consistent iterator, and the `AtomicBoolean` CAS guarantees a single sweeper — no double-sweep, no CME.
- **Selection semantics preserved.** New domains still init `count` to `Long.MIN_VALUE` (fresh upstream preferred first, as before); the min is still taken only over entries whose domain is in the current `domainMap`, so stale/evicted entries never affect routing.
- **Tests remain strong.** `testRemoveStaleCountMapEntries`, `testCountMapNotAffectOtherSelector`, and `testCountMapNotAffectOtherSelectorWhenFirstRemoved` cover both basic eviction and the subtle multi-selector isolation (a live domain belonging only to another selector sharing the same LB instance is not evicted). This is the hard part and it's verified.

### One outstanding non-blocking note (no change needed to merge)
- **The PR body is still inaccurate about the mechanism.** It states entries are evicted "via `countMap.keySet().retainAll(domainMap.keySet())` before selection." The actual code uses a **time-based** sweep (`removeIf` on `lastUpdate` age > 60s), not an immediate `retainAll` against the current domain set. The implemented approach is actually *better* (tolerates a domain being temporarily absent, avoids an O(n) `retainAll` on every call), but the description should be corrected so maintainers/reviewers aren't misled about what shipped. This does not block approval.

### Verdict
Approving again. The fix genuinely closes the unbounded-growth leak, keeps least-active selection correct, the new comment improves readability, and the tests prove the tricky cross-selector case. Please just align the PR description with the time-based implementation.

Aias00:
hi, pls add my wechat: aias00

PR #6901, opened 2026-08-08
[ISSUE-6523] Add break after successful long polling to avoid unnecessary retries

### LINKED ISSUE #6523: [BUG] HTTP sync long polling repeats successful polls ten times per loop
### Search before asking

### Apache ShenYu Component

shenyu-sync-data-http

### What happened

`HttpSyncDataService.HttpLongPollingTask.run()` retries failed long-poll requests with a `for (time = 1; time <= retryTimes; time++)` loop, but it does not leave that retry loop after a successful `doLongPolling(server)` call.

Current code in `shenyu-sync-data-center/shenyu-sync-data-http/src/main/java/org/apache/shenyu/sync/data/http/HttpSyncDataService.java`:

```java
while (RUNNING.get()) {
    int retryTimes = 10;
    for (int time = 1; time <= retryTimes; time++) {
        try {
            //do long polling.
            doLongPolling(server);
        } catch (Exception e) {
            if (time < retryTimes) {
                ThreadUtils.sleep(TimeUnit.SECONDS, 5);
                continue;
            }
            ThreadUtils.sleep(TimeUnit.MINUTES, 5);
        }
    }
}
```

When `doLongPolling(server)` succeeds, execution continues to the next `for` iteration. A single outer loop can therefore perform 10 successful long-poll requests back-to-back instead of one successful poll followed by the next outer loop cycle.

This can create unnecessary long-poll traffic and makes the retry counter behave as a fixed repeat count even when no retry is needed.

### What you expected to happen

After a successful `doLongPolling(server)` call, the retry loop should stop for the current outer iteration, for example by `break`-ing out of the `for` loop. The retry loop should only continue after failures.

### How to reproduce

1. Configure ShenYu bootstrap to use HTTP sync.
2. Start admin and bootstrap while the long-poll endpoint is healthy.
3. Observe that `HttpLongPollingTask.run()` can execute `doLongPolling(server)` up to 10 times per outer `while (RUNNING.get())` loop because success does not break the retry loop.

### Debug logs

_No response_

### Environment

Current master branch.

### Are you willing to submit PR?

YALI58:
### Bug Description

`HttpLongPollingTask.run()` retries failed long-poll requests with a `for (time = 1; time <= retryTimes; time++)` loop, but it does not leave that retry loop after a successful `doLongPolling(server)` call. A single outer loop can therefore perform 10 successful long-poll requests back-to-back instead of one.

### Fix

Added `break` after `doLongPolling(server)` succeeds so the retry loop exits on success and only continues on failure.

Closes #6523

### DIFF
--- a/shenyu-sync-data-center/shenyu-sync-data-http/src/main/java/org/apache/shenyu/sync/data/http/HttpSyncDataService.java
+++ b/shenyu-sync-data-center/shenyu-sync-data-http/src/main/java/org/apache/shenyu/sync/data/http/HttpSyncDataService.java
@@ -281,6 +281,7 @@ public class HttpSyncDataService implements SyncDataService {
                     try {
                         //do long polling.
                         doLongPolling(server);
+                        break;
                     } catch (Exception e) {
                         // print warnning LOG.
                         if (time < retryTimes) {


YALI58:
Hi @dengliming, all checks have passed. Could you please review and approve when you have time? Thanks!

dengliming (approved):
Thank you for your contribution!

PR #6902, opened 2026-08-10
feat: implement MQTT Last Will and Testament (LWT)

wy471x:
Parse and store will fields (topic, message, QoS, retain) in WillRepository on CONNECT. Publish will message to subscribers on ungraceful disconnect via channelInactive hook. Clear will on graceful DISCONNECT. Fix DISCONNECT message dispatch in MqttFactory that was previously dropped.

<!-- Describe your PR here; e.g. Fixes #issueNo -->

<!--
Thank you for proposing a pull request. This template will guide you through the essential steps necessary for a pull request.
-->
Make sure that:

- [X] You have read the [contribution guidelines](https://shenyu.apache.org/community/contributor-guide).
- [X] You submit test cases (unit or integration tests) that back your changes.
- [X] Your local test passed `./mvnw clean install -Dmaven.javadoc.skip=true`.

## Summary

### Core changes:                                                                                                                                                                      

  - WillRepository.java (new) — A ConcurrentHashMap-backed repository that stores WillEntry (topic, message bytes, QoS, retain flag) keyed by Netty Channel. Implements the existing 
  BaseRepository interface.
  - Connect.java — On CONNECT, if isWillFlag() is true, extracts the will fields from the MQTT CONNECT payload and stores them via WillRepository.                                   
  - Disconnect.java — On graceful DISCONNECT, removes the will from WillRepository (so it won't fire). Replaces the old // todo Last words placeholder.                              
  - MqttTransportHandler.java — Overrides channelInactive() to detect ungraceful disconnects. If a will is still present for that channel when it goes inactive, it publishes the    
  will via Publish.publishWill() and then clears it.                                                                                                                                 
  - Publish.java — New publishWill() static method that looks up subscribers for the will topic and writes the will message as an MQTT PUBLISH packet to each active subscriber      
  channel.                                                                                                                                                                           
  - MqttFactory.java — Fixes a bug where DISCONNECT was falling through to PUBACK/default instead of calling messageType.disconnect(ctx).
  - pom.xml — Adds JUnit Jupiter, Mockito, and maven-surefire-plugin with --add-opens JVM args for testing.                                                                          
                                                                                                                                                                                     
### Tests (5 new test files): 

- ConnectTest, DisconnectTest, MqttTransportHandlerTest, PublishWillTest, and WillRepositoryTest — covering will storage on connect, will clearance on disconnect, will firing on channel inactivity, inactive channel skipping, empty subscriber handling, and QoS/retain flag propagation. 

[...]

PR #6906, opened 2026-08-12
feat: implement MQTT wildcard subscription matching

### LINKED ISSUE #6851: [BUG] Wildcard subscription matching not implemented — +/-# subscriptions never receive messages
## Description
Subscriptions containing `+` (single-level) or `#` (multi-level) are stored under the literal wildcard string and never matched against published topic names; the matcher is an explicit TODO. `Publish.send` looks up `SubscribeRepository.get(topic)` with the literal published topic (plain `ConcurrentHashMap.getOrDefault` exact-key lookup).

## Location
- `shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/Subscribe.java:63` (`// todo Regular match`)
- `shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/repositories/SubscribeRepository.java:90-92`

## Impact
Wildcard subscriptions never receive any message — a core MQTT feature is silently broken.

## Suggested fix
Implement topic-filter matching per MQTT-4.7 (`+` single level, `#` multi-level at end) and match published topics against all stored filters on each PUBLISH.

wy471x:

<!--
-->
Make sure that:

## Summary:                                                                                                                     
                                            
Problem                                                                                                                                                                            
                                                                                                                                                                                     
Subscriptions using + (single-level wildcard) or # (multi-level wildcard) were silently broken. Publish.send() performed an exact-key lookup (ConcurrentHashMap.getOrDefault), so a publish to sensor/room1/temperature would never match a subscription filter like sensor/+/temperature.                                                                            
                  
Changes

  1. New: TopicMatcher.java — Utility class implementing MQTT topic filter matching per the MQTT-4.7 spec:
    - + matches exactly one topic level
    - # matches any number of levels (must appear at the end of the filter)
    - Wildcards at the first level do not match $-prefixed topics
  3. Modified: SubscribeRepository.java — Added getChannelsByTopic(String topic) method that iterates over all stored subscription filters and returns channels whose filter matches 
  the published topic using TopicMatcher.matches().                                                                                                                                  
  4. Modified: Publish.java:116 — Changed send() from get(topic) (exact-key lookup) to getChannelsByTopic(topic) (wildcard-aware matching).                                          
  5. New: TopicMatcherTest.java — 6 unit tests covering: exact match, + single-level, # multi-level, mixed wildcards, $ topic protection, and null inputs.

close [#6851](https://github.com/apache/shenyu/issues/6851)

### DIFF
--- a/shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/Publish.java
+++ b/shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/Publish.java
@@ -113,7 +113,7 @@ public class Publish extends MessageType {
     }
 
     private void send(final String topic, final ByteBuf payload, final int packetId) {
-        List<Channel> channels = Singleton.INST.get(SubscribeRepository.class).get(topic);
+        List<Channel> channels = Singleton.INST.get(SubscribeRepository.class).getChannelsByTopic(topic);
         //// todo thread pool
         channels.parallelStream().forEach(channel -> {
             if (channel.isActive()) {
--- a/shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/Subscribe.java
+++ b/shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/Subscribe.java
@@ -61,36 +61,40 @@ public class Subscribe extends MessageType {
         int packetId = msg.variableHeader().messageId();
 
         //// todo Regular match
-        List<String> ackTopics = mqttTopicSubscriptions
+        List<MqttTopicSubscription> validSubscriptions = mqttTopicSubscriptions
                 .stream()
                 .filter(topicSub -> topicSub.qualityOfService() != FAILURE)
-                .map(MqttTopicSubscription::topicName)
+                .filter(topicSub -> TopicMatcher.isValidFilter(topicSub.topicName()))
                 .collect(Collectors.toList());
 
-        Singleton.INST.get(SubscribeRepository.class).add(ctx.channel(), mqttTopicSubscriptions);
+        Singleton.INST.get(SubscribeRepository.class).add(ctx.channel(), validSubscriptions);
 
-        for (String ackTopic : ackTopics) {
-            String message = Singleton.INST.get(TopicRepository.class).get(ackTopic);
+        for (MqttTopicSubscription subscription : validSubscriptions) {
+            String message = Singleton.INST.get(TopicRepository.class).get(subscription.topicName());
             if (StringUtils.isNotEmpty(message)) {
-                sendSubMessage(ackTopic, message, packetId, channel);
+                sendSubMessage(subscription.topicName(), message, packetId, channel);
             }
         }
 
-        sendSubAckMessage(packetId, ackTopics, channel);
+        sendSubAckMessage(packetId, mqttTopicSubscriptions, channel);
     }
 
     /**
      * call back request of message.
      * @param packetId packetId
-     * @param ackTopics ackTopics
+     * @param subscriptions subscriptions
      * @param channel channel
      */
-    private void sendSubAckMessage(final int packetId, final List<String> ackTopics, final Channel channel) {
+    private void sendSubAckMessage(final int packetId, final List<MqttTopicSubscription> subscriptions, final Channel channel) {
 
         List<Integer> qos = new ArrayList<>();
-        for (int i = 0; i < ackTopics.size(); i++) {
-            // default qos 0
-            qos.add(MqttQoS.AT_MOST_ONCE.value());
+        for (MqttTopicSubscription subscription : subscriptions) {
+            // invalid topic filters are rejected with 0x80, otherwise default qos 0
+            if (TopicMatcher.isValidFilter(subscription.topicName())) {
+                qos.add(MqttQoS.AT_MOST_ONCE.value());
+            } else {
+                qos.add(MqttQoS.FAILURE.value());
+            }
         }
 
         MqttFixedHeader fixedHeader = new MqttFixedHeader(MqttMessageType.SUBACK, false, AT_MOST_ONCE,
--- a/shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/TopicMatcher.java
+++ b/shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/TopicMatcher.java
@@ -73,4 +73,31 @@ public final class TopicMatcher {
 
         return filterLen == topicLen;
     }
+
+    /**
+     * Validate a topic filter per MQTT-4.7.1: wildcards must occupy an entire
+     * level, and # must be the last level. Filters must not be empty or
+     * contain the null character.
+     *
+     * @param filter the subscription topic filter
+     * @return true if the filter is valid
+     */
+    public static boolean isValidFilter(final String filter) {
+        if (Objects.isNull(filter) || filter.isEmpty() || filter.indexOf((char) 0) >= 0) {
+            return false;
+        }
+        String[] levels = filter.split("/", -1);
+        for (int i = 0; i < levels.length; i++) {
+            String level = levels[i];
+            if (level.indexOf('+') >= 0 || level.indexOf('#') >= 0) {
+                if (level.length() > 1) {
+                    return false;
+                }
+                if ("#".equals(level) && i != levels.length - 1) {
+                    return false;
+                }
+            }
+        }
+        return true;
+    }
 }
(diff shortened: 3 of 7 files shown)


Aias00 (dismissed):
## Review: #6906 — feat: implement MQTT wildcard subscription matching

**Verdict: ✅ Approve (with two follow-up suggestions, one a real correctness edge case)**

This fills a genuinely broken feature — wildcard subscriptions silently never matched because `Publish.send` did an exact `Map.get`. The fix is well-structured.

### What's correct
- **`TopicMatcher` is spec-accurate.** I traced the algorithm across the test matrix and beyond:
  - `+` matches exactly one level (`sport/+/player1` ❌ `sport/tennis/stadium/player1`, ✓ `sport/football/player1`);
  - `#` matches any number of levels incl. the parent (`sport/#` ✓ `sport`);
  - `#` only matches when it's its own level (`sport#` / `sport/tennis#` correctly rejected);
  - `$`-prefixed topics are not matched by a *leading* wildcard (`#`/`+` → false) but are matched by explicit `$SYS/#` / `$SYS/+` — exactly MQTT-4.7.2-1;
  - null inputs return false (no NPE).
- **The `get()` exact-lookup method is retained and still used** by `add()`/`remove()` internally, so this isn't introducing dead code — good call keeping it.
- **`getChannelsByTopic`** is a clean O(N) scan that delegates entirely to `TopicMatcher`; no logic duplicated.
- **`TopicMatcherTest` is thorough** — exact, single-level, multi-level, mixed, `$`-topic, and null cases all covered.

### Suggestions (non-blocking)
1. **Duplicate delivery on overlapping subscriptions (real, please track as a follow-up).** `getChannelsByTopic` does `result.addAll(entry.getValue())` over every matching filter. If one client holds two overlapping subscriptions (e.g. `sport/#` *and* `#`), it appears under both keys, so a publish to `sport/x` adds the same `Channel` twice → the client receives the message twice. MQTT requires at most one delivery per publish per client. Collect into a `Set<Channel>` (or `LinkedHashSet` if you care about order/stability) before returning to avoid this.
2. **Performance fast-path.** Every publish now scans *all* subscriptions. For the very common case where the topic has an exact (non-wildcard) subscriber, you could `result.addAll(get(topic))` first (O(1) exact hit) and then only scan filters containing `+`/`#`. Not necessary for correctness, just a scale consideration.
3. **Minor:** invalid filters (e.g. `#` not as its own level, or trailing text after `#`) silently return `false` here. Optionally reject malformed filters at subscription time in `Subscribe.add` so bad subscriptions fail fast instead of silently never matching.

### Verdict
Approving. The core matching logic is correct and well-tested, and `get()` is correctly preserved. Suggestion #1 (dedupe) is worth a quick follow-up PR before wildcard support sees production traffic with multi-subscription clients.

wy471x:
> ## Review: #6906 — feat: implement MQTT wildcard subscription matching
> **Verdict: ✅ Approve (with two follow-up suggestions, one a real correctness edge case)**
> 
> This fills a genuinely broken feature — wildcard subscriptions silently never matched because `Publish.send` did an exact `Map.get`. The fix is well-structured.
> 
> ### What's correct
> * **`TopicMatcher` is spec-accurate.** I traced the algorithm across the test matrix and beyond:
>   
>   * `+` matches exactly one level (`sport/+/player1` ❌ `sport/tennis/stadium/player1`, ✓ `sport/football/player1`);
>   * `#` matches any number of levels incl. the parent (`sport/#` ✓ `sport`);
>   * `#` only matches when it's its own level (`sport#` / `sport/tennis#` correctly rejected);
>   * `$`-prefixed topics are not matched by a _leading_ wildcard (`#`/`+` → false) but are matched by explicit `$SYS/#` / `$SYS/+` — exactly MQTT-4.7.2-1;
>   * null inputs return false (no NPE).
> * **The `get()` exact-lookup method is retained and still used** by `add()`/`remove()` internally, so this isn't introducing dead code — good call keeping it.
> * **`getChannelsByTopic`** is a clean O(N) scan that delegates entirely to `TopicMatcher`; no logic duplicated.
> * **`TopicMatcherTest` is thorough** — exact, single-level, multi-level, mixed, `$`-topic, and null cases all covered.
> 
> ### Suggestions (non-blocking)
> 1. **Duplicate delivery on overlapping subscriptions (real, please track as a follow-up).** `getChannelsByTopic` does `result.addAll(entry.getValue())` over every matching filter. If one client holds two overlapping subscriptions (e.g. `sport/#` _and_ `#`), it appears under both keys, so a publish to `sport/x` adds the same `Channel` twice → the client receives the message twice. MQTT requires at most one delivery per publish per client. Collect into a `Set<Channel>` (or `LinkedHashSet` if you care about order/stability) before returning to avoid this.
> 2. **Performance fast-path.** Every publish now scans _all_ subscriptions. For the very common case where the topic has an exact (non-wildcard) subscriber, you could `result.addAll(get(topic))` first (O(1) exact hit) and then only scan filters containing `+`/`#`. Not necessary for correctness, just a scale consideration.
> 3. **Minor:** invalid filters (e.g. `#` not as its own level, or trailing text after `#`) silently return `false` here. Optionally reject malformed filters at subscription time in `Subscribe.add` so bad subscriptions fail fast instead of silently never matching.
> 
> ### Verdict
> Approving. The core matching logic is correct and well-tested, and `get()` is correctly preserved. Suggestion #1 (dedupe) is worth a quick follow-up PR before wildcard support sees production traffic with multi-subscription clients.

Thank you for the code review on this PR.

Fixes for the three review comments:                                                                                           
                                                                                                                                                                                     
  1. Duplicate delivery on overlapping subscriptions — SubscribeRepository.getChannelsByTopic now collects channels into a LinkedHashSet before returning, so a client holding       
  overlapping filters (e.g. sport/# and #) receives at most one delivery per publish, per MQTT spec.
  2. Performance fast-path — exact topic subscribers are added via an O(1) map lookup first; the wildcard scan then skips filters without +/#.                                       
  3. Malformed filters fail fast — added TopicMatcher.isValidFilter (MQTT-4.7.1 rules); Subscribe registers only valid filters and sends SUBACK return code 0x80 (FAILURE) for       
  invalid ones.                                                                                                                                                                      
                                                                                                                                                                                     
Tests added: TopicMatcherTest validation cases, new SubscribeRepositoryTest (dedup/fast-path), new SubscribeTest (rejection + SUBACK codes).

Aias00 (dismissed):
## Summary (re-review after update)

Re-reviewed after the branch was updated (prior review was dismissed by a rebase). The current version already addresses the duplicate-delivery concern I raised before: `getChannelsByTopic` now collects results into a `LinkedHashSet`, so a client with overlapping subscriptions (`sport/#` and `#`) receives the message at most once — exactly as MQTT requires.

What is correct:
- New `TopicMatcher` implements `matches(filter, topic)` and `isValidFilter(filter)` per MQTT-4.7, including the `$`-topic rule (leading `+`/`#` must not match `$SYS`-style topics) and the rule that wildcards occupy an entire level with `#` last.
- `Subscribe.subscribe` filters out invalid filters and rejects them in the SUBACK with QoS `FAILURE` (0x80) while accepting valid ones; only valid subscriptions are stored.
- `SubscribeRepository.getChannelsByTopic` scans stored filters, **dedupes channels via `LinkedHashSet`** (overlapping wildcard filters deliver once per client), and uses the matcher for wildcard resolution.
- `Publish.send` now calls `getChannelsByTopic` so publishes reach wildcard subscribers.
- Tests are strong: `TopicMatcherTest` (exact/single/multi/mixed wildcards, `$` topics, null, valid/invalid filters), `SubscribeTest` (invalid filter rejected with 0x80), and `SubscribeRepositoryTest` (exact/wildcard/dedup/multiple subscribers).

## Conclusion

Approved. The matching logic is correct against the MQTT spec, the dedupe I previously flagged is now present, and the tests back the behavior.

## Non-blocking suggestions
- **Performance:** `getChannelsByTopic` iterates the entire `TOPIC_CHANNEL_FACTORY` map on every publish (O(number of distinct subscription filters) per publish). For shenyu's expected scale this is acceptable; if subscription counts grow, an index of wildcard filters would help. Not a blocker.
- **Operator precedence nit:** `filter.equals(topic) || filter.indexOf('+') < 0 && filter.indexOf('#') < 0` relies on `&&` binding tighter than `||`. Correct today, but parentheses would make intent clearer and prevent future mis-edits.
- **Merge ordering:** this PR and #6902 add the same `getChannelsByTopic` to `SubscribeRepository`, while #6913 rewrites `SubscribeRepository` to `Map<Channel, MqttQoS>`. Merge #6902/#6906 first and rebase #6913 (or merge #6913 last) to avoid a conflict. Coordinate the three MQTT PRs.

Thanks for adding the dedupe.

dengliming (changes requested):
This test looks flaky / incorrect because it calls `new Subscribe().subscribe(ctx, msg)` on a plain `EmbeddedChannel`, but `Subscribe.subscribe(...)` returns early when `isConnected(channel)` is false. In that case `repository.add(...)` is never reached, so `awaitUntil(() -> repository.get("sport/#").contains(channel))` can only time out.

Could we update the test setup to put the channel into the same connected/authenticated state that `MessageType.isConnected(...)` expects before invoking `subscribe(...)`? Otherwise this test is asserting a post-condition that the method never has a chance to produce.

Aias00 (changes requested):

PR #6907, opened 2026-08-12
fix(admin): tag_relation updateByPrimaryKeySelective sets non-existent name column

### LINKED ISSUE #6830: [BUG] tag_relation updateByPrimaryKeySelective sets non-existent name column instead of api_id
## Description
`tag_relation updateByPrimaryKeySelective` contains `<if test="apiId != null"> name = #{apiId, jdbcType=VARCHAR}, </if>`. The `tag_relation` table has columns `id, api_id, tag_id, date_created, date_updated` — there is **no `name` column**. When `apiId` is non-null, the generated SQL is `UPDATE tag_relation SET name = ? ...` → `Unknown column 'name'` SQL error. Reachable: `TagRelationServiceImpl.update()` calls `tagRelationMapper.updateByPrimaryKeySelective(tagRelationDO)`, and `TagRelationDO.buildTagRelationDO` populates `apiId` from the DTO.

## Location
- `shenyu-admin/src/main/resources/mappers/tag-relation-sqlmap.xml:125`
- `shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/TagRelationServiceImpl.java:54`

## Impact
Updating a tag-relation whose `apiId` is set throws a runtime SQL exception (HTTP 500) rather than performing the update; the tag-relation edit path is broken whenever apiId is supplied.

## Suggested fix
Change line 125 to `api_id = #{apiId, jdbcType=VARCHAR},`.

## Related existing
Distinct from N23 (#6692, `TagServiceImpl.delete` no cascade) and N55 (TagDTO validation). This is a wrong-column SQL defect in `tag-relation-sqlmap`.

yangjj-iso:
Fixes #6830

### Problem

`tag-relation-sqlmap.xml` line 125 emits `name = #{apiId, jdbcType=VARCHAR}` in the `apiId` branch of `updateByPrimaryKeySelective`, but `tag_relation` only has the columns `id`, `api_id`, `tag_id`, `date_created` and `date_updated`. There is no `name` column.

This is reachable from the REST layer: `PUT /tag-relation/id/{id}` → `TagRelationController.updateTagRelation` → `TagRelationServiceImpl.update` → `TagRelationDO.buildTagRelationDO`, which always populates `apiId` from the DTO. Whenever the request body carries an `apiId`, the `<if>` branch fires and the statement becomes `UPDATE tag_relation SET name = ? ...`, which fails with a SQL grammar error instead of performing the update.

The sibling `updateByPrimaryKey` in the same mapper (line 137) already uses `api_id`, which confirms this is a copy-paste defect rather than intent.

### Change

- `tag-relation-sqlmap.xml`: `name` → `api_id` in `updateByPrimaryKeySelective`.
- `TagRelationMapperTest`: add `testUpdateByPrimaryKeySelective`. This was the only `TagRelationMapper` method the test class did not cover, which is why the defect went unnoticed. The test inserts a row, updates `apiId`, asserts the new value is persisted, and deletes the row so it does not perturb the sibling tests that assert on row counts.

### Verification

Run against JDK 17 (matching the CI matrix):

With the fix applied:

```
[INFO] You have 0 Checkstyle violations.
[INFO] Tests run: 8, Failures: 0, Errors: 0, Skipped: 0 - in org.apache.shenyu.admin.mapper.TagRelationMapperTest
[INFO] BUILD SUCCESS
```

With the mapper change reverted and only the new test in place, to confirm the test actually pins the defect:

```
[ERROR] TagRelationMapperTest.testUpdateByPrimaryKeySelective:96 » BadSqlGrammar
### Cause: org.h2.jdbc.JdbcSQLSyntaxErrorException: Column "NAME" not found
```

Make sure that:

- [x] Your local test passed. Scope note: I ran `./mvnw -pl shenyu-admin -am -DskipTests install` followed by `./mvnw -pl shenyu-admin test -Dtest=TagRelationMapperTest` with Checkstyle enabled, rather than a full-reactor `clean install`. The change touches a single MyBatis statement in `shenyu-admin`, so no other module is affected.

### DIFF
--- a/shenyu-admin/src/main/resources/mappers/tag-relation-sqlmap.xml
+++ b/shenyu-admin/src/main/resources/mappers/tag-relation-sqlmap.xml
@@ -122,7 +122,7 @@
                 date_updated = #{dateUpdated, jdbcType=TIMESTAMP},
             </if>
             <if test="apiId != null">
-                name = #{apiId, jdbcType=VARCHAR},
+                api_id = #{apiId, jdbcType=VARCHAR},
             </if>
             <if test="tagId != null">
                 tag_id = #{tagId, jdbcType=VARCHAR},
--- a/shenyu-admin/src/test/java/org/apache/shenyu/admin/mapper/TagRelationMapperTest.java
+++ b/shenyu-admin/src/test/java/org/apache/shenyu/admin/mapper/TagRelationMapperTest.java
@@ -87,6 +87,23 @@ public class TagRelationMapperTest extends AbstractSpringIntegrationTest {
     }
 
     @Test
+    public void testUpdateByPrimaryKeySelective() {
+        TagRelationDO record = buildTagRelationDO();
+        int count = tagRelationMapper.insertSelective(record);
+        assertThat(count, greaterThan(0));
+
+        record.setApiId("456");
+        int updateCount = tagRelationMapper.updateByPrimaryKeySelective(record);
+        assertThat(updateCount, greaterThan(0));
+
+        TagRelationDO tagRelationDO = tagRelationMapper.selectByPrimaryKey(record.getId());
+        assertThat(tagRelationDO.getApiId(), equalTo("456"));
+
+        int delete = tagRelationMapper.deleteByPrimaryKey(record.getId());
+        assertThat(delete, equalTo(1));
+    }
+
+    @Test
     public void testSelectByQuery() {
         TagRelationDO record = buildTagRelationDO();
         int count = tagRelationMapper.insertSelective(record);


Aias00 (approved):
## Review: #6907 — fix(admin): tag_relation updateByPrimaryKeySelective sets non-existent name column

**Verdict: ✅ Approve**

Textbook copy-paste bug fix for #6830.

### What's correct
- **The column genuinely doesn't exist.** I verified the `tag_relation` DDL (`db/init/oracle/schema.sql`): columns are `id, api_id, tag_id, date_created, date_updated`. There is no `name`. So `UPDATE tag_relation SET name = ?` always threw a `BadSqlGrammar` whenever a request carried an `apiId` — exactly the reported failure on `PUT /tag-relation/id/{id}`.
- **Fix targets the right column.** `name` → `api_id` matches the sibling `updateByPrimaryKey` in the same mapper, which already uses `api_id`. That sibling statement is the strongest evidence this was a copy-paste defect, not intent.
- **Regression test pins the defect.** `testUpdateByPrimaryKeySelective` was the only uncovered mapper method (which is *why* the bug slipped through). It inserts a row, updates `apiId`, asserts persistence, then deletes the row to avoid perturbing the count-based sibling tests. The PR description shows it failing with `BadSqlGrammar` when the mapper change is reverted — proper proof that the test is meaningful.

### Non-blocking suggestions
1. This same `name = #{apiId}` typo pattern could exist in other `*sqlmap.xml` files (the PR notes it's a copy-paste class of bug). A quick repo-wide grep for `name = #{...Id` against tables without a `name` column would be a cheap follow-up to catch siblings — but that's out of scope for this PR.
2. `buildTagRelationDO` always populates `apiId` from the DTO, so the `<if test="apiId != null">` branch fires on essentially every update. Not a problem with this fix, just noting the branch is effectively always-on.

### Verdict
Approving. One-line, verifiably-correct, test-backed. Merge as-is.

PR #6908, opened 2026-08-12
fix(admin): InstanceInfoServiceImpl.findById always returns null

### LINKED ISSUE #6618: [BUG] InstanceInfoServiceImpl.findById always returns null (stub implementation)
- severity: High
- files: `shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/InstanceInfoServiceImpl.java:91-94`
- description: The method body is literally `return null;` — it never reads from `instanceInfoMapper`. `InstanceController.detailInstanceInfo` calls this and returns the null as the response `data`.
- impact: The instance detail endpoint (`GET /instance/{id}`, gated by `system:instance:edit`) always returns null data; the feature is non-functional.
- suggested_fix: Implement `findById` to query `instanceInfoMapper.selectById(id)` and map via `buildInstanceInfoVO`.
- confidence: High
- related_existing: none.

---

yangjj-iso:
Fixes #6618

### Problem

`InstanceInfoServiceImpl.findById` was a stub:

```java
@Override
public InstanceInfoVO findById(final String id) {
    return null;
}
```

It never read from `instanceInfoMapper`. `InstanceController.detailInstanceInfo` (`GET /instance/{id}`, gated by `system:instance:edit`) passes the result straight into the response payload, so the instance detail endpoint always answered with null data regardless of whether the row existed. The feature was non-functional.

Everything needed to implement it was already present in the class: `InstanceInfoMapper.selectById(String)` and the private `buildInstanceInfoVO(InstanceInfoDO)` helper already used by `list()` and `listByPage(...)`.

### Change

- Implement `findById` via `instanceInfoMapper.selectById(id)` and the existing `buildInstanceInfoVO` mapper. A missing row still yields `null`, so the controller contract is unchanged and this stays a pure bug fix rather than a behaviour change.
- `InstanceInfoServiceTest.testFindById` previously pinned the stub in place:

```java
@Test
void testFindById() {
    // current implementation returns null
    assertEquals(null, instanceInfoService.findById("any"));
}
```

  It is replaced by a case asserting the DO is mapped onto the VO, plus a `testFindByIdNotFound` case covering the missing-row path.

### Verification

Run against JDK 17, matching the CI matrix.

With the fix applied:

```
[INFO] You have 0 Checkstyle violations.
[INFO] Tests run: 6, Failures: 0, Errors: 0, Skipped: 0 - in org.apache.shenyu.admin.service.InstanceInfoServiceTest
[INFO] BUILD SUCCESS
```

With the service change reverted to the stub and only the new tests in place, to confirm they actually pin the defect:

```
[ERROR] InstanceInfoServiceTest.testFindById:105 expected: not <null>
[ERROR] InstanceInfoServiceTest.testFindByIdNotFound » UnnecessaryStubbing
```

The second failure is Mockito strict stubbing reporting that the stub implementation never reaches the mapper at all.

Make sure that:

- [x] Your local test passed. Scope note: I ran `./mvnw -pl shenyu-admin -am -DskipTests install` followed by `./mvnw -pl shenyu-admin test -Dtest=InstanceInfoServiceTest` with Checkstyle enabled, rather than a full-reactor `clean install`. The change is confined to one method in `shenyu-admin`.

### DIFF
--- a/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/InstanceInfoServiceImpl.java
+++ b/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/InstanceInfoServiceImpl.java
@@ -90,7 +90,8 @@ public class InstanceInfoServiceImpl implements InstanceInfoService {
 
     @Override
     public InstanceInfoVO findById(final String id) {
-        return null;
+        InstanceInfoDO instanceInfoDO = instanceInfoMapper.selectById(id);
+        return Objects.isNull(instanceInfoDO) ? null : this.buildInstanceInfoVO(instanceInfoDO);
     }
 
     private List<InstanceInfoVO> buildInstanceInfoVO(final List<InstanceInfoDO> instanceInfoDOList) {
--- a/shenyu-admin/src/test/java/org/apache/shenyu/admin/service/InstanceInfoServiceTest.java
+++ b/shenyu-admin/src/test/java/org/apache/shenyu/admin/service/InstanceInfoServiceTest.java
@@ -39,6 +39,7 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.hasSize;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.verify;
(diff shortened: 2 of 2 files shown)


Aias00 (approved):
## Review: #6908 — fix(admin): InstanceInfoServiceImpl.findById always returns null

**Verdict: ✅ Approve**

Small, correct, well-justified bug fix for #6618.

### What's correct
- **Real defect, minimal fix.** `findById` was a stub returning `null`, so `GET /instance/{id}` shipped non-functional regardless of whether the row existed. The fix routes through `instanceInfoMapper.selectById(id)` and reuses the existing single-DO helper `buildInstanceInfoVO(InstanceInfoDO)` (confirmed present at base, used by `list()`/`listByPage()`). No new logic, no new mapper method.
- **Contract preserved.** A missing row still yields `null`, so `InstanceController.detailInstanceInfo` behavior is unchanged — this is a pure bug fix, not a behavior change.
- **Tests actually pin the defect.** The old `testFindById` asserted the broken `null` behavior; it's replaced with a mapping assertion (`instanceIp/port/type/namespaceId`), plus a `testFindByIdNotFound` case for the missing-row path. The PR description even shows the test *failing* when the service is reverted to the stub, which is exactly the right way to prove a regression test is meaningful.

### Non-blocking suggestions
1. Ternary `Objects.isNull(instanceInfoDO) ? null : buildInstanceInfoVO(...)` is fine; if you want consistency with the surrounding code you could use `Optional.ofNullable(instanceInfoDO).map(this::buildInstanceInfoVO).orElse(null)`, but that's purely stylistic.
2. Consider a follow-up: `detailInstanceInfo` returning HTTP 200 with a `null` body for a missing id is arguably a 404 situation, but that's outside this PR's scope and the stub contract — not blocking.

### Verdict
Approving. This is the kind of tight, test-backed fix that should merge as-is.

PR #6909, opened 2026-08-12
[ISSUE #6781] Fix stale proxy selector data after empty HTTP refresh.

### LINKED ISSUE #6781: [BUG] ProxySelectorRefresh.refresh() does not clear stale data on empty refresh (HTTP)
- Severity: Medium
- Location:
`shenyu-sync-data-center/shenyu-sync-data-http/src/main/java/org/apache/shenyu/sync/data/http/refresh/ProxySelectorRefresh.java:55-61`
- 
Description:
Same pattern as SYNC-5. When data is empty, returns without calling `proxySelectorDataSubscribers.forEach(ProxySelectorDataSubscriber::refresh)`. The `ProxySelectorDataSubscriber` interface has a `refresh()` method.
- 
Impact:
Stale proxy selector data persists after HTTP sync refresh when the group is empty.
- 
Suggested fix:
Add `proxySelectorDataSubscribers.forEach(ProxySelectorDataSubscriber::refresh)` in the empty branch.
- 
Confidence: High

---

lymerin:
Fixes #6781.

This PR clears stale proxy selector runtime data when HTTP synchronization receives an empty full snapshot.

Changes:

- propagate empty proxy selector snapshots to subscribers through `refresh()`;
- propagate refresh operations to proxy selector handlers;
- shut down and remove cached TCP bootstrap servers;
- continue clearing remaining servers when one shutdown fails;
- avoid a null-pointer race between individual deletion and full refresh;
- add unit, integration, and public HTTP synchronization path tests.

Although the issue is triggered by HTTP synchronization, `CommonProxySelectorDataSubscriber` is shared. Any synchronization mode that invokes `refresh()` will now clear the TCP proxy selector runtime cache.

Tests:

- full repository `clean install` passed;
- affected 14-module Maven reactor passed;
- Checkstyle passed with 0 violations;
- `git diff --check` passed.

Make sure that:

- [x] You have read the [[contribution guidelines](https://shenyu.apache.org/community/contributor-guide)](https://shenyu.apache.org/community/contributor-guide).

### DIFF
--- a/shenyu-plugin/shenyu-plugin-base/src/main/java/org/apache/shenyu/plugin/base/cache/CommonProxySelectorDataSubscriber.java
+++ b/shenyu-plugin/shenyu-plugin-base/src/main/java/org/apache/shenyu/plugin/base/cache/CommonProxySelectorDataSubscriber.java
@@ -51,6 +51,6 @@ public class CommonProxySelectorDataSubscriber implements ProxySelectorDataSubsc
 
     @Override
     public void refresh() {
-        ProxySelectorDataSubscriber.super.refresh();
+        handlerMap.values().forEach(ProxySelectorDataHandler::refresh);
     }
 }
--- a/shenyu-plugin/shenyu-plugin-base/src/main/java/org/apache/shenyu/plugin/base/handler/ProxySelectorDataHandler.java
+++ b/shenyu-plugin/shenyu-plugin-base/src/main/java/org/apache/shenyu/plugin/base/handler/ProxySelectorDataHandler.java
@@ -40,6 +40,12 @@ public interface ProxySelectorDataHandler {
     void removeProxySelector(String proxySelectorName);
 
     /**
+     * Refresh.
+     */
+    default void refresh() {
+    }
+
+    /**
      * pluginName.
      *
      * @return pluginName
--- a/shenyu-plugin/shenyu-plugin-base/src/test/java/org/apache/shenyu/plugin/base/cache/CommonProxySelectorDataSubscriberTest.java
+++ b/shenyu-plugin/shenyu-plugin-base/src/test/java/org/apache/shenyu/plugin/base/cache/CommonProxySelectorDataSubscriberTest.java
@@ -1,0 +1,53 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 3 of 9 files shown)


Aias00 (approved):
## Review: #6909 — [ISSUE #6781] Fix stale proxy selector data after empty HTTP refresh

**Verdict: ✅ Approve (with minor non-blocking notes)**

This is a clean, well-scoped fix for #6781. Good work.

### What's correct
- **Root cause addressed.** `CommonProxySelectorDataSubscriber.refresh()` previously just called `ProxySelectorDataSubscriber.super.refresh()`, which I verified is a **no-op** (`default void refresh() {}` in the interface). So an empty full snapshot never cleared proxy-selector runtime data. The new implementation delegates to each `ProxySelectorDataHandler.refresh()`, which actually tears down state. This is strictly an improvement with no behavior lost from the (empty) interface default.
- **`ProxySelectorDataHandler.refresh()` added as a `default` method** → existing handlers keep compiling; only the TCP handler overrides it. Good backward-compat design.
- **`TcpBootstrapFactory.clearCache()`** shuts down each cached `BootstrapServer` with a try/catch so one failing shutdown doesn't abort the rest — exactly the "continue clearing remaining servers when one fails" goal. Using the value-conditional `cache.remove(name, server)` is the safe way to mutate a `ConcurrentHashMap` during iteration (no CME, no stale removal).
- **`TcpProxySelectorDataHandler.removeProxySelector`** now does `removeCache()` + null-check before `.shutdown()`, closing a TOCTOU race where the old `inCache()`-then-`removeCache().shutdown()` could NPE if another thread removed the entry in between.
- **Propagation wired through `ProxySelectorRefresh`**: empty snapshot now calls `proxySelectorDataSubscribers.forEach(ProxySelectorDataSubscriber::refresh)`.
- **Test coverage is solid**: subscriber refresh (with/without handlers), TCP handler refresh incl. the "continues when shutdown fails" case and `removeProxySelector`, the HTTP sync data-service path, and a dedicated `ProxySelectorRefreshTest` asserting empty→clear and non-empty→subscribe.

### Non-blocking suggestions
1. **Empty-snapshot semantics are global.** Any sync mode that sends an empty `PROXY_SELECTOR` snapshot will now hard-stop *all* TCP bootstrap servers. That matches the documented intent (#6781), but it's worth a one-line comment in `ProxySelectorRefresh.refresh()` noting that an empty snapshot = "clear everything", so future readers don't treat it as a no-op/clear-local-only path.
2. **`clearCache()` mutates the map from inside `forEach`.** Current implementation is safe for `ConcurrentHashMap`, but if `shutdown()` ever triggers re-registration (a lifecycle listener), a server could be missed. Not an issue today — just flagging the invariant to preserve.

### Verdict
Approving. The fix is correct, the NPE race is real, and the regression tests back the behavior. Nice job.

Aias00:
thx for contribution, pls add my wechat: aias00

PR #6912, opened 2026-08-13
Fix #6820 Fix missing Dubbo rpcExt in ApiDoc registration

### LINKED ISSUE #6820: [BUG] Property key case mismatch: "rpcExt" written vs "RpcExt" read — Dubbo rpcExt silently lost in ApiDoc ext JSON
- Severity: Medium
- Location:
`shenyu-client/shenyu-client-dubbo/shenyu-client-apache-dubbo/src/main/java/org/apache/shenyu/client/apache/dubbo/processor/extractor/DubboServiceProcessor.java:52,57` (writes `"rpcExt"`); `shenyu-client/shenyu-client-core/src/main/java/org/apache/shenyu/client/core/register/registrar/ApiDocRegistrarImpl.java:113` (reads `"RpcExt"`); `shenyu-client/shenyu-client-core/src/main/java/org/apache/shenyu/client/core/register/registrar/AbstractApiDocRegistrar.java:140` (reads `"RpcExt"`)
- 
Description:
`DubboServiceProcessor` writes the rpcExt property with key `"rpcExt"` (lowercase r). Both ApiDoc registrars read it with key `"RpcExt"` (uppercase R). `java.util.Properties` (extends `Hashtable`) is case-sensitive, so `getPropertiesValue("RpcExt")` returns null when the key stored is `"rpcExt"`. The old path reads rpcExt from the `MetaDataRegisterDTO` object directly (`metaData.getRpcExt()`), so it works. The new-path registrars read from the Properties map where the case doesn't match.
- 
Impact:
In the new registrar path, the ApiDoc ext JSON for Dubbo services has `rpcExt: null` instead of group/version/loadbalance/retries/timeout/cluster/serialization data.
- 
Suggested fix:
Standardize on `"rpcExt"` (lowercase) in both writer and readers.
- 
Confidence: High
- Related existing: none — #5943 (dubbo method config) is different.

---

yykaue:
Fixes #6820.

## Summary

`DubboServiceProcessor` stores the Dubbo RPC extension using the lowercase `rpcExt` property key, while both ApiDoc registrars read it using `RpcExt`.

Because `Properties` keys are case-sensitive, the generated ApiDoc extension JSON contained a null `rpcExt` value.

This PR:

- Standardizes both ApiDoc registrars to read the lowercase `rpcExt` key.
- Adds regression tests for both registration paths.
- Verifies that Dubbo RPC extension data is preserved in the generated ApiDoc extension JSON.

### DIFF
--- a/shenyu-client/shenyu-client-core/src/main/java/org/apache/shenyu/client/core/register/registrar/AbstractApiDocRegistrar.java
+++ b/shenyu-client/shenyu-client-core/src/main/java/org/apache/shenyu/client/core/register/registrar/AbstractApiDocRegistrar.java
@@ -137,7 +137,7 @@ public abstract class AbstractApiDocRegistrar extends AbstractApiRegistrar<ApiDo
         ext.setServiceName(apiDefinition.getApiBean().getBeanClass().getName());
         ext.setMethodName(apiDefinition.getApiMethodName());
         ext.setParameterTypes(apiDefinition.getParameterTypes());
-        ext.setRpcExt(apiDefinition.getPropertiesValue("RpcExt"));
+        ext.setRpcExt(apiDefinition.getPropertiesValue("rpcExt"));
         ext.setAddPrefixed(addPrefixed);
         
         if (rpcTypeEnum == RpcTypeEnum.HTTP) {
--- a/shenyu-client/shenyu-client-core/src/main/java/org/apache/shenyu/client/core/register/registrar/ApiDocRegistrarImpl.java
+++ b/shenyu-client/shenyu-client-core/src/main/java/org/apache/shenyu/client/core/register/registrar/ApiDocRegistrarImpl.java
@@ -110,7 +110,7 @@ public class ApiDocRegistrarImpl extends BaseApiRegistrarImpl {
         apiExt.setServiceName(api.getApiBean().getBeanClass().getName());
         apiExt.setMethodName(api.getApiMethodName());
         apiExt.setParameterTypes(api.getParameterTypes());
-        apiExt.setRpcExt(api.getPropertiesValue("RpcExt"));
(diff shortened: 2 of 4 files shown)


Aias00:
Optional, non-blocking suggestions
Consider extracting "rpcExt" into a shared constant (e.g. in org.apache.shenyu.common.constant.Constants) used by both DubboServiceProcessor and the two registrars, so a writer/reader key mismatch like this can't silently reappear.
Tiny nit: the @SuppressWarnings("unchecked") on the new invokeGetExt helper is unnecessary (there is no unchecked cast) — it appears to be copied from invokeGetDocument.

Aias00 (approved):
Fixes #6820. I verified the root cause against `DubboServiceProcessor`: it stores the extension via `apiBean.addProperties("rpcExt", ...)` / `definition.addProperties("rpcExt", ...)` (lowercase key), but both registrars were reading `getPropertiesValue("RpcExt")` (capital R). Since `Properties` keys are case-sensitive, the value came back `null`. Aligning the readers to the lowercase `rpcExt` key is the correct fix, and a code search confirms these two registrars are the only readers of that property key, so the change is complete.

The regression tests cover both registration paths:
- `ApiDocRegistrarImplTest.testGetExtWithRpcExt` exercises `ApiDocRegistrarImpl.getExt`.
- `NoHttpApiDocRegistrarTest.testRpcExtInExtJson` exercises `AbstractApiDocRegistrar.buildExtJson` via `register(...)`.

Both assert the generated ext JSON preserves the Dubbo `rpcExt` value, and would fail if the key case didn't match.

One optional note (non-blocking): only `DubboServiceProcessor` populates the `rpcExt` property key; other RPC types (gRPC/SOFA/TARS/etc.) don't set it, so `ext.rpcExt` stays `null` for them. That's pre-existing behavior and out of scope for #6820 — just flagging for awareness.

No issues — approving.

PR #6914, opened 2026-08-14
fix: invalidate rule match cache when deleting a selector (#6797)

### LINKED ISSUE #6797: [BUG] — Selector DELETE fails to invalidate the RULE match cache → stale rule reads
- severity: Medium
- files: `shenyu-plugin-base/src/main/java/org/apache/shenyu/plugin/base/cache/CommonPluginDataSubscriber.java:287-296` (`removeCacheData` SelectorData branch) vs `:232-245` (`updateCacheData` SelectorData branch)
- description: On selector *update* the code invalidates both caches (incl. `removeRuleDataBySelector` + `removeEmptyRuleData` at lines 243-244). On selector *delete* (lines 292-296) only the SELECTOR match cache is invalidated; the RULE match cache is NOT. `MatchDataCache` stores `path -> RuleData` whose `selectorId` references the now-deleted selector. `AbstractShenyuPlugin.execute` then calls `doExecute(exchange, chain, selectorData, ruleData)` with a rule tied to a dead selector until (and if) the corresponding rule-delete events arrive separately. If rules are orphaned the staleness is permanent.
- impact: Routing to a deleted selector's rule; window of inconsistency under normal cascade, permanent if orphaned.
- suggested_fix: In `removeCacheData` SelectorData branch, add the same two calls present in `updateCacheData`: `MatchDataCache.getInstance().removeRuleDataBySelector(...)` and `removeEmptyRuleData(...)`.
- confidence: High

---

yykaue:
Fixes #6797 

## What this PR does
@Aias00 

Invalidate the rule match cache when a selector is deleted.

Previously, selector deletion only removed entries from the selector match cache. Cached `RuleData` associated with the deleted selector could remain available, potentially causing requests to be routed using rules belonging to a deleted selector.

This PR updates `CommonPluginDataSubscriber` to:

- Remove cached rule data associated with the deleted selector.
- Remove empty rule cache entries.
- Keep rule cache entries belonging to unrelated selectors.

## Tests

- Added a regression test covering selector deletion and rule match cache invalidation.
- Verified that rules associated with the deleted selector are removed.
- Verified that unrelated rule cache entries remain available.
- Passed:

  ```text
  ./mvnw -pl shenyu-plugin/shenyu-plugin-base test
  Tests run: 113, Failures: 0, Errors: 0, Skipped: 0

### DIFF
--- a/shenyu-plugin/shenyu-plugin-base/src/main/java/org/apache/shenyu/plugin/base/cache/CommonPluginDataSubscriber.java
+++ b/shenyu-plugin/shenyu-plugin-base/src/main/java/org/apache/shenyu/plugin/base/cache/CommonPluginDataSubscriber.java
@@ -289,10 +289,14 @@ public class CommonPluginDataSubscriber implements PluginDataSubscriber {
             BaseDataCache.getInstance().removeSelectData(selectorData);
             Optional.ofNullable(handlerMap.get(selectorData.getPluginName()))
                     .ifPresent(handler -> handler.removeSelector(selectorData));
-            // remove selector match cache
+            // remove match cache
             if (selectorMatchConfig.getCache().getEnabled()) {
                 MatchDataCache.getInstance().removeSelectorData(selectorData.getPluginName(), selectorData.getId());
                 MatchDataCache.getInstance().removeEmptySelectorData(selectorData.getPluginName());
+            }
+            if (ruleMatchCacheConfig.getCache().getEnabled()) {
+                MatchDataCache.getInstance().removeRuleDataBySelector(selectorData.getPluginName(), selectorData.getId());
+                MatchDataCache.getInstance().removeEmptyRuleData(selectorData.getPluginName());
             }
         } else if (data instanceof RuleData) {
             RuleData ruleData = (RuleData) data;
--- a/shenyu-plugin/shenyu-plugin-base/src/test/java/org/apache/shenyu/plugin/base/cache/CommonPluginDataSubscriberTest.java
+++ b/shenyu-plugin/shenyu-plugin-base/src/test/java/org/apache/shenyu/plugin/base/cache/CommonPluginDataSubscriberTest.java
@@ -140,13 +140,40 @@ public final class CommonPluginDataSubscriberTest {
 
     @Test
     public void testUnSelectorSubscribe() {
+        final String path = "/selector";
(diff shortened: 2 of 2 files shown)


Aias00:
Two optional, non-blocking suggestions
Test coverage of the empty-rule branch: in the test, emptyRuleData is built with .selectorId(mockSelectorId1), so it is actually removed by removeRuleDataBySelector before removeEmptyRuleData ever applies — the removeEmptyRuleData path is not independently exercised. The real empty sentinel cached by AbstractShenyuPlugin.defaultMatchRule has no selectorId:

RuleData emptyRuleData = RuleData.builder().pluginName(named()).build();
Consider dropping .selectorId(...) from emptyRuleData (or using an unrelated selector id) so the empty-entry cleanup is covered precisely.
Side observation (out of scope for this PR): since the trie cache removal in #6303, positive L1 match-cache hits are no longer used by AbstractShenyuPlugin.execute — a non-empty cached selector/rule is fetched but then immediately overwritten by defaultMatchSelector/defaultMatchRule; only the empty (null-id) sentinel short-circuits. So on current master the stale-routing window from this bug is limited, and this fix is mainly about memory hygiene, consistency with the update path, and future-proofing if the positive L1 short-circuit is ever restored. That unused positive L1 cache may be worth a separate issue.

Aias00 (approved):
Fixes #6797. LGTM.

- The new `if (ruleMatchCacheConfig.getCache().getEnabled())` block in `removeCacheData` is symmetric with the existing `selectorMatchConfig` block and with the rule-deletion path used elsewhere in the same class (around lines 242 and 251), so it follows the established pattern and is easy to reason about.
- Verified `ruleMatchCacheConfig` is a `private final RuleMatchCache` field injected via the constructor (both constructors assign it), so there is no NPE risk in production when a selector is unsubscribed.
- The regression test `testUnSelectorSubscribe` is thorough: it asserts the rule-cache entry for the deleted selector is removed (`removeRuleDataBySelector`), the empty-rule entry is removed (`removeEmptyRuleData`), and an unrelated selector's rule-cache entry is preserved. It also cleans up the shared `MatchDataCache` singleton in a `finally` block, avoiding cross-test contamination.
- The follow-up commit (removing `selectorId` from the `emptyRuleData` builder) correctly turns it into a genuine "empty" rule so `removeEmptyRuleData` actually targets it.

No issues — approving.

PR #6915, opened 2026-08-14
fix: close gzip streams on decompression failure (#6811)

### LINKED ISSUE #6811: [BUG] — `AiResponseTransformerPlugin`: `GZIPInputStream`/`ByteArrayOutputStream` not closed on `IOException` → native Inflater leak
- severity: Low-Medium; files: `shenyu-plugin-ai/shenyu-plugin-ai-response-transformer/.../AiResponseTransformerPlugin.java:312-328`
- description: `new GZIPInputStream(bis)` / `new ByteArrayOutputStream()` are closed only at lines 321-322 after the read loop; the `catch (Exception e)` at line 325 does not close them. `GZIPInputStream` owns a native `Inflater` (zlib); a truncated/corrupt gzip body throws and leaks the native decompression buffer until GC finalization. Sibling `LoggingConsolePlugin.decompressGzip` (`:306-316`) does this correctly with try-with-resources.
- impact: Native (zlib Inflater) memory leak per failed gzip decompression.
- suggested_fix: Wrap `gis`/`bos` in a single try-with-resources.
- confidence: High
- related_existing: SEC-9 covers unbounded decompression DoS (different defect — this is the not-closed-on-error leak).

---

yykaue:
Fixes #6811

Thanks @Aias00 for reporting this issue.

## What this PR does

Use try-with-resources to manage `GZIPInputStream` and `ByteArrayOutputStream` in `AiResponseTransformerPlugin`.

Previously, these streams were closed only after successful decompression. If `GZIPInputStream.read()` threw an exception while processing a truncated or corrupted gzip response, the stream was not closed and its native `Inflater` resources could remain allocated until garbage collection.

With this change, both streams are closed automatically on successful and exceptional paths.

## Tests

- Added a regression test that makes `GZIPInputStream.read()` throw an `IOException`.
- Verified that `GZIPInputStream.close()` is called after decompression fails.
- Passed the response transformer module tests:

  ```text
  ./mvnw -pl shenyu-plugin/shenyu-plugin-ai/shenyu-plugin-ai-response-transformer test
  Tests run: 47, Failures: 0, Errors: 0, Skipped: 0

### DIFF
--- a/shenyu-plugin/shenyu-plugin-ai/shenyu-plugin-ai-response-transformer/src/main/java/org/apache/shenyu/plugin/ai/transformer/response/AiResponseTransformerPlugin.java
+++ b/shenyu-plugin/shenyu-plugin-ai/shenyu-plugin-ai-response-transformer/src/main/java/org/apache/shenyu/plugin/ai/transformer/response/AiResponseTransformerPlugin.java
@@ -48,6 +48,8 @@ import reactor.core.publisher.Mono;
 import org.reactivestreams.Publisher;
 
 import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.StringReader;
 import java.nio.charset.StandardCharsets;
@@ -57,6 +59,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import java.util.zip.GZIPInputStream;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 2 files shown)


Aias00:
Optional, non-blocking nits
Since the project targets Java 17, bos.toString(StandardCharsets.UTF_8.name()) (throws checked UnsupportedEncodingException) could be simplified to bos.toString(StandardCharsets.UTF_8) — pre-existing code, fine to leave.
ByteArrayOutputStream.close() is a no-op, so strictly only gis needs managing — keeping both in the try-with-resources is harmless and arguably clearer, so no change needed.

Aias00 (approved):
Fixes #6811. This is a clean, idiomatic fix.

- Switching to try-with-resources guarantees both `GZIPInputStream` (and its underlying `Inflater`) and `ByteArrayOutputStream` are closed on both the success and exception paths. Previously, if `gis.read()` threw on a truncated/corrupt gzip payload, `gis.close()` was never reached, leaking the native inflater resources until GC.
- The regression test (`testGzipInputStreamClosedWhenDecompressionFails`) correctly uses `MockedConstruction<GZIPInputStream>` to force `read()` to throw and asserts `close()` is invoked. Note that `mockConstruction` skips the real `GZIPInputStream` constructor (which would otherwise read the gzip header), so the mock path faithfully reproduces the intended control flow.
- Imports are cleaned up (no more fully-qualified `java.io.ByteArrayInputStream` in the body) and all added imports are used.

One minor nit (non-blocking): the inline `ByteArrayInputStream` is not managed by try-with-resources, but it holds no native resources, so this is fine.

Looks good — approving.

PR #6917, opened 2026-08-14
Fix 6816 sign test delay

### LINKED ISSUE #6816: [BUG] — Sign "success" tests run with `delay=0` (constructed via `new`, `@Value` unprocessed) → minute-boundary flake
- severity: Low-Medium; files: `shenyu-plugin-sign/.../service/SignServiceVersionOneTest.java:121,222,242,302,320` (`String.valueOf(System.currentTimeMillis())` expecting `VerifyResult.success()`), `SignServiceVersionTwoTest.java:123`; construction `SignServiceVersionOneTest.java:84` (`new ComposableSignService(new DefaultExtractor(), new DefaultSignProvider())`); field `ComposableSignService.java:80-81` (`@Value("${shenyu.sign.delay:5}") private int delay;` → JVM default `0` when constructed outside Spring)
- description: `ComposableSignService` is constructed via `new` in the tests, so Spring's `AutowiredAnnotationBeanPostProcessor` never processes `@Value` and `delay` stays `0`. The expiry check is `Math.abs(acquireMinutesBetween(start, now)) <= delay` i.e. `<= 0`. A timestamp captured at `MM:59.999` verified at `MM+1:00.001` yields `between=1` → `abs(1) <= 0` is false → unexpected `SIGN_TIME_IS_TIMEOUT`.
- impact: Intermittent failures of all "success-expecting" sign tests on slow/contended CI near minute boundaries.
- suggested_fix: Inject a real `delay` (reflection-set to 5, or construct via Spring test context so `@Value` resolves).
- confidence: High

---

yykaue:
Fixes #6816

## What this PR does

Explicitly initialize the sign verification delay in `SignServiceVersionOneTest` and `SignServiceVersionTwoTest`.

Previously, both test classes constructed `ComposableSignService` directly with `new`. As a result, Spring did not process its `@Value` field, leaving `delay` at the JVM default value of `0`.

This could make success-expecting tests fail intermittently when timestamp creation and verification crossed a minute boundary.

This PR:

- Uses an explicit `DELAY = 5` test constant.
- Injects the delay into manually constructed `ComposableSignService` instances with `ReflectionTestUtils`.
- Removes the ineffective `@Value` fields from the test classes.
- Adds regression tests verifying that valid signatures from one minute earlier succeed.
- Updates overdue tests to use the same configured delay.

## Tests

```text
./mvnw -pl shenyu-plugin/shenyu-plugin-security/shenyu-plugin-sign test
Tests run: 41, Failures: 0, Errors: 0, Skipped: 0

### DIFF
--- a/shenyu-plugin/shenyu-plugin-security/shenyu-plugin-sign/src/test/java/org/apache/shenyu/plugin/sign/service/SignServiceVersionOneTest.java
+++ b/shenyu-plugin/shenyu-plugin-security/shenyu-plugin-sign/src/test/java/org/apache/shenyu/plugin/sign/service/SignServiceVersionOneTest.java
@@ -41,10 +41,10 @@ import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.junit.jupiter.MockitoExtension;
 import org.mockito.junit.jupiter.MockitoSettings;
 import org.mockito.quality.Strictness;
-import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.mock.http.server.reactive.MockServerHttpRequest;
 import org.springframework.mock.web.server.MockServerWebExchange;
+import org.springframework.test.util.ReflectionTestUtils;
 import org.springframework.web.server.ServerWebExchange;
 
 import java.util.Collections;
@@ -66,6 +66,8 @@ import static org.mockito.Mockito.mock;
 @MockitoSettings(strictness = Strictness.LENIENT)
 public final class SignServiceVersionOneTest {
 
+    private static final int DELAY = 5;
+
     private SignService signService;
 
     private ServerWebExchange exchange;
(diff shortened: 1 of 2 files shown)


Aias00 (approved):
## Review: Fix #6816 sign test delay

**Verdict: ✅ Approve**

### What it fixes
`SignServiceVersionOneTest` and `SignServiceVersionTwoTest` built `ComposableSignService` with `new`, so Spring never processed the service's `@Value("${shenyu.sign.delay:5}")` — `delay` stayed at the JVM default `0`. The tests' own `@Value private int delay` field was dead code (never propagated to the service), so the success-path verification window was effectively zero tolerance. When sign-timestamp creation and `signatureVerify` crossed a minute boundary, a valid signature was rejected as overdue → intermittent failures.

### Why the fix is correct
- `ComposableSignService.delay` is confirmed to exist (field `private int delay`, line 81, `@Value("${shenyu.sign.delay:5}")`); `ReflectionTestUtils.setField(signService, "delay", DELAY)` correctly injects `5` minutes, matching the intended production default.
- Verification uses `Math.abs(between) <= delay` (ms vs `delay*60_000`), so a 5-minute window now actually applies in tests.
- Removes the two ineffective `@Value` fields from the test classes (they were never wired in) — clean.

### Test validity (verified)
`normalTestWithinConfiguredDelay` signs with `timestamp = now - 60_000ms` (1 min earlier) and expects `success()`. Under `delay=0` this would be overdue and fail; under `delay=5` it passes — so the test genuinely guards the intended behavior rather than just passing by accident. `overdueTest` now uses `DELAY + 1` minutes and the matching error-message format. PR reports `Tests run: 41, Failures: 0`.

### Conclusion
Correct, well-targeted test-stability fix with a meaningful regression test. Approved.

(Reviewed as part of a descending re-pass of open PRs; this PR is new since the last review cycle.)

PR #6919, opened 2026-08-14
[ISSUE #6767] fix(logging): handle missing ClickHouse TTL configuration

### LINKED ISSUE #6767: [BUG] ClickHouseLogCollectClient.initClient0 NPE when ttl config is null
- severity: Medium
- files: `shenyu-plugin/shenyu-plugin-logging/shenyu-plugin-logging-clickhouse/.../ClickHouseLogCollectClient.java:128`
- description: `final String ttl = config.getTtl().isEmpty() ? "30" : config.getTtl();` — `ttl` field has no default. If admin config JSON omits "ttl", Gson sets it to null. `config.getTtl()` returns null, `.isEmpty()` throws NPE.
- impact: ClickHouse logging plugin fails to initialize with NPE when `ttl` not configured.
- suggested_fix: Use `StringUtils.isBlank(config.getTtl())` or `Optional.ofNullable(...).orElse("30")`.
- confidence: High
- related_existing: none

---

yykaue:
Fixes #6767 

`ClickHouseLogCollectClient.initClient0()` previously called `isEmpty()` directly on the configured TTL value. When the `ttl` field was omitted from the admin configuration, Gson deserialized it as `null`, causing the ClickHouse logging plugin to fail during initialization with a `NullPointerException`.

This PR:

- Uses `StringUtils.defaultIfBlank()` to default missing, empty, or blank TTL values to `30`.
- Preserves explicitly configured TTL values.
- Adds a regression test covering configuration JSON without the `ttl` field.

## Verification

- The regression test reproduces the original exception before the fix.
- The targeted test passes after the fix.
- Relevant tests and Checkstyle pass.

@Aias00 Could you please review this PR?

### DIFF
--- a/shenyu-plugin/shenyu-plugin-logging/shenyu-plugin-logging-clickhouse/src/main/java/org/apache/shenyu/plugin/logging/clickhouse/client/ClickHouseLogCollectClient.java
+++ b/shenyu-plugin/shenyu-plugin-logging/shenyu-plugin-logging-clickhouse/src/main/java/org/apache/shenyu/plugin/logging/clickhouse/client/ClickHouseLogCollectClient.java
@@ -29,6 +29,7 @@ import com.clickhouse.client.data.ClickHouseLongValue;
 import com.clickhouse.client.data.ClickHouseOffsetDateTimeValue;
 import com.clickhouse.client.data.ClickHouseStringValue;
 import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.shenyu.common.utils.DateUtils;
 import org.apache.shenyu.plugin.logging.clickhouse.config.ClickHouseLogCollectConfig;
 import org.apache.shenyu.plugin.logging.clickhouse.constant.ClickHouseLoggingConstant;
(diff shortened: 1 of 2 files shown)


Aias00 (approved):
## Review: [ISSUE #6767] fix(logging): handle missing ClickHouse TTL configuration

**Verdict: ✅ Approve**

### What it fixes
`ClickHouseLogCollectClient.initClient0()` previously did `config.getTtl().isEmpty() ? "30" : config.getTtl()`. When the `ttl` field is omitted from the admin config, Gson deserializes it as `null` (the field is a `String`), so `.isEmpty()` throws `NullPointerException` and the ClickHouse logging plugin fails to initialize.

### Why the fix is correct
- `StringUtils.defaultIfBlank(config.getTtl(), "30")` (commons-lang3, already a project dependency) handles `null`, empty, **and** blank — a strict superset of the old `isEmpty()` check, and preserves explicitly configured TTL values.
- Semantics preserved: missing/blank → `"30"`; otherwise the configured value.

### Test validity (verified)
I checked the head source: `initClient0` eagerly connects to ClickHouse, but the connection block is wrapped in `try { ... } catch (Exception e) { LOG.error(...) }`, so connection failures are swallowed. The original NPE fired **before** that try (at the `.isEmpty()` call), so `testInitClientWithoutTtl` genuinely reproduces and guards the NPE path: it asserts `config.getTtl() == null` and that `initClient0` no longer throws. The regression test is meaningful.

### Non-blocking suggestion
The broad `catch (Exception e)` around client init silently swallows connection errors (only `LOG.error`). That's pre-existing and out of scope here, but it can mask plugin-init failures in production — worth tightening separately (e.g., rethrow/log a clear init error) so misconfiguration is visible rather than silent.

### Conclusion
Minimal, correct, well-tested NPE fix. Approved.

(Reviewed as part of a descending re-pass of open PRs; this PR is new since the last review cycle.)

PR #6922, opened 2026-08-16
fix: honor retain flag in mqtt publish message

### LINKED ISSUE #6849: [BUG] Retain flag ignored — every PUBLISH stored as retained; zero-byte retained PUBLISH never deletes
## Description
`TopicRepository.add(topic, message)` is called unconditionally for every inbound PUBLISH regardless of the retain bit (module-wide grep for `isRetain`/`retain` returns zero hits). A zero-byte retained PUBLISH should delete the retained message — that never happens.

## Location
- `shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/Publish.java:55`
- `shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/repositories/TopicRepository.java:34-37`

## Impact
Non-retained messages are persisted and redelivered to future subscribers as if they were retained (via `Subscribe.sendSubMessage` at `Subscribe.java:72-77`, which sends with `retain=true`). Violates MQTT-3.3.1.

## Suggested fix
Read `msg.fixedHeader().isRetain()`; only call `TopicRepository.add` when retain is true; on a zero-byte retained PUBLISH, call `TopicRepository.remove(topic)`.

wy471x:
Only store the message in TopicRepository when the RETAIN flag is set, and remove the retained message when a zero-byte retained PUBLISH is received (MQTT-3.3.1).

Make sure that:

## Summary
Changes:
- Publish.java: publish() now checks msg.fixedHeader().isRetain(). Retained messages with payload are stored via TopicRepository.add; a zero-byte retained PUBLISH calls           
  TopicRepository.remove(topic) to delete the retained message; non-retained messages no longer touch the store (MQTT-3.3.1 compliance). Subscriber fan-out is unchanged.

Test Cases:
- PublishTest.java (new): 3 tests covering retained storage, non-retained no-op, and zero-byte deletion.

close [#6849](https://github.com/apache/shenyu/issues/6849)

### DIFF
--- a/shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/Publish.java
+++ b/shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/Publish.java
@@ -52,7 +52,13 @@ public class Publish extends MessageType {
         String message = byteBufToString(payload);
         //// todo qos
         MqttQoS mqttQoS = msg.fixedHeader().qosLevel();
-        Singleton.INST.get(TopicRepository.class).add(topic, message);
+        if (msg.fixedHeader().isRetain()) {
+            if (payload.isReadable()) {
+                Singleton.INST.get(TopicRepository.class).add(topic, message);
+            } else {
+                Singleton.INST.get(TopicRepository.class).remove(topic);
+            }
+        }
         int packetId = msg.variableHeader().packetId();
         CompletableFuture.runAsync(() -> send(topic, payload, packetId));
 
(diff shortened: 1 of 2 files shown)


Aias00 (approved):
## Review: #6922 honor retain flag in MQTT publish

**Decision: APPROVE**

### What changed
- `Publish.publish(...)`: previously every message was unconditionally stored in `TopicRepository` via `add(topic, message)`. Now the RETAIN flag is honored:
  - `isRetain() == true` and payload readable (non-empty) → `TopicRepository.add(...)` (store retained message).
  - `isRetain() == true` and payload not readable (zero-byte) → `TopicRepository.remove(topic)` (clear retained message per MQTT-3.3.1).
  - `isRetain() == false` → no store/clear (non-retained publishes do not touch the retained store).
- Subscriber fan-out (`send(...)`) is unchanged — still invoked for all publishes.
- Added `PublishTest` (3 cases): retained storage, non-retained no-op, zero-byte retained clears the retained message.

### Verification
- Concern about `payload.isReadable()` being invalidated by the earlier `byteBufToString(payload)` call: verified `byteBufToString` uses `byteBuf.getBytes(readerIndex, bytes)` / `new String(byteBuf.array(), ...)` — both read-only, they do NOT advance `readerIndex`. So `isReadable()` after `byteBufToString` is still correct.
- Empty-payload retained PUBLISH correctly clears the retained entry; non-retained empty payloads leave the store untouched (the block is skipped).
- Semantics match MQTT-3.3.1 (zero-length retained payload deletes the retained message).

### Notes
- Clean, spec-compliant. Approving.

PR #6923, opened 2026-08-16
fix: grant requested QoS in SUBACK instead of hard-coded QoS 0

### LINKED ISSUE #6848: [BUG] SUBACK always grants QoS 0, ignoring the requested QoS
## Description
The SUBACK grant list is built by hard-coding `MqttQoS.AT_MOST_ONCE.value()` for every topic, never reading `topicSub.qualityOfService()`.

## Location
- `shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/Subscribe.java:90-94` (line 93 hard-codes `AT_MOST_ONCE.value()`)

## Impact
A client subscribing at QoS 1 or 2 is told it was granted QoS 0 and (see H-27) will only ever receive QoS 0 delivery. Violates MQTT-3.8.4/3.9.

## Suggested fix
`qos.add(topicSub.qualityOfService().value())` (after the FAILURE filter).

wy471x:
Make sure that:

## Summary

Changes:
- Subscribe.java: sendSubAckMessage now takes the filtered List<MqttTopicSubscription> and builds the SUBACK grant list from topicSub.qualityOfService().value() instead of        
  hard-coding AT_MOST_ONCE for every topic (MQTT-3.8.4/3.9). FAILURE subscriptions remain excluded from the ack. Removed the now-unused MqttQoS and ArrayList imports.

Test Cases:
  - SubscribeTest.java (new): 3 tests — grants requested QoS (2/0/1), excludes FAILURE subscriptions, closes channel when already connected.

close [#6848](https://github.com/apache/shenyu/issues/6848)

### DIFF
--- a/shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/Subscribe.java
+++ b/shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/Subscribe.java
@@ -21,7 +21,6 @@ import io.netty.buffer.Unpooled;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.codec.mqtt.MqttFixedHeader;
-import io.netty.handler.codec.mqtt.MqttQoS;
 import io.netty.handler.codec.mqtt.MqttSubscribeMessage;
 import io.netty.handler.codec.mqtt.MqttTopicSubscription;
 import io.netty.handler.codec.mqtt.MqttMessageType;
@@ -35,7 +34,6 @@ import org.apache.shenyu.common.utils.Singleton;
 import org.apache.shenyu.protocol.mqtt.repositories.SubscribeRepository;
 import org.apache.shenyu.protocol.mqtt.repositories.TopicRepository;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 
@@ -61,9 +59,12 @@ public class Subscribe extends MessageType {
         int packetId = msg.variableHeader().messageId();
 
         //// todo Regular match
-        List<String> ackTopics = mqttTopicSubscriptions
+        List<MqttTopicSubscription> ackSubscriptions = mqttTopicSubscriptions
                 .stream()
                 .filter(topicSub -> topicSub.qualityOfService() != FAILURE)
+                .collect(Collectors.toList());
+
+        List<String> ackTopics = ackSubscriptions.stream()
                 .map(MqttTopicSubscription::topicName)
                 .collect(Collectors.toList());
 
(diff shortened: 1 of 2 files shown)


Aias00 (approved):
## Review: #6923 honor requested QoS in SUBACK

**Decision: APPROVE**

### What changed
- `Subscribe.sendSubAckMessage(...)`: previously the SUBACK always granted QoS 0 regardless of what the client requested. Now it grants each subscription the QoS the client requested (`topicSub.qualityOfService().value()`), which is the correct MQTT behavior.
- The FAILURE-filtered subscription list is now collected into `ackSubscriptions` (kept before the topic-name mapping) and passed through to `sendSubAckMessage`.
- Removed now-unused imports (`java.util.ArrayList`, `io.netty.handler.codec.mqtt.MqttQoS`).
- Added `SubscribeTest` (Mockito + ArgumentCaptor) covering: granted QoS matches requested (2/0/1), FAILURE subscriptions are excluded from SUBACK and not looked up in `topicRepository`, and connected-state closes the channel without writing a SUBACK.

### Verification
- Granted QoS list length matches the filtered subscription count; FAILURE entries are dropped before SUBACK construction, consistent with prior behavior.
- Honoring the requested QoS (server may grant equal-or-lower; here it grants exactly requested) is within MQTT spec and strictly better than always-0.
- Tests assert `grantedQoSLevels()` ordering and exclusion, validating the logic.

### Notes
- None blocking. Behavior improvement, well-tested.

Approving.

dengliming (changes requested):
This test is still calling the old `setConnected` signature. `MessageType#setConnected` now requires the `Channel` plus the boolean flag, so this won’t compile in CI.

Suggested fix:

```java
subscribe.setConnected(channel, true);

**Comment by Aias00** [comment]
CI note (PMC Aias00): the only failing check is CodeQL `Analyze (java)`. Please review whether this is a genuine security finding introduced by the change or CodeQL noise, and address/confirm it before merge. The rest of the suite is green.

**Review by Aias00 — APPROVED** [review]
Correct fix — SUBACK was hard-coding QoS 0 regardless of what the subscriber asked for.

```java
// before
for (int i = 0; i < ackTopics.size(); i++) {
    qos.add(MqttQoS.AT_MOST_ONCE.value());   // always 0
}
```

MQTT-3.8.4 requires the SUBACK payload to carry, for each subscription in order, the QoS level the server actually granted. Reporting 0 unconditionally tells a client that requested QoS 1 or 2 that it will get at-most-once delivery, so well-behaved clients downgrade their delivery guarantees (and stop sending PUBACK/PUBREC) even though the broker would have honoured the higher level. Carrying the requested `qualityOfService().value()` through is the right behaviour.

Also good: the filtered `ackSubscriptions` list is now what drives both the repository registration and the SUBACK, so the granted levels line up with the topics actually registered instead of being derived from a parallel list.

Tests are solid: `testSubAckGrantsRequestedQoS` asserts `[2, 0, 1]` for subscriptions requested in the order QoS2/QoS0/QoS1 — which also locks in that ordering is preserved, not just the set of values.

Two non-blocking observations:

1. **Failed subscriptions.** MQTT-3.8.4-5 says a subscription that cannot be granted should get the failure code `0x80` in the SUBACK, keeping the response positionally aligned with the request. This PR filters `FAILURE` entries out of the list instead, so the SUBACK ends up shorter than the request. Clients that match by index will mis-align. Returning `0x80` for those entries would be spec-correct; `testSubAckExcludesFailureSubscriptions` currently documents the filtering behaviour, so changing it means updating that test.

2. `subscribe.setConnected(true)` is used in `testSubscribeWhenConnectedClosesChannel` — if that setter is only intended for tests, a brief comment would help; if it's production API, it's fine as-is.

PR #6924, opened 2026-08-16
fix: handle null password in MqttContext.isValid to avoid NPE

### LINKED ISSUE #6847: [BUG] MqttContext.isValid NPE when client omits password (anonymous/username-only CONNECT)
## Description
`msg.payload().passwordInBytes()` returns null for any CONNECT whose connect-flags password bit is 0 (valid MQTT client — anonymous or username-only). It is passed into `MqttContext.isValid`, which calls `new String((byte[]) null)` → NPE.

## Location
- `shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/Connect.java:60-62`
- `shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/MqttContext.java:48`

## Impact
Handler thread throws; client never receives a CONNACK and the connection hangs.

## Suggested fix
Null-check `passwordInBytes` in `MqttContext.isValid` (treat null as empty).

## Related existing
Distinct from MQTT-09 (#6748, charset decoding of a non-null password) — this is a null-deref on a missing password.

wy471x:
CONNECT with password flag 0 yields a null passwordInBytes, which caused new String((byte[]) null) to throw and hang the connection without a CONNACK. Treat null as empty so the client receives CONNECTION_REFUSED_BAD_USER_NAME_OR_PASSWORD, and add unit tests.

Make sure that:

### Changes:                                                                                                                                                                               
  - Fix NPE in MqttContext.isValid: new String(passwordInBytes) throws when msg.payload().passwordInBytes() is null — any CONNECT with the password flag bit set to 0 (anonymous or  
  username-only clients). The handler thread dies and the client never receives a CONNACK, leaving the connection hanging.                                                           
  - Null passwordInBytes is now treated as empty, so validation fails gracefully and the client receives CONNECTION_REFUSED_BAD_USER_NAME_OR_PASSWORD instead of a hang.             

### Test Cases:
  - Added junit-jupiter test dependency and MqttContextTest with 6 cases covering null/empty/wrong password, null/empty username, and valid credentials.

close [#6847](https://github.com/apache/shenyu/issues/6847)

### DIFF
--- a/shenyu-protocol/shenyu-protocol-mqtt/pom.xml
+++ b/shenyu-protocol/shenyu-protocol-mqtt/pom.xml
@@ -46,6 +46,12 @@
             <artifactId>reflections</artifactId>
             <version>0.9.11</version>
         </dependency>
+
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
 </project>
--- a/shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/MqttContext.java
+++ b/shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/MqttContext.java
@@ -19,6 +19,8 @@ package org.apache.shenyu.protocol.mqtt;
 
 import org.apache.commons.lang3.StringUtils;
 
+import java.util.Objects;
+
 /**
  * mqtt env.
  */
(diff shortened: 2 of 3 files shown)


Aias00 (dismissed):
## Review: #6924 fix NPE in MqttContext.isValid on null passwordInBytes

**Decision: APPROVE**

### What changed
- `MqttContext.isValid(...)`: when `passwordInBytes` is `null` (CONNECT with password flag bit = 0, i.e. anonymous / username-only clients), `new String((byte[]) null)` previously threw NPE and killed the handler thread, so the client never got a CONNACK and the connection hung. Now `null` is treated as empty string via `Objects.isNull(passwordInBytes) ? "" : new String(passwordInBytes)`, so validation fails gracefully with `CONNECTION_REFUSED_BAD_USER_NAME_OR_PASSWORD`.
- Added `import java.util.Objects;`.
- Added `junit-jupiter` test dependency (test scope) and `MqttContextTest` with 6 cases (correct creds, null/empty password, wrong password, null/empty username).

### Verification
- `Objects` import is added in the same class, so it compiles.
- The `null` branch is handled before `new String(...)`, eliminating the NPE path.
- Test covers both the null and empty-password cases.

### Notes
- Clean, minimal, spec-reasonable fix. No behavioral side effects for valid clients.

Good work — approving.

dengliming (approved):

PR #6925, opened 2026-08-16
fix: avoid NPE when selector/rule sync data is null (#6890)

### LINKED ISSUE #6890: [BUG] CONNACK always returns sessionPresent=true, even for first connect and clean-session clients
## Description
`if (CollectionUtils.isEmpty(data))` is true when `data` is null, then the next line `data.forEach(pluginDataSubscriber::unSelectorSubscribe)` dereferences null → NPE. Contrast `PluginDataRefresh`/`DiscoveryUpstreamDataRefresh`/`ProxySelectorRefresh` which return early correctly.

## Location
- `shenyu-sync-data-center/shenyu-sync-data-http/src/main/java/org/apache/shenyu/sync/data/http/refresh/SelectorDataRefresh.java:72-74`
- `shenyu-sync-data-center/shenyu-sync-data-http/src/main/java/org/apache/shenyu/sync/data/http/refresh/RuleDataRefresh.java:72-74`

## Impact
If the admin returns a response where selector or rule data deserializes to null, the HTTP refresh thread NPEs; the retry loop hits the same NPE for 10 attempts then sleeps 5 minutes, blocking selector/rule sync.

## Suggested fix
Add `return;` inside the `isEmpty(data)` branch (as the sibling refreshes do).

## Related existing
None.

juicewcode:
Fixes #6890 
## summary
- Restructure refresh() to clear the cache unconditionally first, following the same implementation as `PluginDataRefresh`, and return early when the data is empty, so a null payload clears the cache instead of crashing. Behavior for empty and non-empty data is unchanged.
- Add regression tests covering refresh with null data for both refreshers.

Make sure that:

### DIFF
--- a/shenyu-sync-data-center/shenyu-sync-data-http/src/main/java/org/apache/shenyu/sync/data/http/refresh/RuleDataRefresh.java
+++ b/shenyu-sync-data-center/shenyu-sync-data-http/src/main/java/org/apache/shenyu/sync/data/http/refresh/RuleDataRefresh.java
@@ -69,14 +69,11 @@ public class RuleDataRefresh extends AbstractDataRefresh<RuleData> {
 
     @Override
     protected void refresh(final List<RuleData> data) {
+        pluginDataSubscriber.refreshRuleDataAll();
         if (CollectionUtils.isEmpty(data)) {
             LOG.info("clear all rule cache, old cache");
-            data.forEach(pluginDataSubscriber::unRuleSubscribe);
-            pluginDataSubscriber.refreshRuleDataAll();
-        } else {
-            // update cache for UpstreamCacheManager
-            pluginDataSubscriber.refreshRuleDataAll();
-            data.forEach(pluginDataSubscriber::onRuleSubscribe);
+            return;
         }
+        data.forEach(pluginDataSubscriber::onRuleSubscribe);
     }
 }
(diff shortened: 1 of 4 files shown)


Aias00 (approved):
LGTM.

Root cause correctly identified and fixed. In the old `refresh(List<...> data)`:
```java
if (CollectionUtils.isEmpty(data)) {   // true when data == null
    data.forEach(pluginDataSubscriber::unRuleSubscribe);  // NPE on null
    ...
}
```
`CollectionUtils.isEmpty(null)` is `true`, so a null payload entered the branch and then `data.forEach(...)` dereferenced null → NPE.

The restructure calls `refreshRuleDataAll()` first and `return`s early on empty/null, eliminating the null dereference. Behavior for non-null data is unchanged (refresh-all + on*Subscribe), and it now matches the existing `PluginDataRefresh` implementation.

Verification:
- `refreshRuleDataAll()` / `on*Subscribe` / `un*Subscribe` all exist on `PluginDataSubscriber` (used by the prior code and the new tests).
- Regression tests for `refresh(null)` on both `RuleDataRefresh` and `SelectorDataRefresh` assert no NPE and that only `refreshRuleDataAll()` is invoked (never `un*Subscribe`/`on*Subscribe`).

Approving.

PR #6926, opened 2026-08-17
Fixes #6786:关闭钩子反射访问失败时避免线程异常退出

### LINKED ISSUE #6786: [BUG] TakeoverOtherHooksThread throws uncaught NPE when ApplicationShutdownHooks reflection fails
- Severity: Medium
- Location:
`shenyu-client/shenyu-client-core/src/main/java/org/apache/shenyu/client/core/shutdown/ShenyuClientShutdownHook.java:85-96`
- 
Description:
In `TakeoverOtherHooksThread.run()`, the reflection block (lines 86-93) attempts to get `java.lang.ApplicationShutdownHooks.hooks`. If this fails (`ClassNotFoundException`, `NoSuchFieldException`, `IllegalAccessException` — restricted SecurityManager, non-standard/future JDK where the internal class changes), `hooks` remains null. The catch logs the error but execution continues to line 95's while loop. At line 96, `Objects.requireNonNull(hooks)` throws `NullPointerException` outside any try-catch; the thread terminates with an uncaught exception. The entire "delay other shutdown hooks" mechanism (which ensures the shenyu offline event reaches admin before Spring's shutdown hooks close connections) is silently broken.
- 
Impact:
On environments where `ApplicationShutdownHooks` reflection fails, shutdown-hook delay is inactive. The client's `closeRepository`/`offline` may race with or run after Spring's connection-closing hooks, so the offline event never reaches admin. Stale URI entries persist on the admin side.
- 
Suggested fix:
Add `if (hooks == null) { LOG.warn("Could not access ApplicationShutdownHooks, skipping hook delay"); return; }` immediately after the try-catch block (before line 94).
- 
Confidence: High
- Related existing: none

---

BobSong-dev:
Fixes #6786
## 背景

客户端关闭时会尝试通过反射获取 JVM 的 shutdown hooks，用于延后其他关闭钩子的执行。

在 Java 21 这类有模块访问限制的环境里，反射访问可能抛出
`InaccessibleObjectException`。原来的逻辑只记录异常，后面仍会继续使用
空的 hooks 对象，最终导致后台线程再次抛出 NPE。

## 修改内容

反射访问失败或获取到的 hooks 为空时，记录告警后直接跳过延迟逻辑。

这样在无法访问 JVM 内部 shutdown hooks 的环境中，会正常降级，而不会让
关闭钩子线程异常退出。

同时补充了回归测试，覆盖反射目标不可访问的场景。

## 验证

已执行 `ShenyuClientShutdownHookTest`，测试通过。

### DIFF
--- a/shenyu-client/shenyu-client-core/src/main/java/org/apache/shenyu/client/core/shutdown/ShenyuClientShutdownHook.java
+++ b/shenyu-client/shenyu-client-core/src/main/java/org/apache/shenyu/client/core/shutdown/ShenyuClientShutdownHook.java
@@ -88,12 +88,17 @@ public class ShenyuClientShutdownHook {
                 Field field = clazz.getDeclaredField(props.getProperty("applicationShutdownHooksFieldName", "hooks"));
                 field.setAccessible(true);
                 hooks = (IdentityHashMap<Thread, Thread>) field.get(clazz);
-            } catch (ClassNotFoundException | NoSuchFieldException | IllegalAccessException ex) {
-                LOG.error(ex.getMessage(), ex);
+            } catch (ClassNotFoundException | NoSuchFieldException | IllegalAccessException | RuntimeException ex) {
+                LOG.warn("Failed to access application shutdown hooks, skip delaying other hooks.", ex);
+                return;
+            }
+            if (Objects.isNull(hooks)) {
+                LOG.warn("Application shutdown hooks are null, skip delaying other hooks.");
+                return;
             }
             long s = System.currentTimeMillis();
             while (System.currentTimeMillis() - s < delayOtherHooksExecTime) {
-                for (Iterator<Thread> iterator = Objects.requireNonNull(hooks).keySet().iterator(); iterator.hasNext();) {
+                for (Iterator<Thread> iterator = hooks.keySet().iterator(); iterator.hasNext();) {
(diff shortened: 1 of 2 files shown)


Aias00 (approved):
LGTM.

The defensive fix is sound. `TakeoverOtherHooksThread.run()` reflects into the JVM's application shutdown-hooks map to delay other hooks. Under Java 21+ module restrictions, `field.setAccessible(true)` can throw `InaccessibleObjectException` (a `RuntimeException`), which the original catch clause did not cover — so the exception propagated and the takeover thread died. Now:
- `RuntimeException` is added to the catch (covering `InaccessibleObjectException`); on failure it logs a warning and returns early, degrading gracefully.
- A new `Objects.isNull(hooks)` guard returns early instead of the previous `Objects.requireNonNull(hooks)` path that would NPE.

Verification:
- `java.util.Objects` is imported in the class.
- `props` is a `private static` field, so the regression test's `ReflectionTestUtils.setField(ShenyuClientShutdownHook.class, "props", ...)` (static set) is valid.
- `TakeoverOtherHooksThread` is a static nested class with an accessible no-arg constructor, matching the test's reflective instantiation.
- The test sets an invalid shutdown-hooks class name so `Class.forName` fails, and asserts `run()` does not throw.

Approving.

PR #6927, opened 2026-08-17
Fixes #6886:移除 RocketMQ 日志客户端重复注册的关闭钩子

### LINKED ISSUE #6886: [BUG] RocketMQ log client registers a second untracked shutdown hook leaked across reconfig
## Description
`routeMap.entrySet().removeIf(entry -> entry.getKey().startsWith(normalizedPath) || entry.getKey().startsWith(uri))` uses plain `String.startsWith`. Routes are registered as `/<path>` and `/<path>/**`. Removing server `/foo` also removes routes for `/foobar` and `/foobar/**` because `"/foobar/**".startsWith("/foo")` is true.

## Location
- `shenyu-plugin/shenyu-plugin-mcp-server/src/main/java/org/apache/shenyu/plugin/mcp/server/manager/ShenyuMcpServerManager.java:356`

## Impact
Deleting one MCP server silently breaks routing for any sibling MCP server whose path shares a prefix.

## Suggested fix
Compare on path-segment boundaries: `entry.getKey().equals(normalizedPath) || entry.getKey().startsWith(normalizedPath + "/")` (and same for `uri`).

## Related existing
None.

BobSong-dev:
Fixes #6886
## 背景

`RocketMQLogCollectClient` 在初始化成功后会自行注册一个 shutdown hook。

不过它的父类 `AbstractLogConsumeClient` 已经负责统一注册和管理关闭钩子。
子类额外注册的 hook 不会被父类追踪，配置刷新多次后可能不断累积；应用退出时
也可能重复执行 producer 的关闭逻辑。

## 修改内容

移除 RocketMQ 客户端中重复注册的 shutdown hook，统一使用父类已有的生命周期管理逻辑。

日志发送、配置刷新和正常关闭流程均未改动。

## 验证

已执行 RocketMQ 日志模块及其依赖模块测试：

- Tests run: 23
- Failures: 0
- Errors: 0
- BUILD SUCCESS

### DIFF
--- a/shenyu-e2e/shenyu-e2e-case/shenyu-e2e-case-spring-cloud/compose/script/e2e-springcloud-sync-compose.sh
+++ b/shenyu-e2e/shenyu-e2e-case/shenyu-e2e-case-spring-cloud/compose/script/e2e-springcloud-sync-compose.sh
@@ -33,13 +33,14 @@ docker network create -d bridge shenyu
 for sync in "${SYNC_ARRAY[@]}"; do
   echo -e "------------------\n"
   echo "[Start ${sync} synchronous] create shenyu-admin-${sync}.yml shenyu-bootstrap-${sync}.yml "
-  docker compose -f "$SHENYU_TESTCASE_DIR"/compose/sync/shenyu-sync-"${sync}"-eureka.yml up -d --quiet-pull
+  SYNC_COMPOSE_FILE="$SHENYU_TESTCASE_DIR"/compose/sync/shenyu-sync-"${sync}"-eureka.yml
+  docker compose -f "$SYNC_COMPOSE_FILE" up -d --quiet-pull shenyu-mysql shenyu-admin || exit 1
+  sh "$SHENYU_TESTCASE_DIR"/k8s/script/healthcheck.sh http://localhost:31095/actuator/health || exit 1
+  docker compose -f "$SYNC_COMPOSE_FILE" up -d --quiet-pull shenyu-bootstrap || exit 1
+  sh "$SHENYU_TESTCASE_DIR"/k8s/script/healthcheck.sh http://localhost:31195/actuator/health || exit 1
+  docker compose -f "${PRGDIR}"/shenyu-examples-springcloud-compose.yml up -d --quiet-pull || exit 1
   sleep 30s
-  sh "$SHENYU_TESTCASE_DIR"/k8s/script/healthcheck.sh http://localhost:31095/actuator/health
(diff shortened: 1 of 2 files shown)


Aias00 (approved):
LGTM.

Core fix is correct: `RocketMQLogCollectClient.initClient0()` previously registered its own `Runtime.getRuntime().addShutdownHook(new Thread(this::close))` on every successful init, while the parent `AbstractLogConsumeClient` already registers and centrally manages the shutdown hook (registered at line 78, removed in `close()` at line 84). On repeated config refreshes the child's hook would accumulate and the producer's close logic could run multiple times at JVM exit. Removing the duplicate is the right call, and cleanup semantics are preserved by the parent.

e2e script change is a sensible hardening: staged bring-up (mysql/admin → healthcheck → bootstrap → healthcheck → examples) with `|| exit 1` on the healthchecks so the script fails fast on unhealthy containers instead of silently continuing.

Minor note (non-blocking): the script assumes the service names `shenyu-mysql`, `shenyu-admin`, `shenyu-bootstrap` exist in `shenyu-sync-${sync}-eureka.yml` and that `k8s/script/healthcheck.sh` is present — worth a quick CI run to confirm, but it's test-only infra.

Approving.

BobSong-dev:
> LGTM.
> 
> Core fix is correct: `RocketMQLogCollectClient.initClient0()` previously registered its own `Runtime.getRuntime().addShutdownHook(new Thread(this::close))` on every successful init, while the parent `AbstractLogConsumeClient` already registers and centrally manages the shutdown hook (registered at line 78, removed in `close()` at line 84). On repeated config refreshes the child's hook would accumulate and the producer's close logic could run multiple times at JVM exit. Removing the duplicate is the right call, and cleanup semantics are preserved by the parent.
> 
> e2e script change is a sensible hardening: staged bring-up (mysql/admin → healthcheck → bootstrap → healthcheck → examples) with `|| exit 1` on the healthchecks so the script fails fast on unhealthy containers instead of silently continuing.
> 
> Minor note (non-blocking): the script assumes the service names `shenyu-mysql`, `shenyu-admin`, `shenyu-bootstrap` exist in `shenyu-sync-${sync}-eureka.yml` and that `k8s/script/healthcheck.sh` is present — worth a quick CI run to confirm, but it's test-only infra.
> 
> Approving.

Thanks for the careful review and approval.
I also checked the point about the compose service names and the healthcheck script with the CI run. The Spring Cloud E2E job passed with the staged startup flow, and the full E2E workflow is green, so the referenced services and k8s/script/healthcheck.sh are available in this scenario.
Appreciate the review.

PR #6929, opened 2026-08-17
[ISSUE #6773] fix: handle unset response status in response transformer

### LINKED ISSUE #6773: [BUG] Response template assembleMessage NPEs when response status is unset
- severity: Low-Medium
- files: `shenyu-plugin/shenyu-plugin-ai/shenyu-plugin-ai-response-transformer/.../template/AiResponseTransformerTemplate.java:190`
- description: `responseNode.put("status", exchange.getResponse().getStatusCode().value())` unconditionally calls `.value()` on `getStatusCode()`. If the upstream has not set a status code, `getStatusCode()` returns null → NPE. Caught by outer `onErrorResume` → silently skips transformation.
- impact: Responses with unset status silently skip transformation and log stack traces.
- suggested_fix: Null-check `getStatusCode()` before `.value()`.
- confidence: Medium
- related_existing: none

---

yykaue:
Fixes #6773.

`AiResponseTransformerTemplate.assembleMessage()` previously called `value()` directly on the response status code. When the upstream response had no explicitly configured status code, `getStatusCode()` returned `null`, causing a `NullPointerException` and preventing response transformation.

This PR:

- Checks whether the response status code is present before accessing its value.
- Omits the `response.status` field when no status has been explicitly set.
- Preserves the existing behavior for responses with a configured status code.
- Adds a regression test covering responses with an unset status code.

## Verification

- The regression test reproduces the original `NullPointerException` before the fix.
- The regression test passes after the fix.
- Compilation and Checkstyle pass.

@Aias00 Could you please review this PR?

### DIFF
--- a/shenyu-plugin/shenyu-plugin-ai/shenyu-plugin-ai-response-transformer/src/main/java/org/apache/shenyu/plugin/ai/transformer/response/template/AiResponseTransformerTemplate.java
+++ b/shenyu-plugin/shenyu-plugin-ai/shenyu-plugin-ai-response-transformer/src/main/java/org/apache/shenyu/plugin/ai/transformer/response/template/AiResponseTransformerTemplate.java
@@ -23,6 +23,7 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import org.springframework.core.io.buffer.DataBuffer;
 import org.springframework.core.io.buffer.DataBufferUtils;
 import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpStatusCode;
 import org.springframework.http.MediaType;
 import org.springframework.http.server.reactive.ServerHttpRequest;
 import org.springframework.web.server.ServerWebExchange;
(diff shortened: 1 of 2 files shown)


Aias00 (approved):
LGTM.

Root cause addressed correctly: `exchange.getResponse().getStatusCode()` can return `null` when no status is explicitly set, and the old code called `.value()` on it directly → NPE. The new `Objects.nonNull(responseStatus)` guard prevents that.

Verification:
- `java.util.Objects` is already imported in `AiResponseTransformerTemplate.java` (used elsewhere, e.g. `Objects.isNull(body)`), so `Objects.nonNull` compiles without a new import.
- `HttpStatusCode` import added is correct for Spring Framework 6 (`getStatusCode()` returns `HttpStatusCode`).
- Behavior for responses with a configured status is preserved; the `response.status` field is simply omitted when unset.
- Regression test `AiResponseTransformerTemplateStatusTest` reproduces the null-status path and asserts the field is absent.

Approving.

PR #6934, opened 2026-08-17
[Fix] Guard InstanceCheckService executor shutdown (#6705)

### LINKED ISSUE #6705: [BUG] InstanceCheckService.close() NPEs if setup() was never invoked
- severity: Medium
- files: `shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/InstanceCheckService.java:178-183`
- description: `close()` calls `executor.shutdown()` unconditionally, but `executor` is only assigned in `setup()` (line 86). If the bean is constructed but `setup()` is not called (conditional startup, profile, or test context), `executor` is null and `close()` throws NPE.
- impact: NPE during application shutdown in configurations where instance checking is not initialized.
- suggested_fix: Guard with `if (Objects.nonNull(executor)) { executor.shutdown(); }`.
- confidence: Medium
- related_existing: none.

---

yykaue:
## Description

Fixes #6705.

`InstanceCheckService.close()` previously called `executor.shutdown()` unconditionally. When `setup()` was not invoked, the executor remained `null`, causing an NPE during application shutdown.

This PR:

- Adds a null check before shutting down the executor.
- Preserves the existing database synchronization and instance cleanup behavior.
- Adds a regression test for calling `close()` without invoking `setup()`.

## Tests

- `InstanceCheckServiceTest`: 8 tests passed.
- `shenyu-admin`: 1,440 tests passed, 0 failures, 1 skipped.
- Checkstyle passed with 0 violations.

@Aias00 Please review this PR when you have time.

### DIFF
--- a/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/InstanceCheckService.java
+++ b/shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/InstanceCheckService.java
@@ -39,6 +39,7 @@ import java.util.Arrays;
 import java.util.Deque;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
@@ -179,7 +180,9 @@ public class InstanceCheckService {
     public void close() {
         syncDB();
         instanceHealthBeatInfo.clear();
-        executor.shutdown();
+        if (Objects.nonNull(executor)) {
+            executor.shutdown();
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Aias00 (approved):
## Summary

Guards `InstanceCheckService.close()` against NPE when `setup()` was never called (#6705).

- Added `if (Objects.nonNull(executor))` before `executor.shutdown()` in `close()`.
- Added `import java.util.Objects;`.
- Added regression test `testCloseWithoutSetup` verifying `close()` does not throw when `setup()` was not invoked (executor still `null`).

## Verification

- `executor` is an existing field of the class (the original unconditional `executor.shutdown()` already referenced it), so only the null guard is new — no missing-field/compile risk.
- `Objects` import added explicitly.
- Test `assertDoesNotThrow(instanceCheckService::close)` covers the previously-failing path while the existing sync/clear behavior is preserved.

## Conclusion

Small, correct defensive fix with a regression test. Approving.

PR #6937, opened 2026-08-17
fix: cancel MCP plugin chain on tool call timeout

### LINKED ISSUE #6888: [BUG] JsonSchemaUtil never emits required array — all MCP tool parameters advertised as optional
## Description
`executeToolCall` subscribes fire-and-forget (`chain.execute(decoratedExchange).subscribe(...)` at line 266) then blocks on `responseFuture.get(DEFAULT_TIMEOUT_SECONDS, …)`. On `TimeoutException` it throws, but the `Disposable` returned by `subscribe()` is never captured/cancelled, and `doFinally` does not cancel the chain. The downstream request keeps running after the tool call has already returned an error to the MCP client.

## Location
- `shenyu-plugin/shenyu-plugin-mcp-server/src/main/java/org/apache/shenyu/plugin/mcp/server/callback/ShenyuToolCallback.java:266-283`

## Impact
Timed-out tool calls leak downstream work/connections and may produce side effects (e.g. upstream API calls) after the client has received an error.

## Suggested fix
Capture the `Disposable`, and on `TimeoutException` call `disposable.dispose()` before throwing.

## Related existing
None — distinct from MCP-01..05 (#6640-#6644, response decorator) and MCP-11 (notifications).

SouthwestAsiaFloat:
Fixes #6888

## Description

`ShenyuToolCallback` subscribes to the plugin chain asynchronously and waits
for the response with a timeout. However, the `Disposable` returned by
`subscribe()` was not retained, so the downstream plugin chain continued
running after the tool call had already timed out.

This change captures the subscription and disposes it when
`CompletableFuture#get` throws `TimeoutException`, preventing leaked
downstream work and unintended side effects after a timeout response.

## Changes

- Capture the `Disposable` returned by the plugin-chain subscription.
- Dispose the subscription when the tool call times out.
- Preserve the existing default timeout behavior.
- Add a regression test verifying that timeout cancels a non-terminating
  Reactor chain.


Aias00 (approved):
## Summary

Fixes a downstream-plugin-chain leak on MCP tool-call timeout (#6888).

- Captures the `Disposable` returned by the plugin-chain subscription (`chain.execute(...)` returns a `Mono`, which implements `Disposable`).
- On `CompletableFuture#get` throwing `TimeoutException`, calls `disposable.dispose()` to cancel the still-running downstream chain.
- Default timeout behavior preserved by adding an overloaded `executeToolCall(..., long timeoutSeconds)` with `DEFAULT_TIMEOUT_SECONDS` as the default; the public method delegates to it.

## Verification

- `final Disposable disposable = chain.execute(decoratedExchange).doOnSubscribe(...).doOnError(...)` — `Mono<Void>` is assignable to `Disposable`, compiles.
- `TimeoutException` import added; `reactor.core.Disposable` import added.
- Regression test `testExecuteToolCallDisposesPluginChainOnTimeout` stubs `chain.execute` with `Mono.never().doOnCancel(...)`, calls with `timeoutSeconds=0`, and asserts the timeout exception is thrown **and** the chain was cancelled (`doOnCancel` fired) — proving `dispose()` triggers cancellation.

## Suggestion (non-blocking)

Consider disposing on other failure paths too (e.g. `ExecutionException` wrapping a real error), not only `TimeoutException`, so any leaked subscription is cleaned up. The current scope already covers the main leak scenario.

## Conclusion

Logic is sound and the regression test is meaningful. Approving.

Aias00 (approved):
Approved as PMC (Aias00). Passing CI, mergeable, small targeted fix. Review only - not merged.

PR #6939, opened 2026-08-17
fix: prevent NPE on mqtt bootstrap server shutdown after failed start

### LINKED ISSUE #6747: [BUG] MqttBootstrapServer shutdown NPE on failed start; start swallows InterruptedException
- severity: Medium
- files: `shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/MqttBootstrapServer.java:57-79`
- description: `start()` catches `InterruptedException` from `bootstrap.bind(...).sync()` and does nothing (no log, no interrupt-flag restore, no cleanup). If bind fails, `future` stays null. `shutdown()` calls `future.channel().close()` → NPE on `future`.
- impact: Failed/incomplete start leads to NPE on shutdown and leaked event-loop groups; interrupt status lost.
- suggested_fix: Null-check `future`/groups in shutdown; on bind failure dispose groups and re-interrupt.
- confidence: Medium
- related_existing: none

---

wy471x:

<!--
-->
Make sure that:

## Summary

### Changes:

MqttBootstrapServer (shenyu-protocol-mqtt):                                                                                                                                        
                         
  - Failed-start NPE fix: start() previously swallowed InterruptedException, leaving future null so shutdown() threw NPE. Now bind failures are logged, event-loop groups are        
  disposed via shutdown(), and the interrupt flag is restored.                                                                                                                       
  - Null-safe shutdown: shutdown() null-checks bossGroup, workerGroup, and future using Objects.nonNull (checkstyle requirement).                                                    
  - Async start: bind logic moved into CompletableFuture.runAsync(...) (module convention), resolving the //// todo thread start mqtt server marker; exceptions are handled inside   
  the task since callers can no longer observe them. Fields made volatile for cross-thread visibility.                                                                               
  - Logging: added LOG.info on successful start and LOG.error in init()/start() catch blocks.                                                                                        
                                                                                                                                                                                    
### Test Cases: 

MqttBootstrapServerTest:

  - new MqttBootstrapServerTest with 3 cases — shutdown without start, port-in-use start then shutdown, free-port start (await connect) then shutdown.

close [#6747](https://github.com/apache/shenyu/issues/6747)

### DIFF
--- a/shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/MqttBootstrapServer.java
+++ b/shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/MqttBootstrapServer.java
@@ -26,56 +26,74 @@ import io.netty.util.ResourceLeakDetector;
 import org.apache.shenyu.common.utils.Singleton;
 import org.apache.shenyu.protocol.mqtt.repositories.BaseRepository;
 import org.reflections.Reflections;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.Locale;
+import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
 
 /**
  * mqtt server.
  */
 public class MqttBootstrapServer implements BootstrapServer {
 
+    private static final Logger LOG = LoggerFactory.getLogger(MqttBootstrapServer.class);
+
     private static final String REPOSITORY_PACKAGE_NAME = "org.apache.shenyu.protocol.mqtt.repositories";
 
     private static final MqttContext ENV = new MqttContext();
 
-    private EventLoopGroup bossGroup;
+    private volatile EventLoopGroup bossGroup;
 
-    private EventLoopGroup workerGroup;
+    private volatile EventLoopGroup workerGroup;
 
-    private ChannelFuture future;
+    private volatile ChannelFuture future;
 
     @Override
     public void init() {
         try {
             initRepositories();
         } catch (Exception e) {
-            //// todo log
+            LOG.error("MQTT server init repositories failed", e);
         }
     }
 
     @Override
     public void start() {
-        //// todo thread start mqtt server
         ResourceLeakDetector.setLevel(ResourceLeakDetector.Level.valueOf(ENV.getLeakDetectorLevel().toUpperCase(Locale.ROOT)));
-        bossGroup = new NioEventLoopGroup(ENV.getBossGroupThreadCount());
-        workerGroup = new NioEventLoopGroup(ENV.getWorkerGroupThreadCount());
-        ServerBootstrap bootstrap = new ServerBootstrap();
-        bootstrap.group(bossGroup, workerGroup)
-                .channel(NioServerSocketChannel.class)
-                .childHandler(new MqttTransportServerInitializer(ENV.getMaxPayloadSize()));
-        try {
-            future = bootstrap.bind(ENV.getPort()).sync();
-            //// todo log
-        } catch (InterruptedException e) {
-            //// todo log
-        }
+        CompletableFuture.runAsync(() -> {
+            bossGroup = new NioEventLoopGroup(ENV.getBossGroupThreadCount());
+            workerGroup = new NioEventLoopGroup(ENV.getWorkerGroupThreadCount());
+            ServerBootstrap bootstrap = new ServerBootstrap();
+            bootstrap.group(bossGroup, workerGroup)
+                    .channel(NioServerSocketChannel.class)
+                    .childHandler(new MqttTransportServerInitializer(ENV.getMaxPayloadSize()));
+            try {
+                future = bootstrap.bind(ENV.getPort()).sync();
+                LOG.info("MQTT server started on port {}", ENV.getPort());
+            } catch (InterruptedException e) {
+                LOG.error("MQTT server bind interrupted", e);
+                shutdown();
+                Thread.currentThread().interrupt();
+            } catch (Exception e) {
+                LOG.error("MQTT server failed to bind on port {}", ENV.getPort(), e);
+                shutdown();
+            }
+        });
     }
 
     @Override
     public void shutdown() {
-        bossGroup.shutdownGracefully();
-        workerGroup.shutdownGracefully();
-        future.channel().close();
+        if (Objects.nonNull(bossGroup)) {
+            bossGroup.shutdownGracefully();
+        }
+        if (Objects.nonNull(workerGroup)) {
+            workerGroup.shutdownGracefully();
+        }
+        if (Objects.nonNull(future)) {
+            future.channel().close();
+        }
     }
 
     private void initRepositories() throws IllegalAccessException, InstantiationException {
--- a/shenyu-protocol/shenyu-protocol-mqtt/src/test/java/org/apache/shenyu/protocol/mqtt/MqttBootstrapServerTest.java
+++ b/shenyu-protocol/shenyu-protocol-mqtt/src/test/java/org/apache/shenyu/protocol/mqtt/MqttBootstrapServerTest.java
@@ -1,0 +1,90 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


Aias00 (dismissed):
## Review: #6939 prevent NPE on mqtt bootstrap server shutdown after failed start

**Decision: APPROVE**

### What changed
- `MqttBootstrapServer.shutdown()`: null-checks `bossGroup`, `workerGroup`, and `future` before `shutdownGracefully()` / `channel().close()`, so calling `shutdown()` before a successful `start()` (or after a failed start) no longer throws NPE.
- `bossGroup` / `workerGroup` / `future` are now `volatile` for safe publication between the async start thread and `shutdown()`.
- `start()` now runs the Netty bootstrap inside `CompletableFuture.runAsync(...)`; on `InterruptedException` or any bind failure it logs and calls `shutdown()` to clean up partially-initialized groups (and re-interrupts on `InterruptedException`). Added `Logger` and info/error logging (replacing `//// todo log` placeholders).
- Tests:
  - `shutdownWithoutStartDoesNotThrow` — directly validates the null-safe shutdown fix.
  - `startWithPortInUseThenShutdownDoesNotThrow` — occupies the port, asserts the server self-shuts-down its groups on bind failure.
  - `startOnFreePortThenShutdownDoesNotThrow` — normal start then shutdown.

### Verification
- `MqttContext` fields (`port`, `bossGroupThreadCount`, `maxPayloadSize`, `leakDetectorLevel`, etc.) are `static`, so the test's `new MqttContext().setPort(...)` actually configures the static `ENV` singleton used by `start()` — the integration tests exercise the intended port and behavior.
- `volatile` + null-guards make the async start / shutdown interaction safe.
- `shutdown()` is invoked defensively on bind failure, preventing leaked `EventLoopGroup`s.

### Notes
- Correct NPE fix with proper concurrency handling and tests. Approving.

Aias00 (dismissed):
## Summary

This PR fixes a `NullPointerException` in `MqttBootstrapServer.shutdown()` that occurred when `shutdown()` was invoked before `start()` assigned the Netty `EventLoopGroup`/`ChannelFuture` fields (e.g. after a failed bind), and adds proper logging/cleanup.

What is correct:
- `bossGroup`, `workerGroup` and `future` are now guarded with `Objects.nonNull(...)` in `shutdown()` — the NPE on a never-started or partially-started server is gone.
- `init()` now logs the repository-init failure instead of a silent `// todo log`, so startup problems are observable.
- `start()` wraps bind in try/catch and calls `shutdown()` on `InterruptedException`/bind failure, preventing leaked event-loop groups after a failed start.
- `start()` is now executed via `CompletableFuture.runAsync(...)` so the caller is not blocked on `bind().sync()`; fields are `volatile` so the async assignment is safely published to a later `shutdown()` on another thread.
- The three new tests cover the previously-untested lifecycle paths (shutdown-without-start, bind-failure, normal start+shutdown).

## Conclusion

Approved. The core NPE fix and the added tests are sound.

## Non-blocking suggestions
- `start()` is now non-blocking (async bind). Callers that previously relied on `start()` returning only after the port was bound may now observe a "started" server before it actually accepts connections. A short comment on `start()` documenting the async semantics would help future readers.
- Minor: the `CompletableFuture.runAsync` uses the common `ForkJoinPool`; if MQTT startup should be isolated, an explicit executor could be passed, but that is optional.

Thanks for adding the regression tests.

Aias00:
CI note (PMC Aias00): the PR's check rollup showed FAILURE, but at the check-run level there are no actual failures — only cancelled checks from superseded runs. This PR appears CI-clean. Please re-run the cancelled checks to get a clean green status before merge.

Aias00 (changes requested):
## Review (PMC Aias00) — `fix: prevent NPE on mqtt bootstrap server shutdown after failed start`

**Decision: REQUEST CHANGES**

### 🔴 Blocking — making `start()` asynchronous breaks `MqttPluginDataHandler`

`start()` now hands the bind off to `CompletableFuture.runAsync(...)`, so **`start()` returns before the port is bound** and there is no completion signal. `MqttPluginDataHandler` depends on the old synchronous contract:

```java
private final BootstrapServer server = new MqttBootstrapServer();

public void handlerPlugin(final PluginData pluginData) {
    if (pluginData.getEnabled()) {
        server.start();     // <- used to block until bound
    } else {
        server.shutdown();
    }
}
```

CI already proves this is broken — `build (21, ubuntu-latest)` on this exact head (`5202a904`) fails with:

```
MqttPluginDataHandlerTest.testEnableConfiguration:58 expected: <true> but was: <false>
... Failed to execute goal ...maven-surefire-plugin... on project shenyu-plugin-mqtt: There are test failures.
```

The test calls `handlerPlugin(enabled)` and then immediately asserts the port is in use. With async bind the assertion races and loses.

The production consequence is worse than the test failure: an admin toggling the MQTT plugin **enable → disable** can now run `shutdown()` while `start()` is still in flight. `shutdown()` reads all three fields, finds them `null`, and silently returns — then the async task binds the port *afterwards*, leaving `NioEventLoopGroup` threads and a listening socket that nothing owns. The plugin reports "disabled" while still accepting connections, and there is no way to stop it short of a restart.

**Please either:**
- keep `start()` synchronous (restore the blocking `bootstrap.bind(port).sync()`), or
- if async is genuinely needed, hold the returned `CompletableFuture` in a field, expose it (or a `started` latch) so callers can await completion, and make `shutdown()` cancel/interrupt a pending start before it binds.

The second option is more work; the first is a one-line change and matches what `BootstrapServer.start()` callers already assume.

### 🟠 Also worth fixing

2. **`CompletableFuture.runAsync` uses `ForkJoinPool.commonPool()`.** The task runs `bind().sync()`, which blocks. Blocking common-pool threads is an anti-pattern (parallelism is `cpus - 1`, shared with every other `runAsync` caller in the JVM), so MQTT startup can be starved by unrelated work. If you keep it async, use a dedicated named executor (e.g. `ShenyuThreadFactory.create("mqtt-bootstrap", false)`).

3. **The title's NPE is not what this PR actually fixes.** `shutdown()` already null-guards `future`, `bossGroup` and `workerGroup`, so I cannot reproduce an NPE on shutdown after a failed start. What this PR *does* fix is real and valuable: on bind failure the old code leaked both `NioEventLoopGroup`s (the `//// todo log` catch block swallowed the exception and never shut them down). Please retitle/reword so reviewers and the release notes describe the actual fix.

4. **`startOnFreePortThenShutdownDoesNotThrow` does not test a free port.** The `ServerSocket` stays open inside the try-with-resources block, so the MQTT server still hits "address already in use":
   ```java
   try (ServerSocket socket = new ServerSocket(0)) {
       int freePort = socket.getLocalPort();
       env.setPort(freePort);
       server.start();
       await()...until(() -> canConnect(freePort));   // succeeds: connects to *this* socket
   ```
   `canConnect(freePort)` is satisfied by the test's own `ServerSocket`, not by the MQTT server — so the assertion is vacuous. Close the socket (or release the port) before `server.start()` if you want a genuine success-path test.

### ✅ What is good

- Replacing the four `//// todo log` placeholders with real `LOG.error`/`LOG.info` — including the previously silent `init()` failure.
- `volatile` on `bossGroup` / `workerGroup` / `future` is correct and necessary now that they are written from another thread.
- Calling `shutdown()` from the failure handlers, and `Thread.currentThread().interrupt()` after `InterruptedException`.
- `shutdownWithoutStartDoesNotThrow` is a useful guard for the never-started case.

Happy to re-review once `start()` gives callers a completion signal.

shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/MqttBootstrapServer.java
```diff
@@ -23,52 +23,64 @@
...
-        bootstrap.group(bossGroup, workerGroup)
-                .channel(NioServerSocketChannel.class)
-                .childHandler(new MqttTransportServerInitializer(ENV.getMaxPayloadSize()));
-        try {
-            future = bootstrap.bind(ENV.getPort()).sync();
-            //// todo log
-        } catch (InterruptedException e) {
-            //// todo log
-        }
+        CompletableFuture.runAsync(() -> {
```
Aias00, line 65: 🔴 **Blocking — this makes `start()` fire-and-forget and breaks its callers.**

`MqttPluginDataHandler` relies on `start()` blocking until the port is bound:
```java
if (pluginData.getEnabled()) {
    server.start();
} else {
    server.shutdown();
}
```
Now `start()` returns immediately, so:
- `MqttPluginDataHandlerTest.testEnableConfiguration` fails on this PR's CI (`expected: <true> but was: <false>` — the port isn't bound yet);
- worse in production: an enable → disable toggle can call `shutdown()` while the bind is still in flight. `shutdown()` sees all three fields as `null`, returns silently, and the server then comes up unowned — leaked `NioEventLoopGroup`s and a listening port that can never be stopped.

Also, `runAsync` runs on `ForkJoinPool.commonPool()` while the task blocks on `bind().sync()` — blocking common-pool threads is best avoided.

Please restore the synchronous bind, or keep the `CompletableFuture` in a field and expose it so callers can await completion, with `shutdown()` cancelling any pending start.

PR #6943, opened 2026-08-18
fix: Log desensitization silently replaces non-JSON bodies/query stri…

### LINKED ISSUE #6845: [BUG] Log desensitization silently replaces non-JSON bodies/query strings with {} (data loss)
## Description
`DataDesensitizeUtils.desensitizeBody` does `Map<String,String> bodyMap = JsonUtils.jsonToMap(source, String.class)`. `JsonUtils.jsonToMap` catches `IOException` and returns an empty `LinkedHashMap` for any non-JSON input. The code then `bodyMap.forEach(...)` (no-op) and returns `JsonUtils.toJson(bodyMap)` = `"{}"` for all non-JSON input. `queryParams` is `request.getURI().getQuery()` (e.g. `a=1&b=2`) — never JSON. `requestBody`/`responseBody` are arbitrary content (XML, form-encoded, plain text, HTML).

## Location
- `shenyu-plugin/shenyu-plugin-logging/shenyu-plugin-logging-desensitize-api/src/main/java/org/apache/shenyu/plugin/logging/desensitize/api/utils/DataDesensitizeUtils.java:87-100`
- `shenyu-plugin/shenyu-plugin-logging/shenyu-plugin-logging-common/src/main/java/org/apache/shenyu/plugin/logging/common/collector/AbstractLogCollector.java:207-211`
- `shenyu-plugin/shenyu-plugin-logging/shenyu-plugin-logging-common/src/main/java/org/apache/shenyu/plugin/logging/common/AbstractLoggingPlugin.java:109`

## Impact
Whenever desensitization is enabled, legitimate non-JSON request/response bodies and query strings are replaced in the collected log with `"{}"` — permanent silent data loss in the audit log plus false appearance of empty payloads.

## Suggested fix
Only invoke `desensitizeBody` when `source` parses as JSON (pre-check or fall back to `desensitizeForSingleWord`); for `queryParams`, apply key/value desensitization on the parsed query string rather than treating it as JSON.

## Related existing
Distinct from LOG-13 (#6770, `KeyWordMatch` regex) and #6581 (`LogCollectUtils` JSON-on-request-thread).

hengyuss:
…ngs with {} (data loss)

  Fixes #6845.

  Log desensitization currently treats request bodies, response bodies, and query strings as JSON. When the source is
  not a JSON object, `JsonUtils.jsonToMap` returns an empty map, causing the original content to be replaced with `{}`.

  ## Changes

  - Preserve the original body when it is not a valid JSON object.
  - Add a JSON parsing method that distinguishes parsing failures from valid empty objects.
  - Preserve unmatched JSON value types with `Map<String, Object>`.
  - Handle query parameters separately instead of parsing them as JSON.
  - Reuse `desensitizeSingleKeyword` for query parameter desensitization.
  - Add regression tests for non-JSON bodies, JSON arrays, query parameters, duplicate parameters, and malformed
  encoding.

  ## Tests

  - `JsonUtilsTest`
  - `DataDesensitizeUtilsTest`
  - Compilation of `shenyu-plugin-logging-common` and required modules

Make sure that:


Aias00 (dismissed):
## Review: #6943 fix Log desensitization silently replaces non-JSON bodies/query strings

**Decision: APPROVE**

### What changed
- `JsonUtils.tryJsonToMap(...)`: new helper returning `Optional<Map<String,Object>>`; returns `Optional.empty()` when the input is not a JSON object (catches `IOException`) instead of corrupting it.
- `DataDesensitizeUtils.desensitizeBody(...)`: now uses `tryJsonToMap`; if the source is not a JSON object (plain text / XML / JSON array) it is returned unchanged — fixing the silent data-loss bug. Adds `Objects.nonNull(value)` guard and uses `String.valueOf(value)`.
- `DataDesensitizeUtils.desensitizeQueryParams(...)` (new): desensitizes query-string values by parameter name; splits on `&`, URL-decodes key/value, desensitizes matched values, re-encodes; leaves valueless params and `IllegalArgumentException` cases untouched.
- `AbstractLogCollector`: query params now routed through `desensitizeQueryParams` instead of `desensitizeForBody`.
- Tests: `JsonUtilsTest.tryJsonToMap`, `DataDesensitizeUtilsTest` now asserts plainText/XML/jsonArray pass through unchanged, plus `desensitizeQueryParamsTest` (incl. `name=%invalid` malformed-encoding case).

### Verification
- `DataDesensitizeFactory.selectDesensitize(String, String)` exists (master L38) — `String.valueOf(value)` matches.
- `desensitizeSingleKeyword(boolean, String, String, KeyWordMatch, String)` 5-arg overload exists (master L69) — query-param call matches.
- The non-JSON passthrough genuinely fixes data loss; null guard prevents NPE on null values; query-param path is URL-safe with proper fallbacks.

### Notes
- Solid, behavior-preserving fix with good coverage. Approving.

Aias00 (approved):
## Review: #6943 fix: Log desensitization silently replaces non-JSON bodies/query strings with data loss

**Verdict: APPROVE**

### Analysis
The previous `desensitizeBody` called `JsonUtils.jsonToMap(source)` which, for non-JSON input (plain text, XML, JSON arrays), returns an empty map; the method then returned `JsonUtils.toJson(emptyMap)` = `"{}"`, **silently discarding the original logged body**. This PR fixes the data-loss behavior and adds proper query-param handling.

**Changes verified:**
- `JsonUtils.tryJsonToMap` (new) returns `Optional<Map<String,Object>>`, empty when the source is not a JSON object — used to short-circuit and return the original `source` unchanged.
- `desensitizeBody` now returns the original body for non-JSON input, guards null values (`String.valueOf(value)`), and only re-serializes when the source is a real JSON object.
- `desensitizeQueryParams` (new) splits the query string, URL-decodes key/value, desensitizes the value when the key matches, and re-encodes — so query params get per-name desensitization instead of being force-parsed as a JSON body. `AbstractLogCollector` now routes query params through it; decode failures fall back to the original parameter (no data loss).

**Tests:** `DataDesensitizeUtilsTest` adds `desensitizeQueryParamsTest` (multi-value, `%invalid` decode failure) and extends `desensitizeBodyTest` to assert plain text / XML / JSON-array bodies are returned unchanged. `JsonUtilsTest.tryJsonToMap` covers object / plain-text / array cases. Together they lock in the no-data-loss guarantee.

### Non-blocking note
`desensitizeQueryParams` re-encodes values via `URLEncoder` (application/x-www-form-urlencoded), which normalizes already-encoded special characters in logged values. Acceptable for a logging feature, just worth noting.

### Decision
APPROVE.

PR #6944, opened 2026-08-18
Fixes #6769：Ensure RabbitMQ connection closes when channel close fails

### LINKED ISSUE #6769: [BUG] RabbitmqLogCollectClient.close0 leaks connection if channel.close() throws
- severity: Medium
- files: `shenyu-plugin/shenyu-plugin-logging/shenyu-plugin-logging-rabbitmq/.../RabbitmqLogCollectClient.java:126-138`
- description: `close0()` calls `channel.close()` then `connection.close()` within a single try-catch that only catches `IOException`. If `channel.close()` throws, `connection.close()` is never reached. The RabbitMQ TCP connection is leaked.
- impact: Each channel-close failure leaks a RabbitMQ connection; exhausts server connection limit over time.
- suggested_fix: Use separate try-finally blocks for `channel.close()` and `connection.close()`.
- confidence: Medium
- related_existing: none

---

BobSong-dev:
Fixes #6769

## Background

`RabbitmqLogCollectClient.close0()` previously closed the channel and connection in the same `try` block.

If `channel.close()` threw an `IOException`, control jumped to the `catch` block and `connection.close()` was skipped. This could leave the RabbitMQ connection open.

## Changes

- Handle channel and connection closure separately;
- Attempt to close the connection in `finally` even when channel closure fails;
- Log channel and connection close failures separately;
- Add a regression test to verify that the connection is still closed when channel closure fails.

## Verification

- `RabbitmqLogCollectClientTest`: 2 tests run, 0 failures, 0 errors;
- Full tests for the RabbitMQ module and its dependencies passed;
- RabbitMQ module: 17 tests run, 0 failures, 0 errors;
- `BUILD SUCCESS`;
- `git diff --check` passed.


Aias00 (approved):
Thanks @BobSong-dev!

The fix is correct: previously `channel.close()` throwing `IOException` skipped `connection.close()` entirely, leaking the connection. Separating the two closures and moving the connection close into `finally` with its own try/catch guarantees the connection is attempted regardless of the channel outcome, and the failure messages now correctly attribute which resource failed.

The regression test directly covers the leak scenario (channel close throws → connection.close() still invoked).

One minor, non-blocking observation: `com.rabbitmq.client.Channel#close()` also declares `TimeoutException`, which the `catch (IOException)` around the channel close won't intercept — in that case the exception propagates to the caller of `close0()` (which is fine since the signature allows `Exception` and the `finally` still closes the connection). Not worth blocking on, but could be caught in a follow-up if you want the channel failure logged consistently too.

LGTM, approving.

Issue #6709, opened 2026-08-03
[BUG] ScaleRuleServiceImpl.update uses full updateByPrimaryKey, nulling omitted fields


Aias00:
- severity: Medium
- files: `shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/ScaleRuleServiceImpl.java:124-131`
- description: `update(...)` calls `scaleRuleMapper.updateByPrimaryKey(after)` (full column update). Contrast with `create` which uses `insertSelective` and the convention elsewhere (`AlertReceiverServiceImpl`/`DetailServiceImpl`/`FieldServiceImpl` use `updateByPrimaryKeySelective`). Any field the DTO does not populate (e.g. `sort`, `status`, `minimum`, `maximum`) is written as NULL. Also rebuilds the DO a second time on line 128 (same double-build pattern as N11).
- impact: Partial PUT payloads silently wipe scale-rule thresholds/state.
- suggested_fix: Use `updateByPrimaryKeySelective(after)` and reuse `after` instead of rebuilding.
- confidence: Medium
- related_existing: none.

---
_Identified during the 2026-08-02 deep re-scan; full list in [`docs/scan2-2026-08-02/06-medium-tiers.md`](docs/scan2-2026-08-02/06-medium-tiers.md)._

PR #6945, opened 2026-08-18
fix: use computeIfAbsent to prevent lost subscriber updates

### LINKED ISSUE #6746: [BUG] SubscribeRepository.add lost-update race loses subscribers
- severity: Medium
- files: `shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/repositories/SubscribeRepository.java:43-48,56-62,90-92`
- description: `add(Channel, List<MqttTopicSubscription>)` runs in `CompletableFuture.runAsync` + `parallelStream`, calls `get(topic)` which returns a fresh `new CopyOnWriteArrayList<>()` via `getOrDefault` for an absent topic (a transient list not inserted into the map), `channels.add(channel)`, then `put`. Two concurrent subscribers to a new topic each get an empty list, add themselves, and `put` — last write wins, first subscriber silently dropped.
- impact: Subscribers silently lost under concurrent subscription to the same topic; messages not delivered to all.
- suggested_fix: Use `computeIfAbsent` to obtain the canonical list and atomically add.
- confidence: High
- related_existing: none

---

wy471x:
Make sure that:

### Changes:
  
Fix: lost-update race in SubscribeRepository.add()
  Both add() overloads used a get-then-put pattern: get(topic) returned a transient CopyOnWriteArrayList for an absent topic (via getOrDefault), the channel was added to that transient list, then the list was put into TOPIC_CHANNEL_FACTORY. Under concurrent subscription to a new topic, each thread obtained its own empty list and the last put won —silently dropping the first subscriber.

 Fix: both overloads now use TOPIC_CHANNEL_FACTORY.computeIfAbsent(topic, t -> new CopyOnWriteArrayList<>()) to atomically obtain the canonical list stored in the map, followed by a thread-safe add/addAll on the CopyOnWriteArrayList.

### Test Cases:
  
Tests: unit tests for SubscribeRepository
Added SubscribeRepositoryTest with 4 tests: basic registration via both add() overloads, plus two concurrency regression tests asserting 50 concurrent subscribers to a new topic are all retained.

close [#6746](https://github.com/apache/shenyu/issues/6746)

### DIFF
--- a/shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/repositories/SubscribeRepository.java
+++ b/shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/repositories/SubscribeRepository.java
@@ -41,11 +41,8 @@ public class SubscribeRepository implements BaseRepository<List<String>, List<Ch
 
     @Override
     public void add(final List<String> topics, final List<Channel> channels) {
-        CompletableFuture.runAsync(() -> topics.parallelStream().forEach(s -> {
-            List<Channel> list = get(s);
-            list.addAll(channels);
-            TOPIC_CHANNEL_FACTORY.put(s, list);
-        }));
+        CompletableFuture.runAsync(() -> topics.parallelStream().forEach(s ->
+                TOPIC_CHANNEL_FACTORY.computeIfAbsent(s, topic -> new CopyOnWriteArrayList<>()).addAll(channels)));
     }
 
     /**
@@ -54,11 +51,8 @@ public class SubscribeRepository implements BaseRepository<List<String>, List<Ch
      * @param mqttTopicSubscription mqtt subscription info
      */
     public void add(final Channel channel, final List<MqttTopicSubscription> mqttTopicSubscription) {
-        CompletableFuture.runAsync(() -> mqttTopicSubscription.parallelStream().forEach(s -> {
-            List<Channel> channels = get(s.topicName());
-            channels.add(channel);
-            TOPIC_CHANNEL_FACTORY.put(s.topicName(), channels);
-        }));
+        CompletableFuture.runAsync(() -> mqttTopicSubscription.parallelStream().forEach(s ->
+                TOPIC_CHANNEL_FACTORY.computeIfAbsent(s.topicName(), topic -> new CopyOnWriteArrayList<>()).add(channel)));
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


Aias00 (dismissed):
## Review: fix: use computeIfAbsent to prevent lost subscriber updates (#6709)

Reviewed at head `40c8f0`. Correct fix for a real concurrency bug, with solid tests.

**The bug:** the old `add` ran inside `CompletableFuture.runAsync(() -> topics.parallelStream().forEach(...))`. Inside, it did `List<Channel> list = get(s); list.addAll(channels); TOPIC_CHANNEL_FACTORY.put(s, list);`. Under a parallel stream, two threads can both `get(s)`, both `addAll`, then both `put` — the second `put` overwrites the first, so subscribers are silently lost. Worse, if `get(s)` returned a different/empty instance than what's stored, the `put` could clobber the live list.

**The fix (lines 14-29):** `TOPIC_CHANNEL_FACTORY.computeIfAbsent(s, topic -> new CopyOnWriteArrayList<>()).addAll(channels)` (and the single-channel variant). `computeIfAbsent` on the `ConcurrentHashMap` (line 40) atomically gets-or-creates the list, and the list is a `CopyOnWriteArrayList` (line 30 import, created per key), so concurrent `add`/`addAll` are thread-safe and no subscriber is dropped. `remove` paths operate on the same COW list (`TOPIC_CHANNEL_FACTORY.get(topic).remove(channel)`) and remain safe.

**Tests (new `SubscribeRepositoryTest`):** four Awaitility-based tests, including two 50-subscriber concurrent-add scenarios (`concurrentAddsToSameTopicDoNotLoseSubscribers`, `concurrentAddsForTopicsDoNotLoseChannels`) that assert all 50 channels survive. This directly exercises the previously-broken path.

Approving.

Aias00 (changes requested):

Aias00 (approved):
Good concurrency fix — this is a textbook lost-update bug.

Both `add` overloads did a non-atomic read-modify-write:

```java
List<Channel> list = get(s);      // read
list.addAll(channels);            // mutate
TOPIC_CHANNEL_FACTORY.put(s, list); // write back
```

Two concurrent subscribers to the same topic could both read the same list, both append, and whichever `put` landed last would win — silently dropping the other's subscription. Because this runs inside `CompletableFuture.runAsync` and uses `parallelStream`, concurrent subscribes are the normal case rather than an edge case.

`TOPIC_CHANNEL_FACTORY.computeIfAbsent(s, topic -> new CopyOnWriteArrayList<>()).addAll(channels)` makes the get-or-create atomic (assuming `TOPIC_CHANNEL_FACTORY` is a `ConcurrentHashMap`, which `computeIfAbsent` guarantees to be) and the subsequent `add`/`addAll` safe for concurrent mutation thanks to `CopyOnWriteArrayList`. Both overloads were fixed consistently.

Tests are the valuable part: `concurrentAddsToSameTopicDoNotLoseSubscribers` and `concurrentAddsForTopicsDoNotLoseChannels` drive 50 concurrent subscribes through both entry points and assert the final size is exactly 50. That's a genuine regression test — the old code is very likely to fail it — and the `@BeforeEach` fresh `SubscribeRepository` keeps the cases independent.

Two non-blocking notes:
1. `CompletableFuture.runAsync` without a returned future means `add` is fire-and-forget and exceptions are swallowed. Pre-existing, but worth knowing the test's `await()` is what makes it observable.
2. `CopyOnWriteArrayList` is right here (subscribes are rare relative to reads), but each `add` copies the array. With a very large fan-out per topic this could get expensive — fine at ShenYu's current scale, just noting the trade-off.

PR #6948, opened 2026-08-19
fix: clear stale tag relations when updating an API with empty tags (#6518)

### LINKED ISSUE #6518: [BUG] Updating an API with an empty tag list leaves old tag relations
### Search before asking

### Apache ShenYu Component

shenyu-admin

### What happened

`ApiServiceImpl.update(...)` only refreshes tag relations when `apiDTO.getTagIds()` is not empty:

```java
if (updateRows > 0) {
    if (CollectionUtils.isNotEmpty(apiDTO.getTagIds())) {
        List<String> tagIds = apiDTO.getTagIds();
        ...
        tagRelationMapper.deleteByApiId(apiDO.getId());
        tagRelationMapper.batchInsert(tags);
    }
}
```

This means an API that already has tags cannot be updated to have no tags. If the update request sends `tagIds: []`, the condition is false, so `deleteByApiId(...)` is skipped and the old `tag_relation` rows remain in the database.

The API update succeeds, but later queries/filtering still see the previous tags.

### Expected behavior

An explicit empty tag list should clear existing API tag relations. If `null` is intended to mean "do not change tags", it should be handled separately from an empty list. The current logic treats both `null` and `[]` as "leave old tags untouched".

### How to reproduce

1. Create an API and associate one or more tags.
2. Update the same API with `tagIds` set to an empty list:

```json
{
  "id": "<api-id>",
  "tagIds": []
}
```

3. The update returns success.
4. Query the API or inspect `tag_relation` rows for that API.
5. The old tag relations are still present because `deleteByApiId(...)` only runs when `tagIds` is non-empty.

### Debug logs

_No response_

### Environment

Current `master` branch.

yykaue:
## What is this PR for?

Fixes #6518.

This PR fixes stale API tag relations when an API is updated with an explicit empty `tagIds` list.

## Changes

- Delete existing tag relations when `tagIds` is an empty list.
- Avoid invoking batch insertion for an empty list.
- Preserve the existing behavior where `null` means the API tag relations should not be changed.
- Add regression tests covering both empty and `null` tag lists.

## Verification

- `ApiServiceTest`: 8 tests passed.
- Checkstyle passed with no violations.

@Aias00 Could you please help review this PR? Thank you!


Aias00 (dismissed):
# PR #6948 Review — fix: clear stale tag relations when updating an API with empty tags (#6518)

## Verdict: APPROVE

### Scope
`shenyu-admin` `ApiServiceImpl.update`: fix stale `tag_relation` rows when an API is updated with
an explicit empty `tagIds` list.

### Verification
- Bug: original `if (CollectionUtils.isNotEmpty(apiDTO.getTagIds()))` skipped the whole block for
  an empty list, leaving stale tag relations untouched.
- Fix: guard on `Objects.nonNull(apiDTO.getTagIds())` (head file L24 imports `java.util.Objects`)
  → always `deleteByApiId(...)`; then re-insert only when non-empty. Semantics: `null` = leave
  relations unchanged (preserved), empty list = clear, non-empty = replace. Exactly the intended
  behavior.
- `tagRelationMapper.deleteByApiId`/`batchInsert` are existing methods (unchanged).
- Tests added: `testUpdateWithEmptyTagIdsClearsRelations` asserts `deleteByApiId` invoked and
  `batchInsert` never; `testUpdateWithNullTagIdsKeepsRelations` asserts neither called.

### Notes
- Correct, minimal bug fix with solid regression coverage. Approving.

dengliming (approved):

PR #6951, opened 2026-08-19
fix: fix path-variable off-by-one and guard placeholder substitution in rewrite plugin (#6882)

### LINKED ISSUE #6882: [BUG] MockPlugin NPE when httpStatusCode is null in MockHandle (unboxing null Integer to HttpStatus.valueOf)
## Description
When `rewriteHandle.getReplace()` contains `{`, the code enters the placeholder-substitution branch: `PathMatchUtils.replaceAll(replace, regex.substring(regex.indexOf("{")), rewriteUri.substring(regex.indexOf("{") + 1))`. (1) Off-by-one: `regex.indexOf("{")` is the position of `{` in the regex, which in the URI corresponds to the first character of the path-variable value (regex and URI share the same prefix up to the `{`). The `+1` skips that first character. E.g. regex=`/http/findById/{id}` (`indexOf("{")`=15), URI=`/http/findById/123`, `substring(16)`=`23` instead of `123`. (2) `StringIndexOutOfBoundsException`: if `replace` contains `{` but `regex` does not, `regex.indexOf("{")` returns -1, `regex.substring(-1)` throws. Neither failure mode is covered by tests.

## Location
- `shenyu-plugin/shenyu-plugin-rewrite/src/main/java/org/apache/shenyu/plugin/rewrite/RewritePlugin.java:77-79`

## Impact
(1) Silent wrong upstream path for every request using path-variable rewrite templates — upstream receives a truncated path segment, likely 404 or wrong resource. (2) 500 on every request when replace template has `{` but regex is a plain regex without `{`.

## Suggested fix
Change `rewriteUri.substring(rewriteHandle.getRegex().indexOf("{") + 1)` to `rewriteUri.substring(rewriteHandle.getRegex().indexOf("{"))` (remove `+1`). Guard the entire branch with a check that `regex.indexOf("{") >= 0`.

## Related existing
None — distinct from #6805 (RewriteHandle equals/hashCode omit percentage) which is about the DTO's equals/hashCode, not the rewrite logic.

juicewcode:
Fixes #6882 

## Changes

  - Remove the `+1` in `rewriteUri.substring(regex.indexOf("{"))`, so the full
    path-variable value is extracted instead of skipping its first character —
    avoids the silent truncated upstream path (`/123` → `/23`).
  - Guard the placeholder-substitution branch with `rewriteHandle.getRegex().contains("{")`
    so it is only entered when both the replace template and the regex contain `{`;
    otherwise fall back to the plain `replaceAll` branch — avoids the
    `StringIndexOutOfBoundsException` / 500 on every matching request.

## Tests

  - `shouldRewritePathVariableWithCompleteValue` — verifies the path-variable
    value is rewritten completely (`/shenyu/123`), not truncated.
  - `shouldNotThrowWhenOnlyReplaceContainsPlaceholder` — verifies no exception is
    thrown when the replace template has `{` but the regex does not.

Make sure that:


Aias00 (dismissed):
## Review: fix: fix path-variable off-by-one and guard placeholder substitution in rewrite plugin (#6882)

Approved.

### Verification
- Root cause (truncation): in the placeholder branch, `rewriteUri.substring(regex.indexOf("{") + 1)` dropped the first character after `{`, turning `/123` into `/23` when mapping `/shenyu/{id}` → `/shenyu/123`. The fix removes the `+1` so the full value is substituted → `/shenyu/123`.
- Root cause (crash): the branch was entered whenever `getReplace().contains("{")`, even if `getRegex()` had no `{`. Then `getRegex().substring(getRegex().indexOf("{"))` indexed at `-1` → `StringIndexOutOfBoundsException` on every matching request. The new guard `getReplace().contains("{") && getRegex().contains("{")` avoids this; non-placeholder templates fall back to the plain `replaceAll` branch.
- Confirmed `indexOf("{")` is evaluated against `getRegex()`, not `rewriteUri`, so with the guard in place it is always `>= 0` — no out-of-bounds.
- Confirmed `PathMatchUtils.replaceAll(path, regex, replacement)` uses `Pattern.quote(regex)` + `Matcher.quoteReplacement(replacement)`, so `{id}` is treated as a literal placeholder, not a regex. `{id}` → `/123` yields `/shenyu/123` as the test expects.
- New tests: `shouldRewritePathVariableWithCompleteValue` (asserts `/shenyu/123`) and `shouldNotThrowWhenOnlyReplaceContainsPlaceholder` (regex `/shenyu/.*`, replace `/new/{id}` → result `/new/{id}`, no exception).

### Notes
- Both bugs are fixed with clear regression coverage. No further changes needed.

Thanks!

juicewcode:
Resolved the merge conflict in RewritePlugin. The conflict was positively caused by whitespace.Please review again. Thank you！

shenyu-plugin/shenyu-plugin-rewrite/src/main/java/org/apache/shenyu/plugin/rewrite/RewritePlugin.java
```diff
@@ -76,7 +76,7 @@ protected Mono<Void> doExecute(final ServerWebExchange exchange, final ShenyuPlu
                 && ThreadLocalRandom.current().nextInt(100) < percentage) {
             rewriteUri = rewriteHandle.getReplace().contains("{") && rewriteHandle.getRegex().contains("{")
                     ? PathMatchUtils.replaceAll(rewriteHandle.getReplace(), rewriteHandle.getRegex().substring(rewriteHandle.getRegex().indexOf("{")),
-                            rewriteUri.substring(rewriteHandle.getRegex().indexOf("{")))
+                    rewriteUri.substring(rewriteHandle.getRegex().indexOf("{")))
```
yu199195, line 79: the code is same？

juicewcode: Hi, the tab change here is just from resolving a merge conflict between my branch and master. My actual code changes are in the first commit ([f4a3193](https://github.com/apache/shenyu/pull/6951/commits/f4a3193637d16251d7d6c03ab9bf7bbdf04c055a)) of this PR.It is likely that the same changes were already merged into master, so they no longer appear in the diff.

Aias00 (approved):

PR #6953, opened 2026-08-19
fix: avoid NPE when mock rule omits httpStatusCode (#6881)

### LINKED ISSUE #6881: [BUG] MockPlugin SpEL generators share a static EvaluationContext — cross-request data race on the req variable
## Description
`MockHandle.httpStatusCode` is declared as `private Integer httpStatusCode` with no default initializer (unlike `RedirectHandle` which defaults to 302). When a mock rule is saved without specifying `httpStatusCode`, Gson leaves the field null. `MockPlugin.doExecute` calls `HttpStatus.valueOf(mockHandle.getHttpStatusCode())` — `HttpStatus.valueOf(int)` receives a null `Integer`, auto-unboxing throws `NullPointerException`. Triggered on every request matching a mock rule whose handle JSON omits or nulls `httpStatusCode`.

## Location
- `shenyu-plugin/shenyu-plugin-mock/src/main/java/org/apache/shenyu/plugin/mock/MockPlugin.java:56`
- `shenyu-common/src/main/java/org/apache/shenyu/common/dto/convert/rule/MockHandle.java:27`

## Impact
Every request to a mock route with a null status code returns 500 instead of the mock response.

## Suggested fix
Default the field: `private Integer httpStatusCode = 200;` in `MockHandle`, or guard in `MockPlugin`: `HttpStatus.valueOf(Optional.ofNullable(mockHandle.getHttpStatusCode()).orElse(200))`.

## Related existing
None — distinct from #6657 (GeneralContextPlugin NPE on null cached handle) which is about cache-miss null handle, not null field within a non-null handle.

juicewcode:
Fixes #6881

 ## Changes
  - `MockPlugin.java`: guard `httpStatusCode` with
    `Optional.ofNullable(...).orElse(200)`, so a missing/null status code
    defaults to `200 OK` and the mock response is returned.
  - `MockPluginTest.java`: new unit test covering
    - null `httpStatusCode` → 200 with mock body (issue regression),
    - an explicitly specified status code (e.g. 404) being honored,
    - a missing cached handle falling through to the plugin chain.

Make sure that:


Aias00 (approved):
## Review: fix: avoid NPE when mock rule omits httpStatusCode (#6881)

Approved.

### Verification
- `MockPlugin.doExecute` now wraps `mockHandle.getHttpStatusCode()` with `Optional.ofNullable(...).orElse(200)`, so a missing/null status code defaults to `200 OK` instead of throwing `HttpStatus.valueOf(null)` NPE.
- Confirmed `MockHandle.getHttpStatusCode()` returns `Integer` (nullable) — so the original `HttpStatus.valueOf(...)` would indeed NPE on an omitted field; the guard is correct and necessary.
- Confirmed the new test references valid symbols: `MockPluginHandler.CACHED_HANDLE` (a `Supplier<CommonHandleCache<...>>`) with `cachedHandle` / `removeHandle` exist on the head branch.
- New `MockPluginTest` covers three scenarios: null `httpStatusCode` → 200 with body, explicit 404 honored, and missing cached handle falls through to the chain.

### Notes
- Clean, well-tested defensive fix.

Thanks!

PR #6954, opened 2026-08-20
[ISSUE #6855] fix: default RequestTemplate path to empty string

### LINKED ISSUE #6855: [BUG] RequestTemplate.request() concatenates url with null path — literal null segment in request URL
## Description
`SpringMvcContract.parseRequestTemplate` constructs `new RequestTemplate()` (field `path` defaults to `null`) and only calls `setPath` when `methodMapping.value().length > 0`. A method such as `@GetMapping String health()` (valid Spring MVC, maps to the class root) leaves `path == null`. `ShenyuClientInvocationHandler.buildMethodHandlerMap` then does `requestTemplate.setPath(shenyuClientFactoryBean.getPath() + requestTemplate.getPath())` — with a class-level `@ShenyuClient(path="/api")`, `"/api" + null` evaluates to `"/apinull"`. With no class path, `RequestTemplate.request()` line 82 does `this.url + this.path` → `url + null` → `"http://hostnull"`. Java string concatenation appends the literal `"null"`.

## Location
- `shenyu-sdk/shenyu-sdk-core/src/main/java/org/apache/shenyu/sdk/core/common/RequestTemplate.java:82`
- `shenyu-sdk/shenyu-sdk-spring/src/main/java/org/apache/shenyu/sdk/spring/support/SpringMvcContract.java:80-94`
- `shenyu-sdk/shenyu-sdk-spring/src/main/java/org/apache/shenyu/sdk/spring/proxy/ShenyuClientInvocationHandler.java:110-112`

## Impact
Any `@ShenyuClient` method whose `@RequestMapping`/`@GetMapping` has no `value` (root-mapped) issues requests to a corrupted URL ending in `null` (e.g. `http://gateway/apinull` or `http://gatewaynull`), silently 404'ing or hitting the wrong path. This is a common, valid Spring-MVC annotation pattern.

## Suggested fix
Default `path` to `""` in `RequestTemplate` (or guard `request()`: `this.url + (this.path == null ? "" : this.path)`); ensure `SpringMvcContract` sets `path` to `""` when no value is present.

## Related existing
None — distinct from #6632 (GW-SYNC-2, AbstractNodeDataSyncService wrong key segment, unrelated module).

yykaue:
## What this PR does

Fixes #6855.

- Defaults `RequestTemplate.path` to an empty string.
- Prevents root-mapped methods from generating URLs ending in `null`, such as `http://gatewaynull` or `/apinull`.
- Adds regression tests for requests without a path and Spring MVC root mappings.

## Verification

- SDK core and Spring tests passed.
- Related Maven reactor build and Checkstyle passed.

@Aias00, could you please help review this PR? Thank you!


Aias00 (approved):
## Review: [ISSUE #6855] fix: default RequestTemplate path to empty string

Approved.

### Verification
- `RequestTemplate.path` is now defaulted to `""` instead of `null`. The `request()` method builds the URL as `this.url + this.path` (line 82), so a root-mapped method (no path) now produces `http://gateway` instead of `http://gatewaynull`.
- Confirmed on the head branch that `request()` concatenates `url + path`; with `path = ""` the result is exactly the configured URL.
- New regression tests cover: (1) `RequestTemplateTest.testRequestWithoutPath` asserts `getUrl() == "http://gateway"`; (2) `SpringMvcContractTest.parseRootMappingRequestTemplateTest` asserts a `@GetMapping` (no path) root mapping yields `path = ""` and `HttpMethod.GET`.

### Notes
- Minimal, safe change. No behavioral regression for methods that already supply a path.
- Optional (non-blocking): consider whether an explicit `/` would be more semantically correct than `""` for root mappings, but `""` is consistent with the existing URL-building contract and the tests confirm the expected output.

Good fix — thanks!

PR #6958, opened 2026-08-20
[ISSUE #6831] Fix stale entries accumulating during common multi-data refresh

### LINKED ISSUE #6831: [BUG] AbstractNodeDataChangedListener.onCommonMultiChanged REFRESH unions instead of replacing — stale selector/rule ids accumulate (nacos/apollo/polaris)
## Description
`AbstractNodeDataChangedListener.onCommonMultiChanged`'s switch merges `REFRESH`, `MYSELF` and `default` into one branch that only publishes each changed entry and calls `putChangedMapToList`, which reads the old per-plugin id list and `addAll`s new ids not already present — it **never removes** ids no longer in the changed set and **never `delConfig`s** stale per-entry nodes. So a `REFRESH` that is supposed to *replace* a plugin's full selector/rule set instead *unions* it: stale selector/rule ids (and their config nodes) deleted elsewhere accumulate forever in the per-plugin LIST node and in the config center.

`syncAllByNamespaceId(REFRESH)` (NamespacePluginController:210, ConfigsExportImportController:141) groups selectors/rules across ALL plugins of a namespace into one REFRESH event, so every namespace resync after a deletion leaves the deleted selectors/rules live in nacos/apollo/polaris.

## Location
- `shenyu-admin-listener/shenyu-admin-listener-api/src/main/java/org/apache/shenyu/admin/listener/AbstractNodeDataChangedListener.java:228-234` (switch), `242-255` (`putChangedMapToList`); entry points `onSelectorChanged:172-180`, `onRuleChanged:183-194`, `onProxySelectorChanged:272-280`, `onDiscoveryUpstreamChanged:283-291`

## Impact
On nacos/apollo/polaris backends, after any namespace resync or config import that follows selector/rule deletions, stale selector and rule entries persist in the config center and in the per-plugin id list. Gateways re-reading the list fetch stale configs and may route to deleted selectors or apply deleted rules — silent routing drift.

## Suggested fix
Add a distinct `REFRESH` branch in `onCommonMultiChanged` that, before publishing, computes the old per-plugin id list, `removeAll`s the new ids, and `delConfig`s the remainder (mirroring `delChangedMapToList`), then publishes the new list — replace-then-add rather than add-only.

## Related existing
Distinct from SYNC-5/SYNC-6/SYNC-7 (#6780/#6781/#6782, gateway-side `shenyu-sync-data-center` discovery/proxySelector refresh no-clear). Those are gateway-side consumers; this is the admin-side publisher (nacos/apollo/polaris) — different base class and root cause (REFRESH merged into additive default vs missing refresh()-on-empty).

yykaue:
Closes #6831

This PR fixes the stale configuration issue in `AbstractNodeDataChangedListener.onCommonMultiChanged()`.

For `REFRESH` events, the listener now:

- Handles `REFRESH` separately from additive events.
- Removes configuration nodes that are no longer present in the refreshed data.
- Replaces each plugin's ID list with the latest list instead of merging it with the old one.
- Adds a regression test covering stale selector cleanup across multiple plugins.

This prevents deleted selector, rule, proxy-selector, and discovery-upstream entries from remaining in Nacos, Apollo, or Polaris.

@Aias00, could you please help review this PR? Thanks!

### DIFF
--- a/shenyu-admin-listener/shenyu-admin-listener-api/src/main/java/org/apache/shenyu/admin/listener/AbstractNodeDataChangedListener.java
+++ b/shenyu-admin-listener/shenyu-admin-listener-api/src/main/java/org/apache/shenyu/admin/listener/AbstractNodeDataChangedListener.java
@@ -226,6 +226,9 @@ public abstract class AbstractNodeDataChangedListener implements DataChangedList
                     this.delChangedMapToList(nameToIdMap, configKeyPrefix);
                     break;
                 case REFRESH:
+                    changedList.forEach(changedData -> publishConfig(configKeyPrefix + mappingKey.apply(changedData) + DefaultNodeConstants.JOIN_POINT + mappingValue.apply(changedData), changedData));
+                    this.refreshChangedMapToList(nameToIdMap, configKeyPrefix);
+                    break;
                 case MYSELF:
                 default:
                     changedList.forEach(changedData -> publishConfig(configKeyPrefix + mappingKey.apply(changedData) + DefaultNodeConstants.JOIN_POINT + mappingValue.apply(changedData), changedData));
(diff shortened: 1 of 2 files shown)


Aias00 (approved):
## Summary

Fixes #6831 — during a `REFRESH` of common multi-data (e.g. selectors), stale entries accumulated because the `REFRESH` branch did nothing (empty `break`).

- `onCommonMultiChanged` now handles `REFRESH` by re-publishing each changed item **and** calling a new `refreshChangedMapToList`, which:
  - reads the existing `*POINT_LIST` node,
  - removes the still-present ids (`oldNodeList.removeAll(listIds)`),
  - deletes the config nodes for any id no longer present,
  - republishes the list with the current ids.
- This mirrors the existing `putChangedMapToList` / `delChangedMapToList` helpers and keeps the `*_LIST` index in sync with the actual data on refresh.

## Verification

- `DefaultNodeConstants.JOIN_POINT` / `POINT_LIST` are existing constants already used by sibling methods in the same class.
- New `AbstractNodeDataChangedListenerTest.testOnSelectorChangedRefreshReplacesPluginSelectorIds` seeds two plugins with `[RETAINED, STALE]` ids, issues a `REFRESH` with `[RETAINED]` + a new `ADDED` for plugin 1, and asserts: plugin-1 list becomes `[RETAINED, ADDED]`, plugin-2 list becomes `[RETAINED]`, and both `STALE` data nodes are deleted. This is exactly the stale-entry cleanup the PR intends.

## Conclusion

Clean fix for a real stale-cache bug, with a focused regression test. Approving.

PR #6959, opened 2026-08-20
fix: avoid consuming logs after client init failure

### LINKED ISSUE #6846: [BUG] AbstractLogConsumeClient.initClient marks started even when initClient0 bailed — NPE in consume0 (Kafka/RocketMQ/Pulsar/Tencent/RabbitMQ)
## Description
`AbstractLogConsumeClient` does `this.initClient0(config); isStarted.set(true);` — unconditionally sets `started` after `initClient0` returns, even though `initClient0` returns early on invalid config without creating the underlying producer/channel. `consume` only guards `!isStarted.get()`; since `isStarted` was set true despite the failed init, `consume0` runs against a null producer/channel.

## Location
- `shenyu-plugin/shenyu-plugin-logging/shenyu-plugin-logging-common/src/main/java/org/apache/shenyu/plugin/logging/common/client/AbstractLogConsumeClient.java:75-76`
- `KafkaLogCollectClient.java:69-72` → NPE at `:128`; `RocketMQLogCollectClient.java:71-74` → `:114`; `PulsarLogCollectClient.java:58-61` → `:75`; `TencentClsLogCollectClient.java:74-77` → `:136`; `RabbitmqLogCollectClient.java:55-62` → `:95`

## Impact
After a bad/expired config refresh or transient broker failure at init time, every subsequent log batch throws inside `consume0`; clients log spam "push logs error" forever while dropping 100% of logs and never self-recover.

## Suggested fix
Set `isStarted.set(true)` only on successful `initClient0` (return a boolean / throw on failure); guard `consume0` against null producer/channel.

## Related existing
Distinct from LOG-15 (AliyunSLS NPE if producer is null but client is non-null) — this covers the Kafka/RocketMQ/Pulsar/Tencent/RabbitMQ instances of the same root cause.

hengyuss:
Fixes: #6846
 
Ensure logging clients are marked as started only after successful initialization, preventing log consumption with
uninitialized producers. Add null guards, failure cleanup, and related tests.

Make sure that:

### DIFF
--- a/shenyu-plugin/shenyu-plugin-logging/shenyu-plugin-logging-aliyun-sls/src/main/java/org/apache/shenyu/plugin/aliyun/sls/client/AliyunSlsLogCollectClient.java
+++ b/shenyu-plugin/shenyu-plugin-logging/shenyu-plugin-logging-aliyun-sls/src/main/java/org/apache/shenyu/plugin/aliyun/sls/client/AliyunSlsLogCollectClient.java
@@ -72,15 +72,16 @@ public class AliyunSlsLogCollectClient extends AbstractLogConsumeClient<AliyunLo
      * init aliyun sls client.
      *
      * @param config config
+     * @return true if the client was initialized successfully
      */
     @Override
-    public void initClient0(@NonNull final AliyunLogCollectConfig.AliyunSlsLogConfig config) {
+    public boolean initClient0(@NonNull final AliyunLogCollectConfig.AliyunSlsLogConfig config) {
         String accessId = config.getAccessId();
         String accessKey = config.getAccessKey();
         String host = config.getHost();
         if (StringUtils.isBlank(accessId) || StringUtils.isBlank(accessKey) || StringUtils.isBlank(host)) {
             LOG.error("init aliyun sls client error, please check accessId, accessKey or host");
-            return;
+            return false;
         }
         client = new Client(host, accessId, accessKey);
         // create LogStore, if you don't create logStore, shenyu will do it.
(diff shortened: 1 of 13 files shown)


Aias00 (approved):
## Summary

Fixes #6846 — logging clients were marked started (`isStarted=true`) even when `initClient0` failed, so `consume` then touched uninitialized producers/sinks and threw NPEs.

- `AbstractLogConsumeClient.initClient0` return type changed from `void` to `boolean`; `initClient` now sets `isStarted` to the init result and only registers the shutdown hook when initialization succeeded.
- All 9 concrete clients (Aliyun SLS, ClickHouse, ElasticSearch, Huawei LTS, Kafka, Pulsar, RabbitMQ, RocketMQ, Tencent CLS) updated: return `false` on blank-config / failure (with best-effort `close0()` cleanup where the producer/channel was partially created), `true` on success.
- Added null guards in `consume0` for Kafka (`producer`), Pulsar (`producer`), RabbitMQ (`channel`), RocketMQ (`producer`), Tencent CLS (`client`) — skip publishing with a warning instead of NPE.
- Pulsar `close0` restructured so `client.close()` runs even if `producer` is null.
- New `AbstractLogConsumeClientTest` covers not-started-skip, started-runs, and failed-reinit leaves client stopped.

## Verification

- **Compile safety**: grepped `extends AbstractLogConsumeClient` across the repo. The only extra match is `AbstractLogCollector`, which uses it only as a generic type bound (`T extends AbstractLogConsumeClient<?, L>`) — it does **not** override `initClient0`, so no missing override. All 9 real subclasses are in the diff and changed.
- Master `initClient` already calls `this.close()` when `isStarted.get()` before re-init; that "close previous" path is what makes `testFailedReinitializationLeavesClientStopped` assert `closeCount==1` (init#1 starts, init#2 closes it then fails). Behavior is internally consistent.
- Existing `ClickHouseLogCollectClientTest.testConsume` / `PulsarLogCollectClientTest.testConsume` changed from expecting a thrown exception to `assertDoesNotThrow` — correct now that a failed init no longer marks the client started.

## Conclusion

Correct, well-scoped fix with proper cleanup and good test coverage. Approving.

PR #6960, opened 2026-08-20
fix: handle invalid JSON fields in cryptor plugin

### LINKED ISSUE #6878: [BUG] RateLimiterPluginDataHandler leaks the old LettuceConnectionFactory when redis config is updated
## Description
`parser(String json, String fieldName)` calls `GsonUtils.getInstance().toObjectMap(json)` at line 51. The underlying deserializer returns `null` whenever the body is not a JSON object (e.g. a JSON array `[{...}]`, a primitive, or invalid JSON). `parser` never null-checks `map`: (a) non-dotted field path, line 64 `map.get(fieldName)` → NPE; (b) dotted field path, line 55 `(JsonObject) map.get(split[0])` → NPE. Even for a valid object, the dotted path has no null-safety: line 60 `jsonObject.getAsJsonObject(split[i])` returns null for a missing/non-object intermediate, and the next loop iteration NPEs; line 55 `(JsonObject) map.get(split[0])` CCEs when the configured top-level field is present but is a primitive/array. These exceptions propagate out of `MapTypeEnum.convert` before `AbstractCryptorPlugin.convert` can fall back to `fieldErrorParse`.

## Location
- `shenyu-plugin/shenyu-plugin-security/shenyu-plugin-cryptor/src/main/java/org/apache/shenyu/plugin/cryptor/utils/JsonUtil.java:50-67` (reached via `MapTypeEnum.java:54-55,75,80`; `AbstractCryptorPlugin.java:77-83`; `CryptorRequestPlugin.java:54-55`; `CryptorResponsePlugin.java:40`)

## Impact
Request side: any client can send a JSON-array body (or a body where a configured nested prefix is non-object/absent) to a cryptor-protected route and trigger an uncaught NPE/CCE → 500 with stack trace (pre-auth). Response side: a legitimate upstream list endpoint returning `[{...}]` with cryptor-response field encryption configured throws NPE → 500 on otherwise-valid responses.

## Suggested fix
In `JsonUtil.parser`, null-check `map` after `toObjectMap` and return `null` early when the body is not a JSON object; in the dotted branch, null-check each `getAsJsonObject(...)` intermediate and return `null` (let the caller's `fieldErrorParse` handle it) instead of dereferencing.

## Related existing
None — distinct from GW-MOD-3 (#6635, the shared-mutable-depth bug in `replaceJsonNode` at lines 93-131, a different method).

SouthwestAsiaFloat:
Fixes #6878

- Return `null` when the request or response body is not a JSON object.
- Validate every node in a nested field path before accessing it.
- Prevent `NullPointerException`, `ClassCastException`, and `IllegalStateException` when a configured field is missing or has an unexpected JSON type.
- Add unit tests covering non-object bodies and invalid nested paths.


Aias00 (approved):
# PR #6960 Review — fix: handle invalid JSON fields in cryptor plugin

## Verdict: APPROVE

### Scope
`JsonUtil.parser(String json, String fieldName)` in `shenyu-plugin-cryptor`. Adds null/type
safety when extracting a (possibly nested) field from a JSON body.

### Verification
- `import java.util.Objects` is already present (head file L18) — compile is fine.
- Top-level guard: `GsonUtils.getInstance().toObjectMap(json)` returns `null` for non-object
  JSON (array `"[...]"`, bare string, malformed). Now returns `null` instead of propagating.
- Nested branch rewrite:
  - First segment cast to `JsonObject` is now guarded by `instanceof` (was an unchecked cast
    that could throw `ClassCastException`).
  - Every intermediate node is checked for `null` and `isJsonObject()` before descending.
  - Final node is checked for `isJsonPrimitive()` before `getAsString()` (was
    `getAsJsonPrimitive()` which threw on objects/arrays).
  - All early-exit paths return `null` consistently.
- Non-nested `else` branch is unchanged: `map.get(fieldName) == null ? null : .toString()`.
  Callers (e.g. the `Set<String>` overload in `parser(json, fieldNames)`) already filter out
  blank/null results, so returning `null` here is safe and matches prior missing-field behavior.
- Logic traced against every new test case; all 8 negative cases resolve to `null` and both
  positive cases (`name` and `data.nested.name`) return the expected value.

### Tests
`JsonUtilTest` added (49 lines): positive + non-object body + 6 invalid-nested-path cases.
Covers the exact exception paths this PR eliminates (NPE / CCE / IllegalStateException).

### Notes
- Behavioral change: a nested path whose final node is a JSON object/array now returns `null`
  instead of throwing. In the cryptor context `null` already means "no field to process", so
  this is the intended safe fallback, not a silent data-loss.
- No public API/signature change; no new dependency.

Clean, minimal, well-tested null-safety fix. Approving.

PR #6965, opened 2026-08-21
fix: avoid NPE in IngressReconciler.getZookeeperUrl when zookeeper service has no endpoints (#6865)

### LINKED ISSUE #6865: [BUG] IngressReconciler.getZookeeperUrl NPE when zookeeper service has no Endpoints
## Description
In `getZookeeperUrl`, when the zookeeper register address is a k8s service name (not a direct IP), line 312 fetches `V1Endpoints v1Endpoints = namespace.get(zookeeperK8sIpUrl)`. Line 313 immediately calls `v1Endpoints.getSubsets()` without a null check. If the zookeeper service has no endpoints, `v1Endpoints` is null and `.getSubsets()` NPEs. The subsequent line 314 `if (Objects.isNull(subsets) || ...)` is too late.

## Location
- `shenyu-kubernetes-controller/src/main/java/org/apache/shenyu/k8s/reconciler/IngressReconciler.java:312-313`

## Impact
A Dubbo/Sofa ingress referencing a zookeeper service that is not yet ready NPEs during reconcile, blocking route configuration.

## Suggested fix
Add `if (Objects.isNull(v1Endpoints)) { ... throw/log ... }` before line 313.

## Related existing
None.

juicewcode:
Fixes #6865

 ## Summary
- For a Dubbo/Sofa ingress that uses a zookeeper service name as the register address,
  `IngressReconciler.getZookeeperUrl` fetched `V1Endpoints` from the lister. When the service has no Endpoints yet,
  `v1Endpoints` is null and `v1Endpoints.getSubsets()` threw an NPE during reconcile, blocking route configuration. This PR
  adds a null check on `v1Endpoints`: when no Endpoints are found it logs a clear message and falls through to the
  existing `ShenyuException` fallback (the same path as empty subsets), so no NPE is thrown.

 - Added `IngressReconcilerZookeeperUrlTest` with two cases — `testReconcileWithoutZookeeperEndpoints` verifies
  the graceful `ShenyuException` fallback instead of an NPE (regression); `testReconcileWithZookeeperEndpoints` verifies
  the service name is still resolved to the Pod IP (`zookeeper://127.0.0.1:2181`) and written into the dubbo plugin config
  (normal path).

Make sure that:


Aias00 (approved):

PR #6966, opened 2026-08-21
 fix: guard ingress upstream protocol index (#6864)

### LINKED ISSUE #6864: [BUG] DivideIngressParser protocol[i++] throws AIOOBE when protocol array shorter than endpoints
## Description
In `parseUpstream`, the `protocol` array is populated from `annotations.get(UPSTREAMS_PROTOCOL_ANNOTATION_KEY).split(",")` (line 307). Then for each endpoint address, line 325 does `protocol[i++]` where `i` increments per address. If the user provides fewer comma-separated protocol values than there are endpoint addresses (e.g. 2 protocols for 3 endpoints), `i` exceeds the array bounds and throws `ArrayIndexOutOfBoundsException`. The sibling `DubboIngressParser` (line 348-349) has the same `protocol[i++]` pattern.

## Location
- `shenyu-kubernetes-controller/src/main/java/org/apache/shenyu/k8s/parser/DivideIngressParser.java:325` (same at `DubboIngressParser.java:348-349`)

## Impact
An ingress with a mismatched `upstreams-protocol` annotation (fewer protocols than endpoints) crashes the reconcile, blocking route configuration.

## Suggested fix
Guard with `protocol.length > i ? protocol[i++] : "http://"` (or cycle: `protocol[i % protocol.length]`), and add a bounds check.

## Related existing
None — #6598 covers http:// hardcode in `EndpointsReconciler`; this is an AIOOBE in the parser.

juicewcode:
Fixes #6864
## Summary
 - For a Divide or Dubbo ingress whose upstreams-protocol annotation contains fewer values than endpoint addresses, the
  ingress parsers indexed the protocol array once per endpoint. When the endpoint count exceeded the protocol count,
  protocol[i++] or protocols[i++] threw an ArrayIndexOutOfBoundsException during reconcile, blocking route configuration.
  This PR adds protocol array bounds checks: Divide upstreams fall back to http://, while Dubbo upstreams fall back to
  dubbo:// when configured protocols are exhausted. It also removes the duplicate index increment in the Dubbo protocol
  expression.

 - Added regression cases to DivideIngressReconcilerTest and DubboReconcilerTest —
  testReconcileWithFewerProtocolsThanEndpoints verifies that two endpoint addresses with only one configured protocol
  reconcile successfully, preserving the configured protocol for the first upstream and applying the appropriate default
  protocol to the remaining upstream. Existing testReconcile methods remain unchanged.

Make sure that:


Aias00 (dismissed):

juicewcode:
After merging the main branch, the ingress deletion test intermittently failed because the process-wide singleton caches retained data from previous test cases. This caused multiple selectors to be deleted and triggered Mockito's TooManyActualInvocations error.
This change clears IngressCache, IngressSelectorCache, and ServiceIngressCache before each test, ensuring test isolation and preventing stale selector data from affecting verification.ocations.

Aias00 (changes requested):

juicewcode:
> REQUEST_CHANGES (PMC Aias00): this PR has merge conflicts (CONFLICTING) and cannot be merged. Please rebase onto the latest master and resolve the conflicts, then request a re-review.

Conflicts resolved. The changes in this PR are now identical to what's already in master, so there's no functional difference. Could you please take another look? Thanks!

Aias00 (approved):
Approving. The previous request-for-changes was about merge conflicts, and the branch now carries the intended content cleanly.

Two things here, both worthwhile:

**1. Singleton cache isolation in `@BeforeEach`.** `IngressCache`, `IngressSelectorCache` and `ServiceIngressCache` are process-wide singletons, so state leaked between test methods and made results order-dependent. Explicitly removing the fixture keys before each test makes these tests deterministic — and this class of bug is usually what makes "works locally, fails in CI" reports. Nice cleanup.

**2. The new tests pin the "fewer protocols than endpoints" fallback (#6864).** With one protocol configured (`https://` / `dubbo+ssl://`) and two endpoint addresses, the expected behaviour is that the first upstream gets the configured protocol and the second falls back to the plugin default (`http://` / `dubbo://`). Both assertions check that the handle contains *both* values, which is exactly the contract: a configured protocol must not be silently applied to (or dropped from) the remaining endpoints.

Verified the assertions match the production logic — `DivideIngressParser` guards with `protocol == null || i >= protocol.length ? "http://" : protocol[i]`, so index 1 falls back as the tests expect. The Dubbo side has the analogous default (`dubbo://`).

Note this is a test-only PR and depends on that fallback already existing in the parser; it doesn't change production behaviour, which is the right scope for a regression guard.

Non-blocking: `DubboReconcilerTest` now duplicates the same `testReconcileWithFewerProtocolsThanEndpoints` shape as the divide variant. If more plugin types gain this behaviour, a shared assertion helper would keep them in sync.

PR #6979, opened 2026-08-24
[ISSUE #6842] Handle missing data in HTTP long-polling responses

### LINKED ISSUE #6842: [BUG] HttpSyncDataService.doLongPolling NPE on missing/null data key in admin response
## Description
`HttpSyncDataService.doLongPolling` does `JsonElement element = responseFromServer.get("data")` which returns null when the `"data"` key is absent (Gson `JsonObject.get` returns null for missing keys, distinct from `JsonNull`). The next line `element.isJsonNull()` dereferences null → NPE. If the response body is empty/`"null"`, `fromJson(...)` returns null at `:239` and `responseFromServer.get("data")` NPEs on the null `JsonObject`.

## Location
- `shenyu-sync-data-center/shenyu-sync-data-http/src/main/java/org/apache/shenyu/sync/data/http/HttpSyncDataService.java:239-244`

## Impact
A non-standard admin response (e.g. `{"code":500,"message":"error"}` without `"data"`) crashes the long-polling thread with NPE; retries hit the same NPE for 10 attempts (50 s) then sleep 5 min, stalling config sync.

## Suggested fix
Null-check `responseFromServer` and `element` before use; treat missing/null data as "no change".

yykaue:
## What does this PR do?

This PR fixes the null handling in `HttpSyncDataService.doLongPolling()`.

HTTP long-polling responses without configuration change data are now treated as “no change” instead of causing an exception.

The following response cases are handled:

- The `data` field is missing.
- The `data` field is `null`.
- The root response is `null`.
- The response body is empty.

Valid `data` arrays continue to use the existing processing flow.

Closes #6842

@Aias00, could you please help review this PR? Thanks!


Aias00 (commented):
This now treats admin error payloads as successful no-change responses. The admin exception handlers return `ShenyuAdminResult.error(...)` bodies such as `{"code":500,"message":"..."}` without setting a non-200 HTTP status, and this branch returns as soon as `data` is missing/null before checking the result code. In `HttpLongPollingTask`, only exceptions from `doLongPolling` trigger the retry/backoff path, so a listener-side error body can now be swallowed as no change and the loop immediately continues. Could we check `code` first and throw/retry when it is present and not `CommonErrorCode.SUCCESSFUL`, then only ignore missing/null `data` for successful responses?

Aias00 (approved):
## Summary

Fixes a NullPointerException in `HttpSyncDataService.doLongPolling` when the long-polling listener response is missing the `data` field.

## What is correct

- The previous code did `JsonObject element = responseFromServer.get("data"); if (element.isJsonNull())` — when the `data` key is **absent**, Gson's `get` returns `null`, so `element.isJsonNull()` threw an NPE.
- The fix parses as `JsonElement`, guards a null/JSON-null root (`Objects.isNull(responseFromServer) || responseFromServer.isJsonNull()`), and separately guards a missing/null `data` before `getAsJsonArray()`. It also safely handles empty/non-JSON bodies (`"null"`, `""`) instead of throwing.
- Good regression test `testDoLongPollingIgnoresResponseWithoutData` covers `{"code":500,"message":"error"}`, `{"data":null}`, `"null"`, and `""`, asserting none throw.

## Verdict

Approve. No blocking issues.

## Non-blocking notes

- By design (per the added test), any response lacking `data` is treated as "no change" and the method returns — including admin error bodies like `{"code":500,"message":"..."}` (which carry no `data` and are served with HTTP 200). So a listener-endpoint error from admin is swallowed and the long-poll loop simply continues; it will not trigger the retry/backoff path. This is acceptable for long-polling resilience, but be aware admin-side errors on this endpoint won't surface as failures.
- `responseFromServer.getAsJsonObject()` would still throw `IllegalStateException` if the root were a JSON array or primitive. That's pre-existing behavior and an unlikely response shape, so not blocking.

Issue #6873, opened 2026-08-04
[BUG] discovery selectBySelectorNameAndPluginName references non-existent s.name column


Aias00:
## Description
`BaseResultMap` maps id/date_created/date_updated/instance_ip/instance_port/instance_type/instance_info/namespace_id but NOT `instance_state`, even though `Base_Column_List` (line 41) includes `instance_state` and `insert`/`updateById` write it. Every resultMap-based SELECT (`selectAll`, `findAllByNamespaceId`, `selectByQuery`) returns `instanceState = null` for rows that have a non-null state. (`selectById`/`selectOneByQuery` use `resultType` auto-mapping and are unaffected.) Also `Base_Column_List` lists `date_created`/`date_updated` twice (lines 35-36, 43-44) — a copy-paste defect.

## Location
- `shenyu-admin/src/main/resources/mappers/instance-info-sqlmap.xml:21-30,35-36,43-44`

## Impact
Instance-state reads via the resultMap path silently return null state; this is the upstream root cause of the `InstanceCheckService` null-state autoboxing NPEs (N37).

## Suggested fix
Add `<result column="instance_state" jdbcType="INTEGER" property="instanceState"/>` to `BaseResultMap`; remove the duplicate date columns from `Base_Column_List`.

## Related existing
Distinct from N37 (#6706, `InstanceCheckService` doCheck/collectStateData NPE on null instanceState) — N37 is the service-level autoboxing symptom; this is the mapper-level root cause (missing resultMap column). Fixing the resultMap populates the field and prevents the N37 unboxing for the resultMap-read path.

PR #6981, opened 2026-08-24
fix: add pre-check on expireTime in AccessTokenManager

### LINKED ISSUE #6841: [BUG] AccessTokenManager NPE on missing expiredTime — kills scheduled token refresh permanently
## Description
`AccessTokenManager` does `this.tokenExpiredTime = (long) tokenMap.get(Constants.ADMIN_RESULT_EXPIRED_TIME);` — unboxing a null `Long` when the admin login response omits `expiredTime`. The enclosing catch is only `IOException`, so the NPE propagates out of the `ScheduledExecutorService.scheduleWithFixedDelay` task. For `ScheduledExecutorService`, an uncaught `RuntimeException` from a scheduled task suppresses all subsequent executions — no further token refresh attempts are ever made.

## Location
- `shenyu-sync-data-center/shenyu-sync-data-http/src/main/java/org/apache/shenyu/sync/data/http/AccessTokenManager.java:139`

## Impact
The access token remains null/stale; all subsequent long-polling requests fail with 401, permanently breaking HTTP sync until gateway restart.

## Suggested fix
Null-check before unboxing; return false (so the schedule continues) on a malformed response; widen the catch to `Exception`.

hengyuss:
Fixes: #6841 

Make sure that:


Aias00 (commented):
This still leaves the scheduled refresh vulnerable to unchecked parsing failures. The new null check handles a missing `expiredTime`, but `doLogin` only catches `IOException`; for example a successful response with `data: null` makes `tokenMap` null before this check, and a response where `expiredTime` is encoded as a string passes the null check but throws at the `(long) expiredTime` cast. Since `start()` runs `login()` from `scheduleWithFixedDelay`, an uncaught runtime exception can still suppress future refresh attempts. Could you validate the login payload and return `false` for malformed responses, and/or guard the scheduled task so refresh continues after one bad response?

Aias00 (changes requested):
## Summary

Adds a pre-check on `expiredTime` in `AccessTokenManager.doLogin`, addressing one of the malformed-response failure modes I raised earlier.

## What is correct

- The new guard (lines 17-21) checks `Objects.isNull(expiredTime)` right after `tokenMap = convertToMap(tokenJson)` and returns `false` (with a warn log) instead of dereferencing a null. The added test `testDoLoginFailureWhenExpiredTimeMissing` verifies `doLogin` returns `false` and `accessToken` stays null. This is a real improvement and the test is solid.

## Blocking issue

The core availability concern from my earlier comment is **only partially addressed**. `start()` schedules refresh via `scheduleWithFixedDelay(() -> this.login(servers), 5000, 5000, ...)` (line 160). An *uncaught* exception escaping `doLogin` permanently suppresses all future refresh attempts — `ScheduledExecutorService` cancels the task on throw, so the gateway silently loses its admin token and stops syncing.

The new `expiredTime == null` guard closes one path, but two malformed-response paths still throw inside `doLogin` before the method returns:

1. **`data: null` / missing `data`** → `tokenJson = "null"` → `GsonUtils.convertToMap("null")` returns `null` → `tokenMap.get(...)` → **NullPointerException**.
2. **`expiredTime` returned as a JSON string** (e.g. `"1800"`) → passes the null check → `(long) expiredTime` → **ClassCastException**.

Either case still kills the refresh loop, which is the exact availability bug we discussed.

## Required changes (then I'll approve)

Harden `doLogin` so a malformed payload returns `false` instead of throwing, e.g.:

```java
Map<String, Object> tokenMap = GsonUtils.getInstance().convertToMap(tokenJson);
if (Objects.isNull(tokenMap)) {
    LOG.warn("get token from server : [{}] returned null data", server);
    return false;
}
Object expiredTime = tokenMap.get(Constants.ADMIN_RESULT_EXPIRED_TIME);
if (Objects.isNull(expiredTime) || !(expiredTime instanceof Number)) {
    LOG.warn("get token from server : [{}] missing/invalid expired time", server);
    return false;
}
```

(You may also guard `token` similarly if appropriate.)

As defense-in-depth, consider wrapping the scheduler lambda body in `start()` in a try/catch that logs and returns, so a future unexpected exception can never suppress the whole refresh loop.

Also add tests for the `data:null` and string-`expiredTime` cases to lock in the fix (the existing test only covers null `expiredTime`).

## Verdict

REQUEST_CHANGES — good start, but the uncaught-exception-suppresses-scheduling failure mode is not fully closed. Add the null/type guards above and I'll approve.

Aias00 (approved):
## Re-review: fix: add pre-check on expireTime in AccessTokenManager (#6873)

Re-reviewed after the author pushed `fix: handle malformed access token responses` (head `39a662d`). My previous REQUEST_CHANGES raised three points; all are now addressed:

**1. `resultMap` / `tokenMap` null guards (lines 18, 28-31)**
`doLogin` now returns `false` early when `resultMap == null` or `tokenMap == null`, instead of dereferencing a null map. The `data:null` body (-> `tokenMap` resolves to `null`) no longer throws NPE.

**2. Type-safe extraction of `token` / `expiredTime` (lines 32-43)**
- `token` must be a non-blank `String` (line 34), else `return false`.
- `expiredTime` must be a `Number` (line 38-41); otherwise `return false`. This eliminates the `ClassCastException` that previously occurred when the server returned `expiredTime` as a JSON string.
- `this.tokenExpiredTime = ((Number) expiredTime).longValue();` is now a safe unbox.

**3. Refresh loop resilience (lines 52-58) — the core availability gap**
`start()` wraps the scheduled `this.login(servers)` in `try { ... } catch (Exception e) { LOG.error(...) }`. Previously an uncaught exception in `doLogin` permanently cancelled the `scheduleWithFixedDelay` refresh task, so a single malformed response would silently stop all future token renewal. The refresh loop is now preserved across failures.

**Tests (lines 70-148):** three new cases cover `expiredTime` missing, `data:null`, and `expiredTime` as a string — exactly the branches from the request.

Imports verified: `java.util.Objects` (line 43) and `org.apache.commons.lang3.StringUtils` (line 28) both present, no duplicate.

All concerns resolved. Approving.

PR #6983, opened 2026-08-24
fix: enforce per-channel MQTT connection state guard

### LINKED ISSUE #6638: [BUG] Connection-state guard is dead; unconnected clients can PUBLISH/SUBSCRIBE
- severity: High
- files: `shenyu-protocol/shenyu-protocol-mqtt/src/main/java/org/apache/shenyu/protocol/mqtt/MqttFactory.java:34`; `MessageType.java:31-47`; `Publish.java:47-49`; `Subscribe.java:56-59`; `Unsubscribe.java:41-43`
- description: `MqttFactory` constructs a fresh `MessageType` per inbound message; `Connect.connect()` calls `setConnected(true)` on the throwaway `Connect` instance, so `isConnected()` never propagates to other handler instances. Combined with inverted guards (`if (isConnected()) return;`), the "reject operations from unauthenticated clients" enforcement never fires — a client that never sent CONNECT can PUBLISH/SUBSCRIBE/UNSUBSCRIBE. Even if state persisted, `if (isConnected()) return;` would block connected clients.
- impact: No connection-state enforcement; pre-auth PUBLISH/SUBSCRIBE allowed.
- suggested_fix: Track connection state per-channel (Channel attribute or ChannelRepository); invert guards to `if (!isConnected()) { ...; return; }`.
- confidence: High
- related_existing: none

---

wy471x:
Make sure that:

### Changes:
1. `MessageType.java` — connection state moved from a per-instance `volatile boolean` (dead: `MqttFactory` builds a fresh `MessageType` per inbound message, so the state set by `Connect.connect()` never propagated to other handlers) to a per-channel `AttributeKey<Boolean>`; `isConnected(Channel)` / `setConnected(Channel, boolean)` now read/write the channel attribute.
2. `Publish.java`, `Unsubscribe.java`, `PingReq.java` — guards inverted to `if (!isConnected(ctx.channel()))`; like `Subscribe`, the channel is now closed (`FIRE_EXCEPTION_ON_FAILURE`) instead of silently dropping the packet.
3. `Subscribe.java` — inverted the dead `if (isConnected())` guard.
4. `Connect.java` — duplicate CONNECT on the same channel is rejected by closing the connection without sending a CONNACK ([MQTT-3.1.0-2]).

### Test Cases:
1. `ConnectTest` — migrated from Mockito mocks to a real `EmbeddedChannel` (per-channel attributes require a real channel); added `duplicateConnectIsRejected` (second CONNECT receives no CONNACK and closes the channel).
2. `PublishTest` — added `publishBeforeConnectClosesChannel` and the end-to-end `publishAfterConnectOnSameChannelIsAccepted` (real `Connect.connect()` followed by `Publish.publish()` on the same channel).
3. New `PingReqTest` — `pingReqBeforeConnectClosesChannel`, `pingReqAfterConnectSendsPingResp`.
4. New `UnsubscribeTest` — `unsubscribeBeforeConnectClosesChannel`.

## Verification

- `shenyu-protocol-mqtt` module: 13 tests passed (ConnectTest 5, PublishTest 5, PingReqTest 2, UnsubscribeTest 1), JDK 21.
- Checkstyle and RAT (license header) checks passed.

close #6638


Aias00 (commented):
The duplicate-CONNECT path looks like it needs cleanup before closing the channel. A successful first CONNECT registers the channel in \, but this new early return only calls \. I only see repository cleanup in the explicit DISCONNECT flow, and there does not appear to be centralized \ cleanup for MQTT channels, so a client that sends CONNECT twice can leave a closed channel referenced in \. Could you remove the channel from \ before closing here, or add centralized channel-inactive cleanup for all server/client close paths?

Aias00:
Correction to my previous review comment, where inline code formatting was stripped by my local shell:

The duplicate-CONNECT path looks like it needs cleanup before closing the channel. A successful first CONNECT registers the channel in `ChannelRepository`, but this new early return only calls `ctx.close()`. I only see repository cleanup in the explicit DISCONNECT flow, and there does not appear to be centralized `channelInactive` cleanup for MQTT channels, so a client that sends CONNECT twice can leave a closed channel referenced in `ChannelRepository`. Could you remove the channel from `ChannelRepository` before closing here, or add centralized channel-inactive cleanup for all server/client close paths?

wy471x:
> Correction to my previous review comment, where inline code formatting was stripped by my local shell:
> 
> The duplicate-CONNECT path looks like it needs cleanup before closing the channel. A successful first CONNECT registers the channel in `ChannelRepository`, but this new early return only calls `ctx.close()`. I only see repository cleanup in the explicit DISCONNECT flow, and there does not appear to be centralized `channelInactive` cleanup for MQTT channels, so a client that sends CONNECT twice can leave a closed channel referenced in `ChannelRepository`. Could you remove the channel from `ChannelRepository` before closing here, or add centralized channel-inactive cleanup for all server/client close paths?

These issues has been fixed! Please take a look when you hava time, thank you! @Aias00

Aias00 (approved):
## Summary

Fixes MQTT connection-state tracking and channel cleanup. The root cause: `MessageType` stored connection state in a per-instance `volatile boolean connected`, but `MqttFactory` builds a fresh `MessageType` (Connect/Publish/Subscribe/...) per inbound message, so `setConnected(true)` in `Connect.connect()` never propagated to the other handlers — making every `isConnected()` guard dead (always `false`). This PR moves the state to a per-channel `AttributeKey<Boolean>` shared by all handlers on the channel, inverts the guards to the correct MQTT semantics, rejects duplicate CONNECTs, and adds centralized channel-cleanup on close.

## What is right (and what was fixed from the earlier comment)

My previous review pointed out that the duplicate-CONNECT early-return closed the channel without removing it from `ChannelRepository`, leaving a dangling reference. The PR resolves this via the **centralized** option I suggested:

- `MqttTransportHandler.channelInactive()` now calls `Singleton.INST.get(ChannelRepository.class).remove(ctx.channel())` and then `fireChannelInactive()`. This covers **all** close paths (duplicate CONNECT, abrupt close, DISCONNECT), not just one.

Other improvements:
- **Root cause fix:** connection state is now `AttributeKey<Boolean>` on the channel, so `isConnected(channel)` / `setConnected(channel, boolean)` are consistent across `Connect`/`Publish`/`Subscribe`/`Unsubscribe`/`PingReq`.
- **Correct MQTT semantics:** `Publish`/`Subscribe`/`Unsubscribe`/`PingReq` now close the channel when a message arrives before CONNECT (`if (!isConnected(channel)) { close; return; }`), instead of the previously-inverted (and dead) guard.
- **Duplicate CONNECT** rejected by closing the connection without a CONNACK, per [MQTT-3.1.0-2].
- `ChannelRepository.add` made synchronous (removed `CompletableFuture.runAsync`), removing a registration race.
- Tests migrated to real `EmbeddedChannel` (required for per-channel attributes) and added: `duplicateConnectCleansUpChannelRepository`, `abruptChannelCloseCleansUpChannelRepository`, pre-connect-close tests for Publish/PingReq/Unsubscribe, and an end-to-end connect+publish. 13 tests pass.

## Verification

- Confirmed `ChannelRepository` is a `BaseRepository` subtype and `MqttBootstrapServer.initRepositories()` (invoked at startup, line 50) uses Reflections over `org.apache.shenyu.protocol.mqtt.repositories` to auto-register every `BaseRepository` as a Singleton. Therefore `Singleton.INST.get(ChannelRepository.class)` is **non-null in production**.
- `Disconnect.java` already uses the exact same `Singleton.INST.get(ChannelRepository.class).remove(channel)` idiom (line 47), confirming this is the established, safe pattern — not a new unguarded dependency.
- `ChannelRepository.remove(Channel)` and `get(Channel)` both exist.

## Non-blocking suggestions

- The per-channel `CONNECTED` attribute is not reset to `false` on close, but the channel is dead at that point so it is harmless.

## Conclusion

Approving. The earlier dangling-reference concern is fully resolved by the centralized `channelInactive` cleanup, and the root-cause (per-instance vs per-channel state) fix is sound.

PR #6985, opened 2026-08-24
fix: return 202 empty body for JSON-RPC notifications and reset captured response in Streamable HTTP transport

### LINKED ISSUE #6834: [BUG] MCP Streamable-HTTP returns stale/fabricated JSON-RPC response for notifications
## Description
`processWithExistingSession` unconditionally chains `.then(waitForTransportResponse(transport, sessionId, messageId))` for every incoming message, with no detection of `JSONRPCNotification` (messages with no `id`). `waitForTransportResponse` either returns `transport.getLastSentMessage()` (the previous request's response still held, since `resetCapturedMessage` is never called on the existing-session path) when `isResponseReady()` is true, or a fabricated `createJsonRpcResponse(messageId, {})` when false. Notifications must return HTTP 202 with no body.

## Location
- `shenyu-plugin/shenyu-plugin-mcp-server/src/main/java/org/apache/shenyu/plugin/mcp/server/transport/ShenyuStreamableHttpServerTransportProvider.java:515-527`, `792-806`

## Impact
After a `tools/call`/`tools/list` sets `lastSentMessage`, any subsequent notification (`notifications/initialized`, `notifications/cancelled`) sent with the same `Mcp-Session-Id` is answered with HTTP 200 carrying the stale previous JSON-RPC response (wrong `id`), causing client mis-correlation. A notification sent before any request gets a fabricated `{"jsonrpc":"2.0","result":{}}` with null id.

## Suggested fix
In `processWithExistingSession`, detect `message` being a `JSONRPCNotification` (no `id`) and return a 202/empty-body result without invoking `waitForTransportResponse`; call `resetCapturedMessage()` after each completed response on the existing-session path.

## Related existing
Distinct from MCP-01..05 (#6640-#6644, response-decorator body/leak issues). This is the transport-provider notification path.

wy471x:
Make sure that:

## Summary

Fixes #6834.

MCP Streamable HTTP notifications (`notifications/initialized`, `notifications/cancelled`) are JSON-RPC messages without an `id` and must be acknowledged with HTTP 202 and an empty body per the Streamable HTTP spec. Previously `processWithExistingSession` unconditionally chained `waitForTransportResponse` for every incoming message, so a notification on an existing session either replayed the stale response captured by a previous request (HTTP 200 with the wrong `id`, causing client mis-correlation), or received a fabricated `{"jsonrpc":"2.0","result":{}}` response.

### Changes:
1. `ShenyuStreamableHttpServerTransportProvider.java` — `handleUnifiedEndpoint` (POST branch): when the `MessageHandlingResult` has a null response body (the notification path), return the 202 `ServerResponse.BodyBuilder` as-is with an empty body, instead of unconditionally setting the JSON content type and writing the body.
2. `ShenyuStreamableHttpServerTransportProvider.java` — `processWithExistingSession`: detect `McpSchema.JSONRPCNotification` messages and short-circuit — still dispatch to `session.handle(...)` so the MCP framework updates session state, but return `202 Accepted` with a null body without invoking `waitForTransportResponse`. Errors on the notification path are mapped to a 500 JSON-RPC error.
3. `ShenyuStreamableHttpServerTransportProvider.java` — `processWithExistingSession` (request path): call `transport.resetCapturedMessage()` after each completed message so a subsequent message on the same session can never observe a stale response from a previous request.

### Test Cases:
1. `ShenyuStreamableHttpServerTransportProviderTest` — `testNotificationWithExistingSessionReturnsAcceptedEmptyBody`: a `notifications/cancelled` notification sent on an existing session returns HTTP 202 with an empty body and the session id header.
2. `ShenyuStreamableHttpServerTransportProviderTest` — `testNotificationAfterRequestDoesNotReplayStaleResponse`: full handshake (initialize → `notifications/initialized` → `tools/list`), then a `notifications/cancelled` on the same session returns 202 with an empty body instead of replaying the stale `tools/list` response.

## Verification

- `shenyu-plugin-mcp-server` module: `ShenyuStreamableHttpServerTransportProviderTest` 4 tests passed (JDK 21).
- Checkstyle and RAT (license header) checks passed (`mvn -pl shenyu-plugin/shenyu-plugin-mcp-server -am validate`).

close #6834


Aias00 (commented):
`processWithExistingSession` now uses a `null` response body to represent the notification / HTTP 202 empty-body case, but only `handleUnifiedEndpoint` was updated to honor that contract. The gateway plugin path calls `handleMessageEndpoint(exchange, request)` directly and `processStreamableHttpResult` still serializes every result with `getResponseBodyAsJson()`, which turns this `null` into the literal response body `null` and writes it with the configured response headers. That means Streamable HTTP notifications routed through the real `McpServerPlugin` path still do not get an empty 202 response. Could you make `processStreamableHttpResult` return an empty response when `result.getResponseBody() == null`, and add a regression test through the plugin path?

Aias00 (approved):
## Summary

Fixes MCP Streamable HTTP JSON-RPC **notification** handling: notifications (`notifications/initialized`, `notifications/cancelled`) are messages without an `id` and must be acknowledged with **HTTP 202 + empty body** per the Streamable HTTP spec. Previously the gateway either replayed a stale response captured by a previous request (with the wrong `id`) or synthesized a `{"jsonrpc":"2.0","result":{}}` body.

## What is right (and what was fixed from the earlier comment)

My previous review asked for two things; both are now addressed:

1. **Empty body through the real plugin path.** `processStreamableHttpResult` now early-returns `exchange.getResponse().setComplete()` when `result.getResponseBody() == null`, so a notification routed through `McpServerPlugin` returns 202 with an empty body instead of serializing `null` into the response. A regression test `testStreamableHttpNotificationReturnsEmptyBodyViaPluginPath` exercises exactly this path.
2. **Empty body through the transport endpoint.** `handleUnifiedEndpoint` returns the 202 `ServerResponse.BodyBuilder` as-is (empty body) when the response body is `null`.

Additional correctness improvements:
- `processWithExistingSession` detects `McpSchema.JSONRPCNotification` and returns `new MessageHandlingResult(ACCEPTED, null, sessionId)` **without** invoking `waitForTransportResponse(...)`, so it still dispatches to `session.handle(...)` to update framework state but never produces a JSON-RPC response. Notification errors are mapped to a 500 JSON-RPC error.
- `resetCapturedMessage()` is now called after each completed (non-notification) message, preventing a subsequent message on the same session from observing a stale captured response from a previous request.
- `Content-Type: application/json` is moved into the body-writing branch, so the empty 202 path no longer sets it (correct for an empty body).

## Verification

- `MessageHandlingResult.getResponseBody()` exists (returns `Object`), so `Objects.isNull(result.getResponseBody())` compiles and is correct.
- `transport.resetCapturedMessage()` is already used elsewhere in the same class (lines 434/480/549) and `getSessionTransport(...)` returns `StreamableHttpSessionTransport`, so the new call compiles and is consistent.
- Added tests cover: notification → 202 empty body via the plugin path, notification on an existing session → 202 empty body, and a notification after a business request → 202 empty body (no stale `tools/list` replay).

## Non-blocking suggestions

- Consider also asserting the `Mcp-Session-Id` header presence on the empty 202 responses in the transport-provider tests (currently only the plugin-path test checks it), but coverage is already solid.

## Conclusion

Approving. The earlier concern is fully resolved and the fix is well-tested.

PR #6986, opened 2026-08-25
fix: fix polaris listener cache removal

### LINKED ISSUE #6840: [BUG] PolarisSyncDataService.doRemoveListener doesn't remove watchCache entry — delete+recreate misses data
## Description
`PolarisSyncDataService.doRemoveListener` removes the listener from the `ConfigFile` but does **not** remove the entry from `watchCache` (contrast `NacosSyncDataService.doRemoveListener:85` which calls `watchCache.remove(key)`). When `AbstractNodeDataSyncService.unCacheSelectorData/unCacheRuleData/etc` call `removeListener → doRemoveListener`, the stale entry remains. When the same key is re-created, `getServiceConfig` checks `watchCache.containsKey(key)` → true → returns null at `:77-78` without registering a new listener, so the re-created item is never watched/fetched.

## Location
- `shenyu-sync-data-center/shenyu-sync-data-polaris/src/main/java/org/apache/shenyu/sync/data/polaris/PolarisSyncDataService.java:105-111`

## Impact
After any delete+recreate of the same config key, the gateway misses the new data; routes stay stale/missing until full restart.

## Suggested fix
Add `watchCache.remove(removeKey)` in `doRemoveListener`.

## Related existing
None.

hengyuss:
Fixes: #6840

Make sure that:


Aias00 (approved):

PR #6988, opened 2026-08-26
fix: fix ingress deletion cleanup NPE

### LINKED ISSUE #6836: [BUG] IngressReconciler dereferences v1Ingress before null check — NPE on every ingress deletion (delete cleanup is dead code)
## Description
In `reconcile()`, line 134 fetches `v1Ingress` from the ingress lister. When an ingress is deleted from k8s, the lister no longer has it, so `v1Ingress` is null. Line 136 (`Map<String,String> annotations = v1Ingress.getMetadata().getAnnotations();`) dereferences `v1Ingress` **before** the null check on line 138 (`if (Objects.isNull(v1Ingress))`). The NPE propagates up and the entire delete-cleanup block (lines 138–161: selector/rule/SSL cleanup, cache invalidation) is never reached.

## Location
- `shenyu-kubernetes-controller/src/main/java/org/apache/shenyu/k8s/reconciler/IngressReconciler.java:134-161`

## Impact
Deleting a ShenYu ingress leaves orphaned selectors, rules, metadata, SSL config, and service-ingress cache entries in the gateway. The ingress appears gone from k8s but the gateway still routes to the old upstreams. The delete path is effectively dead code.

## Suggested fix
Move line 136–137 (`annotations` fetch + `enablePluginsBasedOnAnnotations`) to after the `Objects.isNull(v1Ingress)` check — they should only run when `v1Ingress` is non-null (the create/update path).

## Related existing
Distinct from #6598 (http:// hardcode) and GOV-T6 (#6679, parser/reconciler zero-test). GOV-T6 notes `IngressReconciler` has no tests; this NPE is the root defect those missing tests would have caught.

hengyuss:
Fixes: #6836 

Make sure that:


Aias00 (approved):
## Summary

Fixes a `NullPointerException` that occurred when an Ingress is deleted: `IngressReconciler.reconcile()` dereferenced `v1Ingress.getMetadata().getAnnotations()` **before** the `Objects.isNull(v1Ingress)` null-check, so a deleted ingress (where `v1Ingress` is `null`) crashed before reaching the deletion cleanup path.

## What is right

- The PR moves the `annotations` extraction and the `enablePluginsBasedOnAnnotations(...)` call to **after** the `if (Objects.isNull(v1Ingress)) { ... return new Result(false); }` block. Now the deletion path executes cleanly and `enablePluginsBasedOnAnnotations` only runs for an existing ingress (where it is actually meaningful).
- This is a minimal, correctly-scoped change that removes the only early dereference of `v1Ingress`.

## Verification

- Read the full `reconcile()` method: the null-check block returns `new Result(false)` before the `Map<String, String> annotations = v1Ingress.getMetadata().getAnnotations();` line, so for a deleted ingress the method now returns normally after cleanup. No other dereference of `v1Ingress` precedes the null-check.
- The new `testReconcileDeletedIngress` simulates an ingress deletion (informer returns `null`) and asserts `reconcile(...)` does **not** throw, returns `new Result(false)`, and that the selector/rule caches are cleaned up.

## Non-blocking suggestions

- None.

## Conclusion

Approving. Correct root-cause fix with a regression test.

PR #6989, opened 2026-08-27
fix: use thread-safe maps for registry instance watchers

### LINKED ISSUE #6835: [BUG] registry impls (consul/etcd/zookeeper/apollo) use non-thread-safe HashMap for watcherInstanceRegisterMap shared by caller + async watcher
## Description
All four `shenyu-registry` instance-register implementations (consul/etcd/zookeeper/apollo) store the cached instance list (`watcherInstanceRegisterMap`) in a plain `java.util.HashMap`. This map is written by an async watcher/callback thread (Consul scheduled executor, etcd `Watch.listener`, Curator `CuratorWatcher`, Apollo `ConfigChangeListener`) while concurrently read and written by the caller thread in `selectInstances` (`containsKey`/`get`/`put`). The Zookeeper impl also races `nodeDataMap` between `persistInstance` (caller) and the `ConnectionState.RECONNECTED` listener; Apollo additionally mutates a local `HashMap childrenList` from the config thread. The Nacos impl already uses `ConcurrentHashMap`/`ConcurrentMap`.

## Location
- `shenyu-registry/shenyu-registry-consul/src/main/java/org/apache/shenyu/registry/consul/ConsulInstanceRegisterRepository.java:76,90,92,160-196` (also `consulIndexes`:76, `watchSelectKeySet`:92)
- `shenyu-registry/shenyu-registry-etcd/.../EtcdInstanceRegisterRepository.java:60,98,118-121`
- `shenyu-registry/shenyu-registry-zookeeper/.../ZookeeperInstanceRegisterRepository.java:67,71,100,122,145-167`
- `shenyu-registry/shenyu-registry-apollo/.../ApolloInstanceRegisterRepository.java:65,128-163`

## Impact
Under normal multi-path operation, concurrent access corrupts the HashMap, loses instance updates (watcher misses instances or serves stale lists), produces inconsistent discovery state, or causes infinite-loop/CME-style hangs on the `selectInstances` path → wrong/missing upstream routing.

## Suggested fix
Replace all `new HashMap<>()` / `new HashSet<>()` field declarations with `Maps.newConcurrentMap()` / `ConcurrentHashMap.newKeySet()` (as the Nacos impl already does), or guard all access with a single `synchronized` lock. For the Apollo local `childrenList`, use `ConcurrentHashMap` or copy before passing to the listener.

## Related existing
Distinct from SYNC-8 (#6783, `ConsulSyncDataService` HashMap) — that covers `shenyu-sync-data-center/shenyu-sync-data-consul` (gateway-side sync service), not `shenyu-registry-consul` (instance register repository). The Nacos registry is already safe.

hengyuss:
Fixes: #6835 

Make sure that:


Aias00 (approved):
## Summary

Replaces the non-thread-safe `HashMap` / `HashSet` instances used for registry instance watcher state with `ConcurrentHashMap` / `ConcurrentHashMap.newKeySet()` across the Apollo, Consul, Etcd and Zookeeper registry implementations, and refactors the `containsKey` + `get` check-then-act pattern into a single `get` + `Objects.nonNull`.

## What is right

- These maps (`watcherInstanceRegisterMap`, `consulIndexes`, `watchSelectKeySet`, `nodeDataMap`, etc.) are written from registry watcher callback threads (Curator cache / Etcd watch / Consul watch) and read from `selectInstances(...)` on request threads. With `HashMap`/`HashSet` this is a genuine data race that can cause `ConcurrentModificationException` or lost updates under concurrent registration/discovery. `ConcurrentHashMap` fixes it correctly.
- Replacing `containsKey(k); get(k)` with `get(k); Objects.nonNull(...)` also removes a TOCTOU window between the two calls.

## Verification

- Confirmed the Nacos registry already uses `ConcurrentHashMap` for its shared listener/instance maps (lines 64/66), and the only remaining `new HashSet` there is a local defensive copy inside `compareInstances(...)`, not a shared field — so the scope (Apollo/Consul/Etcd/Zookeeper) is complete and Nacos needs no change.
- None of these maps store `null` values (they always `put` a non-null `List<InstanceEntity>` / `Long`), so switching to `ConcurrentHashMap` (which forbids `null` keys/values) is safe.

## Non-blocking suggestions

- In Apollo (`childrenList`) and Etcd (`serverNodes`) the `ConcurrentHashMap` is a method-local variable captured by a (sequential) `forEach`/listener, so the concurrency win there is negligible — harmless but unnecessary. Not a problem.

## Conclusion

Approving. The fix addresses a real concurrency hazard with no behavioral downside.

PR #6990, opened 2026-08-27
fix: apply modify response rules to streaming responses

### LINKED ISSUE #6659: [BUG] ModifyResponsePlugin silently skips all modifications for streaming responses (writeAndFlushWith not overridden)
- severity: High
- files: `shenyu-plugin/shenyu-plugin-modify-response/src/main/java/org/apache/shenyu/plugin/modify/response/ModifyResponsePlugin.java:83-107`
- description: `ModifyResponseDecorator` extends `ServerHttpResponseDecorator` and overrides only `writeWith`. The parent's `writeAndFlushWith` delegates directly to `getDelegate().writeAndFlushWith(body)`. When the upstream response is streaming (SSE / `text/event-stream`, NDJSON), `NettyClientMessageWriter` calls `response.writeAndFlushWith(...)` which bypasses the decorator entirely. `modifyResponseHeadersAndStatus()` is never invoked — no header/status modification is applied.
- impact: All ModifyResponse header and status rules are silently ignored for SSE/NDJSON responses, with no error.
- suggested_fix: Override `writeAndFlushWith` in `ModifyResponseDecorator` to call `modifyResponseHeadersAndStatus()` before delegating.
- confidence: High
- related_existing: none — #5594 is content-type via `WebFluxResultUtils`; this is the writeAndFlushWith bypass affecting all header/status mods for streaming.

---

## Medium

---

liudawang001:
### What happened?

`ModifyResponseDecorator` only overrides `writeWith()`.

For streaming responses such as SSE and NDJSON, WebFlux invokes
`writeAndFlushWith()`, which previously bypassed
`modifyResponseHeadersAndStatus()`. As a result, configured response headers
and status codes were not applied.

### What does this PR do?

Override `writeAndFlushWith()` in `ModifyResponseDecorator` and invoke
`modifyResponseHeadersAndStatus()` before delegating the original streaming
body.

The body is neither aggregated nor modified, so the original streaming and
flush behavior is preserved.

### Tests

Added a regression test covering the `writeAndFlushWith()` path.

Verified that:

- Without the fix, the regression test fails because the configured response
  header is not applied.
- With the fix, the regression test passes.
- All 8 tests in `shenyu-plugin-modify-response` pass.
- Checkstyle reports no violations.
- `git diff --check` passes.

Fixes #6659

Make sure that:


Aias00 (approved):
## Summary

This PR correctly applies the configured modify-response rules (headers / status code) to **streaming** responses (SSE, NDJSON) by overriding `writeAndFlushWith()` in `ModifyResponseDecorator`.

## What is right

- WebFlux calls `writeAndFlushWith()` (not `writeWith()`) for streaming bodies, so previously `modifyResponseHeadersAndStatus()` only ran for non-streaming responses and the configured headers/status were silently dropped on SSE/NDJSON.
- The fix adds `writeAndFlushWith()` that calls `modifyResponseHeadersAndStatus()` and then delegates to `super.writeAndFlushWith(body)` **without aggregating or mutating the body** — so the original streaming / flush behavior is fully preserved.
- The regression test `testWriteAndFlushWith` builds the decorator, writes a `Flux.just(Flux.just(dataBuffer))` body, and asserts the added header (`X-Test`) and status (`CREATED`) are applied while the body content (`data: hello\n\n`) is passed through unchanged.

## Verification

- `ModifyResponseDecorator extends ServerHttpResponseDecorator`; `modifyResponseHeadersAndStatus()` is self-contained (reads the decorator's own headers, applies the rule, writes onto the delegate) and does not depend on which write method invoked it, so calling it from both `writeWith` and `writeAndFlushWith` is safe and not double-applied (WebFlux invokes exactly one per response).
- `@Override` on `writeAndFlushWith` is valid against `ServerHttpResponse`.

## Non-blocking suggestions

- Consider asserting in the test that the response is actually flushed as multiple signals (the `Flux.just(Flux.just(...))` shape already implies that), but the current coverage is sufficient.

## Conclusion

Approving. The change is minimal, correct, and well-tested.

Aias00:
pls add my wechat: aias00

PR #6992, opened 2026-08-27
fix: correct scale rule count query(#6870)

### LINKED ISSUE #6870: [BUG] namespace-user-rel batchSave foreach references wrong item variable (namespacePluginRelDO)
## Description
`countByQuery` is `select count(*) from scale_rule <if test="metricName != null"> and metric_name = #{metricName} </if>` with no `<where>` element and no leading `WHERE`. When `metricName` is non-null the rendered SQL is `select count(*) from scale_rule and metric_name = ?` → syntax error. (When null it runs as an unbounded full count.) Additionally `selectByQuery` uses `metric_name LIKE '%...%'` while `countByQuery` uses `metric_name =` — a count/list predicate mismatch. Currently no caller in `ScaleRuleServiceImpl`.

## Location
- `shenyu-admin/src/main/resources/mappers/scale-rule-sqlmap.xml:157-162`

## Impact
Latent broken SQL + count/list divergence; wiring scale-rule pagination would throw on any name filter.

## Suggested fix
Wrap in `<where>` and use `LIKE CONCAT('%', #{metricName}, '%')` to match `selectByQuery`.

## Related existing
Distinct from N22 (#6691, DataPermission selector/rule count-vs-list) — this is `scale_rule`, a different mapper and statement, and additionally has the missing-WHERE syntax error which N22 does not.

juicewcode:
Fixes #6870
Fix the invalid and inconsistent `scale_rule` count query 
  ## Changes

  - Wrap `countByQuery` conditions with MyBatis `<where>`.
  - Change metric filtering from exact match to `LIKE CONCAT('%', ..., '%')`.
  - Add mapper integration tests for filtered and unfiltered count queries.

Make sure that:


Aias00 (approved):
## Summary

Good fix. The old `countByQuery` produced invalid SQL whenever a `metricName` filter was supplied.

## What is correct

- Old SQL: `select count(*) from scale_rule <if metricName != null> and metric_name = ? </if>`. With a non-null `metricName` this rendered `... from scale_rule and metric_name = ?` — there is no `WHERE` keyword, so the statement is syntactically invalid and would fail at runtime. With a null `metricName` it happened to work, masking the bug.
- Wrapping the condition in `<where>` fixes this: MyBatis emits `WHERE` only when the block is non-empty and strips the leading `and`, so both the filtered and unfiltered paths are valid.
- The filter was also changed from exact `=` to `LIKE CONCAT('%', ?, '%')`. Verified `selectByQuery` already uses the same semantics (`<bind name="nameLike" value="('%' + metricName + '%')"/>` + a `LIKE` predicate), so count and select are now consistent — count of a filtered query equals the size of the filtered select, which the new tests assert.
- Integration tests cover both the filtered (`countByQueryUsesSameLikePredicateAsSelectByQuery`) and unfiltered (`countByQueryWithoutMetricNameCountsAllRules`) paths.

## Verdict

Approve. No blocking issues.

## Non-blocking note

Exact-match `=` → `LIKE %x%` is a behavior change for the count filter (broader matches). This is intentional to align with `selectByQuery`, but callers depending on exact-match counting should be aware.

PR #6993, opened 2026-08-27
fix: include manually assigned id in scale history inserts(#6871)

### LINKED ISSUE #6871: [BUG] scale_rule countByQuery omits WHERE keyword — SQL syntax error when metricName set + count/list mismatch
## Description
`insert`/`insertSelective` declare `<selectKey keyProperty="id" order="AFTER" resultType="java.lang.String"> SELECT LAST_INSERT_ID() </selectKey>`, but `scale_history.id` is `varchar(128) NOT NULL` (a manually-assigned PK, not auto-increment), and `id` is omitted from the insert column list. On MySQL strict mode the INSERT fails (no default for a NOT NULL varchar); even if it ran, `LAST_INSERT_ID()` returns 0 for a non-auto-increment insert → the DO's id is set to `"0"`. No `ScaleHistoryServiceImpl`/caller exists, so latent.

## Location
- `shenyu-admin/src/main/resources/mappers/scale-history-sqlmap.xml:57-59,68-70`

## Impact
Latent — any future caller of `scaleHistoryMapper.insert` will either fail the insert or receive a bogus id "0".

## Suggested fix
Remove the `<selectKey>` blocks and add `id` to the column list, binding `#{id, jdbcType=VARCHAR}` (let the service generate a UUID via `UUIDUtils`), consistent with other varchar-PK mappers.

## Related existing
None — `scale_history` mapper was not in the filed list.

juicewcode:
Fixes #6871
Fix scale history inserts for the manually assigned `varchar` primary key.

  ## Changes

  - Removed invalid `LAST_INSERT_ID()` selectKey blocks.
  - Added `id` binding to `insert`.
  - Added conditional `id` binding to `insertSelective`.
  - Added mapper integration tests covering both insert paths.

Make sure that:


Aias00 (approved):
## Summary

Good fix. The previous `scale_history` inserts dropped the manually-assigned primary key.

## What is correct

- `scale_history.id` is `varchar(128) NOT NULL PRIMARY KEY` (manually assigned, **not** auto-increment) — confirmed in `shenyu-admin/src/main/resources/sql-script/h2/schema.sql`.
- The old `insert` / `insertSelective` omitted the `id` column entirely and relied on a `SELECT LAST_INSERT_ID()` selectKey with `order="AFTER"`. That is meaningless for a non-auto-increment varchar PK, so the supplied `id` was never persisted (insert would either fail on the NOT NULL PK or store a wrong value).
- The fix adds `id` to both the column list and values of `insert`, and adds a conditional `id` binding (column + value guarded by `<if test="id != null">`) to `insertSelective`, then removes the bogus selectKey block. This is exactly the right shape for a manually-assigned PK.
- Two integration tests (`insertPreservesManuallyAssignedId`, `insertSelectiveIncludesIdAndUsesDatabaseDefaults`) assert the supplied id round-trips through both insert paths.

## Verdict

Approve. No blocking issues.

## Non-blocking note

`insertSelective` now requires the caller to set `id` (it is omitted when null, which would violate the NOT NULL PK). That matches the table contract; just flagging that any existing caller relying on `insertSelective` without an id would now fail — worth a quick grep of `ScaleHistoryMapper.insertSelective` call sites to confirm none pass a null id.

PR #6994, opened 2026-08-29
[type: bug] Fix createParameterSchema not writing required array (#6887)

### LINKED ISSUE #6887: [BUG] removeMcpServer deletes sibling servers' routes via prefix match
## Description
`createParameterSchema` and `recursionConstructPropertiesNodes` set only `type`/`description`/`properties`/`items`. The `required` array is never written, and `McpServerToolParameter.required`/`defaultValue` fields are ignored. Per JSON Schema / MCP, required parameters must be listed in a top-level `required` array.

## Location
- `shenyu-plugin/shenyu-plugin-mcp-server/src/main/java/org/apache/shenyu/plugin/mcp/server/utils/JsonSchemaUtil.java:49-67`

## Impact
MCP clients treat every parameter as optional and won't enforce/validate the tool's real contract; tools silently accept calls missing required arguments.

## Suggested fix
Collect parameters where `parameter.isRequired()` is true into a `required` ArrayNode on the schema root.

## Related existing
None.

zhang-arvin:
Closes #6887


Aias00 (approved):
## Summary

Good fix — `JsonSchemaUtil.createParameterSchema` was generating a JSON Schema with a `properties` map but no `required` array, so required parameters were not signaled to MCP clients.

## What is correct

- Adds the missing `required` array: iterates the parameters and adds `parameter.getName()` to `required` when `parameter.isRequired()` is true.
- The `required` entries use the exact same `parameter.getName()` that is used as the `properties` keys (`properties.putObject(parameter.getName())`), verified in `createParameterSchema`, so every name listed in `required` corresponds to an existing property — no dangling references.
- `schema` is a freshly created `ObjectNode`, so `schema.putArray("required")` does not clash with a pre-existing `required` key.
- Verified `McpServerToolParameter` has `isRequired()` and `getName()`.

## Verdict

Approve. No blocking issues.

## Non-blocking note

The `required` array only covers top-level parameters. If a nested object parameter (`recursionConstructPropertiesNodes` handles nested `properties`/`items`) is itself required, it will not appear in the top-level `required` array. Out of scope for this PR, but worth a follow-up if nested required semantics matter for the tool schema.

PR #6995, opened 2026-08-29
[type: bug] Fix thread safety issue with shared static CONTEXT in ExpressionGenerator (#6880)

### LINKED ISSUE #6880: [BUG] SignPluginDataHandler and AbstractCryptorPluginDataHandler guard removeRule on ruleData.getHandle() — leaking stale rule-handle cache entries
## Description
Both generators hold a `private static final EvaluationContext CONTEXT` (a single shared instance across all requests). `doGenerate` calls `CONTEXT.setVariable("req", mockRequest)` then `PARSER.parseExpression(expression).getValue(CONTEXT)`. Under concurrent load (multiple Netty event-loop threads), thread A's `setVariable("req", reqA)` can be overwritten by thread B's `setVariable("req", reqB)` before thread A's `getValue(CONTEXT)` executes. The SpEL expression then evaluates against the wrong request's `MockRequest` (body, headers, query params, URI).

## Location
- `shenyu-plugin/shenyu-plugin-mock/src/main/java/org/apache/shenyu/plugin/mock/generator/ExpressionGenerator.java:44,56-58`
- `shenyu-plugin/shenyu-plugin-mock/src/main/java/org/apache/shenyu/plugin/mock/generator/StandardExpressionGenerator.java:43,55-57`

## Impact
Cross-request data leak: a mock response for request A may contain request B's body/headers/query parameters. Non-deterministic, intermittent wrong mock responses under concurrency.

## Suggested fix
Create a new `EvaluationContext` per request (or per `doGenerate` call) instead of sharing a static one.

## Related existing
None — distinct from #6544 (DividePlugin `beginTime` shared mutable) which is a different shared-state race in a different plugin.

zhang-arvin:
Closes #6880


Aias00 (approved):
## Summary

Correct concurrency fix for the mock expression generators.

## Root cause (confirmed)

`ExpressionGenerator` (and `StandardExpressionGenerator`) held a `private static final EvaluationContext CONTEXT` that was mutated per request via `CONTEXT.setVariable("req", mockRequest)` before `getValue(CONTEXT)`. Because the context was shared and mutable, concurrent mock evaluations raced on the `req` variable (and any other context state), producing cross-request interference or inconsistent results.

## What the fix does right

- Replacing the shared static with a fresh `initContext()` per `doGenerate` call eliminates the shared mutable state — each invocation gets its own `EvaluationContext`.
- The static `SpelExpressionParser PARSER` remains shared, which is safe: `SpelExpressionParser` and the parsed `Expression` objects are documented thread-safe, so the parser was never the problem — only the context.
- Verified `initContext()` returns a new `EvaluationContext` each call, and no other references to the removed static `CONTEXT` remain in either `ExpressionGenerator` or `StandardExpressionGenerator`.

## Verdict

Approve. No blocking issues.

PR #6996, opened 2026-08-29
[type: bug] Fix TimeUnit.MICROSECONDS to MILLISECONDS in log collect clients (#6885)

### LINKED ISSUE #6885: [BUG] ThreadPool keep-alive is 60 ms instead of 60 s in TencentCLS and AliyunSls log clients
## Description
`initClient0` registers `Runtime.getRuntime().addShutdownHook(new Thread(this::close))` in addition to the base-class hook tracked/removed by `AbstractLogConsumeClient`. On every reconfig, only the base-class hook is removed; the RocketMQ-added hook is never removed and accumulates. At JVM shutdown `close()` is invoked once per leaked hook.

## Location
- `shenyu-plugin/shenyu-plugin-logging/shenyu-plugin-logging-rocketmq/src/main/java/org/apache/shenyu/plugin/logging/rocketmq/client/RocketMQLogCollectClient.java:83`

## Impact
Shutdown-hook leak across config refreshes; `close0` (`producer.shutdown()`) called multiple times at shutdown; hook threads accumulate.

## Suggested fix
Remove the extra `addShutdownHook`; rely on the base-class tracked hook, or track/remove this hook in `close0`.

## Related existing
Distinct from #6768 (Pulsar shutdown-hook, LOG-11) — same pattern, different file.

zhang-arvin:
Closes #6885


Aias00 (approved):
## Summary

Clear, correct fix.

## What is correct

- The `ThreadPoolExecutor` keep-alive time was `60000L` with `TimeUnit.MICROSECONDS`, i.e. **60 milliseconds** — worker threads would be reaped almost immediately, defeating the pool. Changing it to `TimeUnit.MILLISECONDS` makes it **60 seconds**, the clearly intended value.
- Applied consistently across all three logging clients: Aliyun SLS, Huawei LTS, and Tencent CLS.
- Verified all three head files no longer contain `MICROSECONDS`.

## Verdict

Approve.

## Non-blocking suggestion

While here, it may be worth a quick sweep of the other log collect clients (e.g. elasticsearch / kafka / rocketmq / clickhouse) for the same `MICROSECONDS` typo, in case the bug is more widespread. Out of scope for `#6885`, just a suggestion.

PR #6997, opened 2026-08-29
[type: bug] Fix session key mismatch in streamable HTTP transport (#6833)

### LINKED ISSUE #6833: [BUG] MCP Streamable-HTTP session cleanup uses wrong key (auto-UUID) — sessions leak forever
## Description
`ShenyuStreamableHttpServerTransportProvider.handleInitializeRequest` builds `new StreamableHttpSessionTransport()` (constructor sets `this.sessionId = UUID.randomUUID()` — the auto-UUID), but stores the session under `newSessionId = session.getId()` (the `McpServerSession`'s id, different from the transport's auto-UUID). `transport.close()`/`closeGracefully()` call `removeSession(this.sessionId)` (the auto-UUID), which removes nothing; the entries under `newSessionId` stay forever. There is also no TTL/eviction on `sessions`/`sessionTransports`/`ShenyuMcpExchangeHolder.EXCHANGE_MAP`.

## Location
- `shenyu-plugin/shenyu-plugin-mcp-server/src/main/java/org/apache/shenyu/plugin/mcp/server/transport/ShenyuStreamableHttpServerTransportProvider.java:321-329` (configureExchangeForSession + store under newSessionId), `1020-1023` (transport sessionId = auto-UUID), `1071-1088` (close/closeGracefully call removeSession(this.sessionId)), `763-766` (removeSession)

## Impact
Unbounded memory growth / DoS — every Streamable-HTTP session leaks until process shutdown. (The temporary-session path at :433 correctly uses `session.getId()` for cleanup, which highlights that the permanent-session path is the bug.)

## Suggested fix
Store sessions under the same key the transport uses for cleanup (or make `removeSession` look up by transport's sessionId); add a TTL/last-seen eviction sweep.

## Related existing
Distinct from MCP-05 (#6644, `createSessionAndRestoreId` restore-path creation leak) — this is the main initialize-path cleanup-key mismatch plus total absence of expiry.

zhang-arvin:
Fixes #6833

### Problem
`ShenyuStreamableHttpServerTransportProvider.handleInitializeRequest` creates a `StreamableHttpSessionTransport` with auto-generated UUID as sessionId, but stores sessions under `McpServerSession.getId()` (a different key). When `close()`/`closeGracefully()` calls `removeSession(this.sessionId)`, it uses the auto-UUID, so nothing gets removed — sessions leak forever.

### Fix
After creating the session, sync the transport's sessionId to match the McpServerSession's ID via a new `setSessionId()` method, ensuring `close()`/`closeGracefully()` correctly removes the session.


Aias00 (approved):
## Summary

Correct fix for a session-leak bug in the streamable HTTP MCP transport.

## Root cause (confirmed)

`handleInitializeRequest` created a `StreamableHttpSessionTransport` whose no-arg constructor auto-generates a UUID `sessionId`, then stored the session/transport under `session.getId()` (a different key). `close()`/`closeGracefully()` call `removeSession(this.sessionId)` with the transport's auto-UUID, which never matched the store key — so sessions were never removed and leaked forever.

## What the fix does right

- `transport.setSessionId(newSessionId)` (immediately after `session.getId()`) aligns the transport's `sessionId` field with the keys used in `sessions.put`, `sessionTransports.put`, and `ShenyuMcpExchangeHolder`, so `removeSession` now deletes the correct entry.
- Verified in the head file: the inner class's no-arg constructor auto-generates `sessionId`; both `close()` (line 1096) and `closeGracefully()` (line 1085) use `this.sessionId` in `removeSession`; `setSessionId` is invoked right after the session is created and before any close can occur, so there is no window where the field is stale.

## Verdict

Approve. No blocking issues.

## Non-blocking suggestion

Consider adding a regression test that performs initialize → close and asserts `sessions`/`sessionTransports` become empty, to lock in the fix and prevent reintroduction.

zhang-arvin:

Aias00 (changes requested):

zhang-arvin:
@Copilot Thanks — the compile error was real and is now fixed.

You were right that `sessionId` is declared `final` (line 1047) and is already assigned by both constructors, so `setSessionId()` could not assign to it. Confirmed with a local build on the previous head:

```
[ERROR] ShenyuStreamableHttpServerTransportProvider.java:[1073,17] 无法为最终变量sessionId分配值
```

(the compiler's `cannot assign a value to final variable sessionId`).

**Fix:** dropped the `final` modifier so `setSessionId()` can synchronize the transport's ID with the key the provider registers the session under. That synchronization is the whole point of this PR — without it `close()` looked up an ID that never existed in `sessions` / `sessionTransports` and left both registries (plus `ShenyuMcpExchangeHolder`) populated forever.

**Regression test added**, addressing your second comment: after a real initialize handshake, the session ID returned to the client via `Mcp-Session-Id` must be exactly the key used in both `sessions` and `sessionTransports` — i.e. the ID `close()` will later look up. Verified:

`./mvnw -pl shenyu-plugin/shenyu-plugin-mcp-server test -Dtest=ShenyuStreamableHttpServerTransportProviderTest` → **Tests run: 6, Failures: 0, Errors: 0**.

I also re-verified the fix is load-bearing by restoring `final` locally — the module then fails to compile at line 1073, so the change and its test are not vacuous.

@Aias00 the branch is synced with `master`. Would you mind taking another look?

Aias00 (approved):
Good fix for a genuine leak, with a clear root-cause explanation.

The mismatch was subtle: `handleInitializeRequest` created `StreamableHttpSessionTransport()` — whose no-arg constructor **auto-generates its own** `sessionId` — then created the MCP session via `sessionFactory.create(transport)` and registered everything under `session.getId()`. So the transport held id A while `sessions` / `sessionTransports` were keyed by id B. When `close()` / `closeGracefully()` later tried to remove the entry using the transport's id, the lookup missed and both registries (plus `ShenyuMcpExchangeHolder`) kept the entry forever — precisely the leak reported in #6833.

Calling `transport.setSessionId(newSessionId)` right after the session is created aligns the two identifiers, and the added javadoc explains *why* the setter exists, which is valuable because a setter on an otherwise final field looks odd without context.

Notes:
- Changing `sessionId` from `final` to mutable is the minimal way to do this given the two-phase construction (transport first, session id second). Alternatively the id could be passed into the constructor, but that would require restructuring `sessionFactory.create(transport)`.
- The regression test is the right kind: it performs a real initialize handshake, reads the `Mcp-Session-Id` the client actually receives, and asserts **that** id is present in both registries. Testing against the returned id (rather than an internal value) is what makes the test meaningful — a reimplementation that still keyed the registries differently would fail it.

Non-blocking: the id is now assigned after construction, so between `new StreamableHttpSessionTransport()` and `setSessionId(...)` the transport briefly reports a stale id. Nothing reads it in that window, but a constructor parameter would remove the window entirely if you ever touch this class again.

PR #6998, opened 2026-08-29
[type: bug] Fix REFRESH clearing only first plugin's parent path (#6832)

### LINKED ISSUE #6832: [BUG] AbstractPathDataChangedListener REFRESH clears only the first plugin's parent path — stale selectors/rules for other plugins persist (zk/etcd/consul)
## Description
On `REFRESH`, `AbstractPathDataChangedListener.onSelectorChanged` takes `changed.stream().findFirst()` and uses `firstData.getPluginName()` to build `buildSelectorParentPath`, then `deletePathRecursive` on that **single** parent path. The same pattern in `onRuleChanged`. But `syncAll`/`syncAllByNamespaceId` publish ALL selectors (or rules) of an entire namespace — spanning many plugins — in one REFRESH event. Only the first encountered plugin's selector/rule parent subtree is recursively deleted; for every other plugin the stale subtree is not cleared, and the subsequent per-item `createOrUpdate` only adds/updates the new entries without removing deleted siblings.

## Location
- `shenyu-admin-listener/shenyu-admin-listener-api/src/main/java/org/apache/shenyu/admin/listener/AbstractPathDataChangedListener.java:113-121` (selector REFRESH), `158-166` (rule REFRESH)
- trigger: `shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/SyncDataServiceImpl.java:124-129` (`syncAll`), `152-156` (`syncAllByNamespaceId`); callers `NamespacePluginController:210`, `ConfigsExportImportController:141`

## Impact
On zookeeper/etcd/consul sync backends, a namespace resync or config import (reachable via admin controller endpoints) leaves deleted selectors and rules for all but one plugin present in the config tree. Gateways syncing from these backends continue to load and apply stale selectors/rules — persistent routing drift / phantom rules after what should be a full refresh.

## Suggested fix
In the REFRESH block, group `changed` by pluginName (and namespaceId) and `deletePathRecursive(buildSelectorParentPath(ns, pluginName))` for each distinct plugin, instead of only `findFirst()`. Apply the same fix to the rule REFRESH block.

## Related existing
Distinct from SYNC-3/SYNC-4 (gateway-side `shenyu-sync-data-http` `SelectorDataRefresh`/`RuleDataRefresh` NPE on null/empty) and from H-8 (nacos/apollo/polaris node-based). This is the path-based listener (zk/etcd/consul) with a first-plugin-only clear root cause.

zhang-arvin:
Fixes #6832

### Problem
On REFRESH, `AbstractPathDataChangedListener.onSelectorChanged` and `onRuleChanged` use `changed.stream().findFirst()` to get the first element's plugin name, then call `deletePathRecursive` on that single parent path. But `syncAll`/syncAllByNamespaceId` publish ALL selectors/rules of an entire namespace — spanning many plugins — in one REFRESH event. Only the first plugin's subtree is cleared; stale selectors/rules for other plugins persist.

### Fix
Group `changed` by `namespaceId + pluginName` and `deletePathRecursive` for each distinct parent path, instead of only `findFirst()`. Applied to both `onSelectorChanged` and `onRuleChanged` REFRESH blocks.


Aias00 (approved):
## Summary

Fixes `onSelectorChanged` / `onRuleChanged` in `AbstractPathDataChangedListener` so that a `REFRESH` event clears **all** affected parent paths instead of only the first one.

## Problem

The old code used `changed.stream().findFirst()` and deleted only that single selector/rule parent path (`buildSelectorParentPath(firstData.getNamespaceId(), firstData.getPluginName())`). When a REFRESH payload spans multiple plugins (or multiple namespaces), the stale data under the other parent paths was never cleared — leaving orphaned config nodes.

## Fix

Group the changed data by parent path key and delete every distinct one:

```java
changed.stream()
        .collect(Collectors.groupingBy(
                data -> DefaultPathConstants.buildSelectorParentPath(data.getNamespaceId(), data.getPluginName()),
                LinkedHashMap::new,
                Collectors.toList()))
        .keySet()
        .forEach(this::deletePathRecursive);
```

`LinkedHashMap` preserves insertion order; `DefaultPathConstants.buildSelectorParentPath` / `buildRuleParentPath` are already used in the per-item loop below, so no new dependencies are introduced.

## Verification

- The `Optional` import was removed and **no remaining `Optional` usages** exist in the file (verified on head branch).
- New imports `java.util.LinkedHashMap` and `java.util.stream.Collectors` are present.
- The per-item publish loop (`buildSelectorRealPath` / `buildRulePath`) is unchanged, so the fix only affects the stale-path cleanup scope.

## Conclusion

Correct, minimal fix for multi-plugin / multi-namespace REFRESH stale-data cleanup. **Approve.**

PR #6999, opened 2026-08-29
[type: bug] Add namespace_id to alert_receiver Base_Column_List (#6829)

### LINKED ISSUE #6829: [BUG] alert_receiver Base_Column_List omits namespace_id — cross-namespace alert dispatch
## Description
`alert-receiver-sqlmap.xml` `<sql id="Base_Column_List">` lists every column **except** `namespace_id`. `selectAll` uses this list, so every returned `AlertReceiverDTO.namespaceId` is null. `AlertDispatchServiceImpl.matchReceiverByRules` then runs `if (StringUtils.isNotBlank(item.getNamespaceId())) { ... namespaceIdMatch ... }` — the guard is always false, so the namespace-scoping branch never executes. `insert` writes `namespace_id` correctly, so the data is in the DB; only the read path is broken.

## Location
- `shenyu-admin/src/main/resources/mappers/alert-receiver-sqlmap.xml:53-62`
- `shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/AlertDispatchServiceImpl.java:151,160-164`

## Impact
A receiver with `matchAll=false` (level/label matching only) receives alerts from ALL namespaces — a namespace-isolation bypass in the alerting subsystem. Receivers intended to be namespace-scoped silently become global.

## Suggested fix
Add `namespace_id` to `Base_Column_List` in `alert-receiver-sqlmap.xml` (after `levels`, before `date_created`).

## Related existing
Distinct from PERF-A11 (#6810, `selectAll` performance) and N56 (cache staleness across instances). This is a correctness defect: the column is omitted from the projection so namespaceId is always null and the namespace guard is dead.

zhang-arvin:
Fixes #6829

### Problem
`alert-receiver-sqlmap.xml` `Base_Column_List` lists every column **except** `namespace_id`. `selectAll` uses this list, so every returned `AlertReceiverDTO.namespaceId` is null. `AlertDispatchServiceImpl.matchReceiverByRules` then runs `if (StringUtils.isNotBlank(item.getNamespaceId())) { ... namespaceIdMatch ... }` — the guard is always false, so namespace-scoping never executes. Receivers intended to be namespace-scoped silently become global.

### Fix
Add `namespace_id` to `Base_Column_List` (after `levels`, before `date_created`).


Aias00 (approved):
## Summary

Adds `namespace_id` to the `alert_receiver` `Base_Column_List` so that `SELECT` statements populate `AlertReceiverDTO.namespaceId`.

## Verification

- `AlertReceiverDTO` has a `namespaceId` field with getter/setter (verified on `master`), so adding the column to `Base_Column_List` maps correctly into the DTO.
- This is a **pure column addition** — unlike #7015, it does not remove any existing column, so there is no regression risk for other fields.
- `namespace_id` is a real column on the `alert_receiver` table (consistent with the rest of the namespace-isolation series).

## Conclusion

Safe, additive mapper change. **Approve.**

PR #7000, opened 2026-08-29
fix: add namespace_id filter to app_auth selectByCondition (#6828)

### LINKED ISSUE #6828: [BUG] app_auth selectByCondition omits namespace_id — cross-namespace app-auth read
## Description
`app_auth selectByCondition` filters only on `appKey` and `phone` in a `<where>` block; there is no `namespace_id` predicate, unlike `selectByQuery` (which filters `namespace_id = #{namespaceId}`). `AppAuthQuery` carries `namespaceId`. `AppAuthServiceImpl.searchByCondition(condition)` calls this mapper directly.

## Location
- `shenyu-admin/src/main/resources/mappers/app-auth-sqlmap.xml:48-60`
- `shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/AppAuthServiceImpl.java:98-99`

## Impact
`searchByCondition` returns app-auth rows across all namespaces, leaking app-key/phone entries belonging to other namespaces to any caller of this service method.

## Suggested fix
Add `<if test="condition.namespaceId != null and condition.namespaceId != ''"> AND namespace_id = #{condition.namespaceId, jdbcType=VARCHAR} </if>` to `selectByCondition`.

## Related existing
Distinct from N21 (#6700, `app_auth.updateSelective` WHERE omits namespace_id — write path). This is the read path `selectByCondition`.

zhang-arvin:
Fixes #6828


Aias00 (approved):
## Summary

Adds a `namespace_id` filter to the `app_auth selectByCondition` query, continuing the namespace-isolation hardening series (see #7001, #7007, #7003, #6999). When `condition.namespaceId` is provided, the query restricts results to that namespace.

## Verification

- `AppAuthQuery` has a `namespaceId` field with getter/setter (verified on `master`), so `condition.namespaceId` in the mapper resolves correctly.
- This is a pure additive `<if>` condition in `app-auth-sqlmap.xml`; it does not change any method signature and has no callers to update.
- Consistent with the existing pattern used for `selector`/`plugin`/`discovery` queries.

## Conclusion

Minimal, well-scoped SQL change consistent with the surrounding namespace-isolation work. **Approve.**

PR #7001, opened 2026-08-29
fix: add namespace_id filter to selector selectByIdSet and deleteByIds (#6826)

### LINKED ISSUE #6826: [BUG] selector selectByIdSet/deleteByIds omit namespace_id — cross-namespace selector delete (+ cascade)
## Description
`SelectorServiceImpl.deleteByNamespaceId(List<String> ids, String namespaceId)` calls `selectorMapper.selectByIdSet(new TreeSet<>(ids))` whose SQL is `WHERE id IN (...)` with no `namespace_id`, and the eventual `deleteByIds` is `DELETE FROM selector WHERE id IN (...)`. `namespaceId` is passed only to event publishing, never to SELECT or DELETE. Sibling `selectByQuery`/`updateSelective` enforce `namespace_id`, confirming bug.

Controller-exposed at `SelectorController:155`. The fetched selectors (unfiltered by namespace) also drive `unbindDiscovery` and the downstream `BatchSelectorDeletedEvent` rule cleanup, amplifying blast radius.

## Location
- `shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/SelectorServiceImpl.java:297-303,633-636`
- `shenyu-admin/src/main/resources/mappers/selector-sqlmap.xml:80-88` (`selectByIdSet`), `465-471` (`deleteByIds`)
- `shenyu-admin/src/main/java/org/apache/shenyu/admin/controller/SelectorController.java:152-156`

## Impact
Cross-namespace selector deletion (and cascade unbindDiscovery/deleteSelector) — a user can tear down selectors in another namespace, removing routing configuration and bound discovery bindings. Requires authenticated admin with `system:plugin:delete`.

## Suggested fix
Add `AND namespace_id = #{namespaceId}` to `selectByIdSet` and `deleteByIds`, or filter ids via a namespace-scoped select first.

## Related existing
Distinct from H-2 (rule) and N2 (#6615) — different mapper file and operation.

zhang-arvin:
Fixes #6826


Aias00 (approved):
## Review: Add namespace_id filter to selector selectByIdSet and deleteByIds (#6828)

**Verdict: APPROVE**

### Analysis

Scopes selector read/delete by namespace, preventing cross-namespace operations (a multi-tenant isolation fix).

**Contract changes**
- `SelectorMapper.selectByIdSet(Set<String> idSet)` → adds `@Param("namespaceId") String namespaceId`.
- `SelectorMapper.deleteByIds(List<String> ids)` → adds `@Param("namespaceId") String namespaceId`.
- `selector-sqlmap.xml`:
  ```sql
  SELECT <Base_Column_List> FROM selector
  WHERE namespace_id = #{namespaceId, jdbcType=VARCHAR}
    AND id IN ( ... )
  ```
  applied to both `selectByIdSet` and `deleteByIds`.

**Callers updated (`SelectorServiceImpl`)**
- `deleteByNamespaceId(ids, namespaceId)` now passes `namespaceId` into `selectByIdSet(...)` and into the new `deleteSelector(..., namespaceId)`, which forwards it to `deleteByIds(...)`.
- `onPluginDeleted` (`@EventListener` for `BatchNamespacePluginDeletedEvent`) forwards `event.getNamespaceId()` through `deleteSelector`.
- `deleteSelector` signature updated to carry `namespaceId`.

`selector` has a `namespace_id` column (14 references in the mapper), so the new predicate is valid. The `selectByIdSet` result is already namespace-filtered, and `deleteByIds` re-filters by the same `namespaceId` — redundant but safe (idempotent). No other callers of `selectByIdSet`/`deleteByIds` exist in the repo, so the signature change doesn't break compilation elsewhere.

### Conclusion
Consistent, well-scoped namespace isolation for selector deletion, with all in-repo call sites updated. Approving.

zhang-arvin:

dengliming (changes requested):
This PR still misses the actual cause of the CI failure. The failing `it` job is blocked by Checkstyle because `SelectorMapper` has methods with a `namespaceId` parameter but the corresponding Javadocs do not declare `@param namespaceId`.

Please update the Javadocs for:
- `selectByIdSet(..., String namespaceId)`
- `deleteByIds(..., String namespaceId)`

Suggested fix:

```java
/**
 * Select selector by a list of ids.
 *
 * @param idSet a list of ids
 * @param namespaceId namespaceId
 * @return a list of {@linkplain SelectorDO}
 */
List<SelectorDO> selectByIdSet(@Param("idSet") Set<String> idSet, @Param("namespaceId") String namespaceId);

/**
 * delete selector.
 *
 * @param ids primary keys.
 * @param namespaceId namespaceId
 * @return rows int
 */
int deleteByIds(@Param("list") List<String> ids, @Param("namespaceId") String namespaceId);

**Comment by Aias00** [comment]
@zhang-arvin hi, pls add my wechat: aias00, so that i can add you to or contributor group

**Review by Aias00 — CHANGES_REQUESTED** [review]

**Comment by zhang-arvin** [comment]
@dengliming @Aias00 Thanks for the review. The branch was 37 commits behind `master` (which is what produced the `behind` mergeable state and likely the stale CI signal), so I've rebased it onto the current `master` and force-pushed — no code changes, the diff is unchanged.

Also: sorry for the delay in picking up your WeChat invitation — I'm in China and will add you at `aias00` shortly.

Would you mind re-running the checks / taking another look?

**Comment by Aias00** [comment]
> @dengliming @Aias00 Thanks for the review. The branch was 37 commits behind `master` (which is what produced the `behind` mergeable state and likely the stale CI signal), so I've rebased it onto the current `master` and force-pushed — no code changes, the diff is unchanged.
> 
> Also: sorry for the delay in picking up your WeChat invitation — I'm in China and will add you at `aias00` shortly.
> 
> Would you mind re-running the checks / taking another look?

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.4.0:check (validate) on project shenyu-admin: You have 2 Checkstyle violations. -> [Help 1]

plx fix checkstyle

**Review by Aias00 — CHANGES_REQUESTED** [review]
Keeping this at REQUEST_CHANGES — there is a concrete compile break, not just a CI hiccup.

**Blocking: the mapper signature change was not propagated to the tests.**

`SelectorMapper#selectByIdSet` and `#deleteByIds` now take an extra `namespaceId` parameter, but on this branch three call sites still use the old single-argument form:

```
shenyu-admin/src/test/java/.../mapper/SelectorMapperTest.java:75
    List<SelectorDO> selectorList = selectorMapper.selectByIdSet(idSet);

shenyu-admin/src/test/java/.../service/SelectorServiceTest.java:182
    given(selectorMapper.selectByIdSet(Stream.of(correctId).collect(Collectors.toSet()))).willReturn(...);

shenyu-admin/src/test/java/.../service/SelectorServiceTest.java:198
    given(selectorMapper.deleteByIds(ids)).willReturn(ids.size());
```

Since the interface no longer declares the one-arg overloads, `shenyu-admin` test compilation fails. Please update these to pass a namespace id (and, ideally, add an assertion that a selector belonging to a *different* namespace is not returned/deleted — that is the behaviour this PR is introducing and it is currently untested).

**On the production change itself** — the intent is right and I'm happy with it:

- Scoping `selectByIdSet` / `deleteByIds` with `WHERE namespace_id = #{namespaceId}` closes a real namespace-isolation hole: a batch delete could previously remove rows in another namespace if ids were supplied from outside, and `selectByIdSet` could return cross-namespace selectors that were then unbound from their discovery handlers.
- `deleteByNamespaceId` correctly threads `namespaceId` through to `deleteSelector`, and `onPluginDeleted` passes `event.getNamespaceId()`.
- Using `@Param("list")` on `deleteByIds` is required for the two-arg form to bind correctly with the existing `<foreach collection="list">` — good catch.

Once the three test call sites are fixed and CI is green, this is ready to approve.

**Comment by Aias00** [comment]
Follow-up to my review: I have now scanned **all** Java sources on this branch head, and the set of call sites not updated for the new two-argument signature is **larger than the three I listed** — it also includes two production classes, not just tests.

`SelectorMapper#selectByIdSet` / `#deleteByIds` now require `namespaceId`, but these still pass a single argument:

**Production code**
```
shenyu-admin/.../service/impl/RuleServiceImpl.java:478
    Map<String, String> pluginIdMap = Optional.ofNullable(selectorMapper.selectByIdSet(new HashSet<>(ruleDOMap.values()...)))...

shenyu-admin/.../service/impl/RuleServiceImpl.java:525
    (same pattern)

shenyu-admin/.../AiProxyRealKeyResolver.java:226
    java.util.List<SelectorDO> selectors = selectorMapper.selectByIdSet(missing);
```

**Tests**
```
shenyu-admin/src/test/.../mapper/SelectorMapperTest.java:75
    selectorMapper.selectByIdSet(idSet)

shenyu-admin/src/test/.../service/SelectorServiceTest.java:182
    given(selectorMapper.selectByIdSet(Stream.of(correctId).collect(Collectors.toSet()))).willReturn(...)

shenyu-admin/src/test/.../service/SelectorServiceTest.java:198
    given(selectorMapper.deleteByIds(ids)).willReturn(ids.size());

shenyu-admin/src/test/.../service/RuleServiceTest.java:392
    given(this.selectorMapper.selectByIdSet(Sets.newHashSet("456"))).willReturn(...)

shenyu-admin/src/test/.../AiProxyRealKeyResolverTest.java  (lines 130, 137, 153, 161, 170, 177, 183, 193, 209, 225, 242)
    when(selectorMapper.selectByIdSet(selectorIds)).thenReturn(...)
    verify(selectorMapper, times(1)).selectByIdSet(selectorIds);
```

(For clarity: the many other `deleteByIds` hits in that scan — `PluginMapper`, `RuleMapper`, `TagMapper`, etc. — are **different** mappers with their own unrelated `deleteByIds(List)` methods and are unaffected. Only `selectorMapper.*` calls need updating.)

Why this matters beyond compilation: `RuleServiceImpl` and `AiProxyRealKeyResolver` are real runtime paths. `AiProxyRealKeyResolver` resolves API keys to selectors and currently has **no** namespace context at that call site, so it needs a deliberate decision — is it operating in the default namespace (use `Constants.SYS_DEFAULT_NAMESPACE_ID`), or does it need the namespace threaded in from its caller? Same question for both `RuleServiceImpl` sites. Please don't just pass `null` there: with the new SQL the predicate becomes `namespace_id = NULL`, which matches **no rows** in SQL (three-valued logic), so those code paths would silently start returning empty results instead of failing loudly.

Once these are updated consistently — and ideally with a test asserting a selector in another namespace is neither returned nor deleted — this is ready to approve. The change itself is the right fix.

PR #7008, opened 2026-08-30
[type: bug] Fix NPE in Hystrix/Resilience4J/Sentinel/RateLimiter plugins (#6877)

### LINKED ISSUE #6877: [BUG] Hystrix / Resilience4J / Sentinel / RateLimiter plugins NPE when a matched rule has no cached handle
## Description
`handlerPlugin` checks if the new `RedisConfigProperties` differs from the cached singleton. If it differs, it constructs a new `RedisConnectionFactory` (which opens Lettuce connection pools via `afterPropertiesSet()`), stores the new `ReactiveRedisTemplate` and `RedisConfigProperties` into `Singleton.INST`, but never calls `destroy()` on the OLD `LettuceConnectionFactory`. The old factory's connection pool (sockets, NIO event loops, threads) is orphaned. Trigger: any admin update that changes the redis URL, password, database, or pool settings.

## Location
- `shenyu-plugin/shenyu-plugin-fault-tolerance/shenyu-plugin-ratelimiter/src/main/java/org/apache/shenyu/plugin/ratelimiter/handler/RateLimiterPluginDataHandler.java:50-65`
- `shenyu-infra/shenyu-infra-redis/src/main/java/org/apache/shenyu/infra/redis/RedisConnectionFactory.java:43-48`

## Impact
Each redis config change permanently leaks a Lettuce connection pool. Repeated config updates accumulate leaked connections, eventually exhausting file descriptors / sockets and degrading gateway stability.

## Suggested fix
Before installing the new factory, retrieve the old `ReactiveRedisTemplate` from `Singleton.INST`, extract its `LettuceConnectionFactory` (or track the old `RedisConnectionFactory`), and call `destroy()` on it before `Singleton.INST.single(...)`.

## Related existing
None — distinct from #6645 (`RedisRateLimiter` fail-open) because this is a resource-leak in the data handler's config-update path, not the rate-limiting execution path.

zhang-arvin:
Closes #6877


Aias00 (changes requested):
## Review: Fix NPE in Hystrix/Resilience4J/Sentinel/RateLimiter plugins (#6877)

**Verdict: REQUEST CHANGES**

### What's correct
The null-guard logic for all four plugins is sound: when the cached handle is `null`, fall through to `chain.execute(exchange)` instead of dereferencing the handle and throwing an NPE. The Hystrix variant additionally logs the offending rule, which is helpful.

### Compile-blocking defect: `RateLimiterPlugin` is missing the `Objects` import
`RateLimiterPlugin.doExecute` now uses `Objects.isNull(limiterHandle)`:
```java
RateLimiterHandle limiterHandle = RateLimiterPluginDataHandler.CACHED_HANDLE.get()
        .obtainHandle(CacheKeyUtils.INST.getKey(rule));
if (Objects.isNull(limiterHandle)) {     // <-- compile error
}
```
but the file's imports are only:
```java
import java.util.Optional;
import java.util.function.Consumer;
```
There is **no `import java.util.Objects;`**. This will fail to compile (unresolved symbol `Objects`). The other three plugins are fine because they already import `java.util.Objects` (`Objects.requireNonNull(shenyuContext)` was already present in `HystrixPlugin`, `Resilience4JPlugin`, and `SentinelPlugin`), and `HystrixPlugin` already declares `private static final Logger LOG`. So only `RateLimiterPlugin` needs the fix.

### Required change
Add the import at the top of `RateLimiterPlugin.java`:
```java
import java.util.Objects;
```
After that one-line addition the PR compiles and all four guards are correct.

### Conclusion
The fix is the right idea and 3 of the 4 files are ready; please add the missing `import java.util.Objects;` to `RateLimiterPlugin` (and run a full `mvn compile` to confirm). Requesting changes until the import is added.

zhang-arvin:
- Added missing `import java.util.Objects;` in RateLimiterPlugin.java

Please take another look.

Aias00 (approved):
## Re-review: [type: bug] Fix NPE in Hystrix/Resilience4J/Sentinel/RateLimiter plugins (#6877)

Re-reviewed at head `5169568`. My previous REQUEST_CHANGES flagged that `RateLimiterPlugin` calls `Objects.isNull(...)` but was missing `import java.util.Objects;`, so the module would not compile. That is now fixed:

- `RateLimiterPlugin.java` line 24 adds `import java.util.Objects;` (verified: exactly one occurrence, no duplicate import). The `if (Objects.isNull(limiterHandle)) { return chain.execute(exchange); }` guard now compiles and short-circuits the plugin safely when no handle is cached.
- `HystrixPlugin` (lines 9-12), `Resilience4JPlugin` (lines 46-48), `SentinelPlugin` (lines 60-62) each gain the same null-handle guard; `HystrixPlugin` additionally logs via `LOG`, which is confirmed present (slf4j `Logger`/`LoggerFactory` imports + `LOG` field at lines 37-38/53).

Behavior is correct: a missing cached handle now falls through to `chain.execute` instead of throwing NPE on `hystrixHandle.getGroupKey()` / `resilience4JHandle.checkData()` / `sentinelHandle.checkData()`.

The same fault-tolerance NPE fixes are also carried by #7022 (which subsumes this PR). Functionally this PR is correct and compiles; consider closing it as a duplicate once #7022 lands to avoid redundant merges, but it is safe to approve independently.

Approving.

PR #7009, opened 2026-08-30
[type: bug] Fix Cryptor/Sign plugin removeRule to unconditionally cleanup (#6879)

### LINKED ISSUE #6879: [BUG] Cryptor JsonUtil.parser throws NPE/CCE on JSON-array or non-object bodies and missing nested-field segments — 500s on valid and pre-auth traffic
## Description
Both `removeRule` implementations gate the cache eviction behind `Optional.ofNullable(ruleData.getHandle()).ifPresent(s -> CACHED_HANDLE.get().removeHandle(...))`. If the rule-deletion event carries only the rule id (handle == null), `ifPresent` never fires and the cached `SignRuleHandler` / `CryptorRuleHandler` keyed by `selectorId_ruleId` is never evicted. Contrast `JwtPluginDataHandler.removeRule:56-58`, `KeyAuthPluginDataHandler.removeRule:45-47`, and `BasicAuthPluginDataHandler.removeRule:55-57`, which all call `removeHandle` unconditionally.

## Location
- `shenyu-plugin/shenyu-plugin-security/shenyu-plugin-sign/src/main/java/org/apache/shenyu/plugin/sign/handler/SignPluginDataHandler.java:47-49`
- `shenyu-plugin/shenyu-plugin-security/shenyu-plugin-cryptor/src/main/java/org/apache/shenyu/plugin/cryptor/handler/AbstractCryptorPluginDataHandler.java:60-62`

## Impact
Stale rule-handle objects accumulate in the `CommonHandleCache` indefinitely (memory leak). For cryptor, a stale `CryptorRuleHandler` (containing `decryptKey`/`encryptKey`/`strategyName`/`fieldNames`/`way`) persists; for sign, a stale `SignRuleHandler` (the `signRequestBody` flag) persists.

## Suggested fix
Drop the `getHandle()` guard and call `CACHED_HANDLE.get().removeHandle(CacheKeyUtils.INST.getKey(ruleData))` unconditionally (as JWT/key-auth/basic-auth already do), guarding only on `ruleData` itself.

## Related existing
None — distinct from NEW-06 (#6772, `removeRule` handle guard), which lists redirect/rewrite/param-mapping/modify-response/context-path. The sign and cryptor data handlers are the same bug class but were omitted from NEW-06's file list; they are separate plugins with separate cached handle types.

zhang-arvin:
Closes #6879


Aias00 (approved):
## Review: Fix Cryptor/Sign plugin removeRule to unconditionally remove cache (#6879)

**Verdict: APPROVE**

### Analysis

Both `AbstractCryptorPluginDataHandler.removeRule` and `SignPluginDataHandler.removeRule` previously guarded the cache removal with `Optional.ofNullable(ruleData.getHandle())`:
```java
Optional.ofNullable(ruleData.getHandle()).ifPresent(s -> CACHED_HANDLE.get().removeHandle(CacheKeyUtils.INST.getKey(ruleData)));
```
When `getHandle()` was `null` (e.g. the rule's handle was cleared or never set), the cache entry was **never removed**, leaving a stale `CryptorRuleHandler`/`SignRuleHandler` in `CACHED_HANDLE` even after the rule was deleted.

The fix removes the guard so removal always happens:
```java
CACHED_HANDLE.get().removeHandle(CacheKeyUtils.INST.getKey(ruleData));
```

**No NPE / safe-removal verification:**
- `CacheKeyUtils.getKey(RuleData)` builds the key as `String.join("_", ruleData.getSelectorId(), ruleData.getId())` — it does **not** read `getHandle()`, so a null handle cannot cause an NPE here.
- `CommonHandleCache.removeHandle` is `Optional.ofNullable(key).ifPresent(cached::remove)` — a no-op when the key is absent, so repeated/always-on removal is idempotent and safe.

Behavior change is exactly the intended one: a deleted rule now reliably clears its cached handler regardless of whether its handle payload is currently populated.

### Conclusion
Correct, minimal, and safe. The cache key derivation is independent of `handle`, and removal is idempotent. Approving.

PR #7010, opened 2026-08-30
[type: bug] Fix namespaceId null handling in AbstractNodeDataChangedListener (#6876)

### LINKED ISSUE #6876: [BUG] AbstractNodeDataChangedListener.onSelectorChanged builds config-key prefix with raw getNamespaceId() (no SYS_DEFAULT fallback) — selectors invisible when namespaceId null
## Description
Each plugin's `doExecute` calls `CACHED_HANDLE.get().obtainHandle(key)` and immediately dereferences the result without a null check. `CommonHandleCache.obtainHandle(key)` returns `cached.get(key)` which is `null` on cache miss. The cache miss occurs when a rule is created with a null/empty `handle` field — every data handler guards the cache write with `Optional.ofNullable(ruleData.getHandle()).ifPresent(...)`, so a rule with no handle JSON is never cached but CAN still be matched by `AbstractShenyuPlugin.execute` (matching is condition-based, not handle-based).

## Location
- `shenyu-plugin/shenyu-plugin-fault-tolerance/shenyu-plugin-hystrix/src/main/java/org/apache/shenyu/plugin/hystrix/HystrixPlugin.java:59-60`
- `shenyu-plugin/shenyu-plugin-fault-tolerance/shenyu-plugin-resilience4j/src/main/java/org/apache/shenyu/plugin/resilience4j/Resilience4JPlugin.java:66-67`
- `shenyu-plugin/shenyu-plugin-fault-tolerance/shenyu-plugin-sentinel/src/main/java/org/apache/shenyu/plugin/sentinel/SentinelPlugin.java:59-60`
- `shenyu-plugin/shenyu-plugin-fault-tolerance/shenyu-plugin-ratelimiter/src/main/java/org/apache/shenyu/plugin/ratelimiter/RateLimiterPlugin.java:72-74`
- `shenyu-plugin/shenyu-plugin-base/src/main/java/org/apache/shenyu/plugin/base/cache/CommonHandleCache.java:42-44`

## Impact
Every request matching a rule that was created without a handle (or whose handle was cleared) receives an unhandled `NullPointerException` / 500. A single misconfigured rule takes down all traffic matching it.

## Suggested fix
Add a null guard at the top of each `doExecute`, e.g. `if (Objects.isNull(handle)) { return chain.execute(exchange); }`, mirroring the guard added in `ContextPathPlugin` and `MockPlugin` (line 52-54).

## Related existing
Distinct from #6657 (GeneralContextPlugin NPE) — that issue was filed only for `GeneralContextPlugin.java:54`. These are four separate plugins with separate caches and separate `doExecute` implementations, all of which lack the guard that #6657 added for `GeneralContextPlugin` only.

zhang-arvin:
Closes #6876


Aias00 (approved):
## Review: Fix namespaceId null handling in AbstractNodeDataChangedListener (#6876)

**Verdict: APPROVE**

### Analysis

In `onSelectorChanged`, the `configKeyPrefix` was built directly from `selectorData.getNamespaceId()` with no null guard:
```java
final String configKeyPrefix = selectorData.getNamespaceId() + DefaultNodeConstants.JOIN_POINT + ...
```
When `namespaceId` is `null`, the prefix started with the literal `"null" + JOIN_POINT`, producing a wrong config key (and inconsistent with selectors that legitimately belong to the system default namespace).

The fix uses the same fallback already applied consistently throughout this class:
```java
final String configKeyPrefix = StringUtils.defaultString(selectorData.getNamespaceId(), SYS_DEFAULT_NAMESPACE_ID) + ...
```
- `SYS_DEFAULT_NAMESPACE_ID` is statically imported (`org.apache.shenyu.common.constant.Constants.SYS_DEFAULT_NAMESPACE_ID`) and is the same value used by `onSelectorChanged`'s sibling methods (lines 73/84/95/187/290/301 already use `StringUtils.defaultString(value.getNamespaceId(), SYS_DEFAULT_NAMESPACE_ID)`), so the key schema stays consistent.
- `org.apache.commons.lang3.StringUtils` is already imported.

No production caller relies on the previous `"null"`-prefixed key (that was a bug), and the rest of the `onCommonMultiChanged` flow is unchanged.

### Conclusion
Minimal, consistent with existing conventions, and removes a real null-handling bug. Approving.

zhang-arvin:

Aias00 (changes requested):

zhang-arvin:
@Aias00 Thanks for the review. The branch was 37 commits behind `master`, so I've rebased it onto the current `master` and force-pushed. There is no code change — the fix (`StringUtils.defaultString(selectorData.getNamespaceId(), SYS_DEFAULT_NAMESPACE_ID)` in `onSelectorChanged`, bringing it in line with the neighbouring `onRuleChanged`) is unchanged, and the rebase only clears the base drift.

Would you mind re-running the checks / taking another look?

Aias00 (approved):
Approving — this is a small but real correctness fix, and it makes the method consistent with the rest of the file.

`onSelectorChanged` built its config-key prefix directly from `selectorData.getNamespaceId()`. When that field was null (selector synced without a namespace — e.g. legacy data or a sync path that didn't populate it), the key became literally `null.<selectorId>.`, so the node wrote config under a key no reader would ever look up, and the selector silently failed to propagate.

`StringUtils.defaultString(..., SYS_DEFAULT_NAMESPACE_ID)` is exactly the right fix, and it is also *consistent*: I checked the branch and the same `defaultString(value.getNamespaceId(), SYS_DEFAULT_NAMESPACE_ID)` pattern is already used at lines 73, 84, 95, 187, 290 and 301 of this very class. Line 177 was simply the one place that missed it, which is a good sign this is a genuine oversight rather than a design choice.

Verified the symbols resolve on this branch: `org.apache.commons.lang3.StringUtils` (line 20), `DefaultNodeConstants` (line 21) and the static import of `Constants.SYS_DEFAULT_NAMESPACE_ID` (line 46) are all present, so no import changes were needed.

One non-blocking suggestion: this method now has no test. A case that pushes a `SelectorData` with a null `namespaceId` and asserts the resulting key starts with the default namespace id would pin the fix — the surrounding methods have similar coverage gaps, so this is a nice-to-have rather than a blocker.

Issue #6874, opened 2026-08-04
[BUG] instance_info BaseResultMap omits instance_state mapping — null state on resultMap reads (root of N37 NPEs)


Aias00:
## Description
In the REFRESH/MYSELF case, stale-entry removal is guarded by `configDataNames.size() > changedList.size()` (old count must strictly exceed new count). The actual cleanup (`configDataNames.removeAll(changeNames); configDataNames.forEach(this::delConfig)`) is only reached when old size > new size. When a REFRESH replaces the set with equal or larger cardinality but different members (e.g. old=[A,B,C,D], new=[C,D,E,F], 4 vs 4), the guard is false, so stale entries A and B are never deleted from nacos/apollo/polaris — only C,D,E,F are published and the new list overwrites the LIST node. The per-entry config nodes for A and B persist orphaned. This path is reached by `syncAllByNamespaceId(REFRESH, ns)` from `NamespacePluginController:210` and `ConfigsExportImportController:141`, which publish PLUGIN/AUTH/META REFRESH events spanning a whole namespace.

## Location
- `shenyu-admin-listener/shenyu-admin-listener-api/src/main/java/org/apache/shenyu/admin/listener/AbstractNodeDataChangedListener.java:129-140` (guard at :134)

## Impact
After a namespace resync or config import on nacos/apollo/polaris backends, deleted plugin / app-auth / metadata config entries whose count doesn't net-decrease remain in the config center. For app-auth this is security-adjacent: a deleted app key's config node can linger and be served to gateways that read it before the list update propagates.

## Suggested fix
Remove the size guard and always diff: `if (configDataNames != null) { configDataNames.removeAll(changeNames); configDataNames.forEach(this::delConfig); }` (the `removeAll` already yields the correct stale set regardless of sizes).

## Related existing
None — distinct from SYNC-3/SYNC-4 (gateway-side `shenyu-sync-data-http` refresh NPE on null/empty) and SYNC-5/SYNC-6 (gateway-side `DiscoveryUpstream`/`ProxySelectorRefresh` no-clear-on-empty). Those are gateway-side `shenyu-sync-data-center` refresh handlers; this is admin-side `AbstractNodeDataChangedListener` (nacos/apollo/polaris publish path) with a different root cause (faulty size guard vs NPE/no-clear-on-empty).

PR #7011, opened 2026-08-30
[type: bug] Fix stale config cleanup on REFRESH/MYSELF events (#6875)

### LINKED ISSUE #6875: [BUG] AbstractNodeDataChangedListener.onCommonChanged REFRESH stale-removal guarded by size> instead of diff — stale entries persist when |old|<=|new|
## Description
`onSelectorChanged` does `SelectorData selectorData = changed.stream().findFirst().orElseThrow(...); String configKeyPrefix = selectorData.getNamespaceId() + JOIN_POINT + changeData.getSelectorDataId() + JOIN_POINT;` — using `getNamespaceId()` directly with no null-default. Every sibling handler (`onAppAuthChanged:73-75`, `onPluginChanged:84-86`, `onMetaDataChanged:95-97`, `onRuleChanged:187-189`, `onProxySelectorChanged:276-278`, `onDiscoveryUpstreamChanged:287-289`) wraps the value with `StringUtils.defaultString(value.getNamespaceId(), SYS_DEFAULT_NAMESPACE_ID)` before building the prefix. If a `SelectorData` arrives with null `namespaceId`, the prefix becomes the literal `"null.selector."` instead of `"<default-namespace>.selector."`. The gateway-side nacos/apollo reader builds keys with the configured namespace id, so it reads `"<default>.selector."` and never sees the entries written under `"null.selector."` — selectors silently invisible to the gateway.

## Location
- `shenyu-admin-listener/shenyu-admin-listener-api/src/main/java/org/apache/shenyu/admin/listener/AbstractNodeDataChangedListener.java:176-177` (`onSelectorChanged`); compare siblings `:73-75,:84-86,:95-97,:187-189,:276-278,:287-289`

## Impact
Latent: any selector with a null `namespaceId` (partial config, legacy data, or a future code path that forgets to set it) is published under a wrong config key on nacos/apollo/polaris and is invisible to gateways, causing those routes to vanish while admin shows them as registered. The sibling methods defend against exactly this; selector alone does not.

## Suggested fix
Replace `selectorData.getNamespaceId()` at line 177 with `StringUtils.defaultString(selectorData.getNamespaceId(), SYS_DEFAULT_NAMESPACE_ID)` (and use the same streaming pattern as siblings to be consistent).

## Related existing
None — distinct from CLI-2 (#6664, client-side new-registrar never propagates namespaceId) and CLI-3 (rpcExt case mismatch). This is admin-side config-key construction for selector publish on node-based backends.

zhang-arvin:
Closes #6875

### DIFF
--- a/shenyu-plugin/shenyu-plugin-logging/shenyu-plugin-logging-rabbitmq/src/main/java/org/apache/shenyu/plugin/logging/rabbitmq/client/RabbitmqLogCollectClient.java
+++ b/shenyu-plugin/shenyu-plugin-logging/shenyu-plugin-logging-rabbitmq/src/main/java/org/apache/shenyu/plugin/logging/rabbitmq/client/RabbitmqLogCollectClient.java
@@ -142,12 +142,17 @@ public class RabbitmqLogCollectClient extends AbstractLogConsumeClient<RabbitmqL
             if (Objects.nonNull(channel)) {
                 channel.close();
             }
-            if (Objects.nonNull(connection)) {
-                connection.close();
-            }
-            LOG.info("close RabbitMQ connection success");
         } catch (IOException e) {
-            LOG.error("failed to close RabbitMQ connection", e);
+            LOG.error("failed to close RabbitMQ channel", e);
+        } finally {
+            try {
+                if (Objects.nonNull(connection)) {
+                    connection.close();
+                }
+                LOG.info("close RabbitMQ connection success");
+            } catch (IOException e) {
+                LOG.error("failed to close RabbitMQ connection", e);
+            }
         }
     }
 
--- a/shenyu-plugin/shenyu-plugin-logging/shenyu-plugin-logging-rabbitmq/src/test/java/org/apache/shenyu/plugin/logging/rabbitmq/client/RabbitmqLogCollectClientTest.java
+++ b/shenyu-plugin/shenyu-plugin-logging/shenyu-plugin-logging-rabbitmq/src/test/java/org/apache/shenyu/plugin/logging/rabbitmq/client/RabbitmqLogCollectClientTest.java
@@ -17,6 +17,8 @@
 
 package org.apache.shenyu.plugin.logging.rabbitmq.client;
 
+import com.rabbitmq.client.Channel;
+import com.rabbitmq.client.Connection;
 import org.apache.shenyu.common.dto.PluginData;
 import org.apache.shenyu.common.utils.GsonUtils;
 import org.apache.shenyu.plugin.logging.common.entity.ShenyuRequestLog;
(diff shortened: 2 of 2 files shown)


Aias00 (changes requested):
## Review: Fix stale config cleanup on REFRESH/MYSELF events (#6874)

**Verdict: REQUEST CHANGES**

### What's correct
Removing the faulty guard `configDataNames.size() > changedList.size()` is the right call — with equal cardinality the stale entries were never cleaned, so this condition must go.

### Problem: the cleanup still won't actually delete the stale nodes
`delConfig` is still called with the **bare** name:
```java
configDataNames.forEach(this::delConfig);   // bare name
```
But every publish used the **prefixed** key:
```java
changedList.forEach(changedData -> publishConfig(configKeyPrefix + mapperToKey.apply(changedData), changedData));
```
So `delConfig(name)` targets a different key than what `publishConfig` wrote (`configKeyPrefix + name`). The stale config nodes therefore remain in the registry. The size-guard removal alone does **not** achieve the PR's stated goal of "stale config cleanup on REFRESH/MYSELF".

### Overlap / duplication
This is essentially the same one-line fix already delivered (and approved) in **#7014**, which additionally corrects `delConfig` to receive the full key (`configKeyPrefix + staleName`), matching `publishConfig`. Please consolidate: either align this PR with #7014 (add the prefixed-key fix) or close it in favor of #7014.

### Required change
```java
configDataNames.forEach(name -> delConfig(configKeyPrefix + name));
```
(or reuse the same helper already introduced in #7014). Once the `delConfig` key matches the published key, the stale entries will be removed as intended.

### Conclusion
The guard removal is correct, but as written the PR does not fix stale cleanup because of the key mismatch, and it duplicates #7014. Requesting changes until the `delConfig` key is corrected (or the PR is consolidated with #7014).

zhang-arvin:
- Fixed delConfig to use prefixed key (configKeyPrefix + name) in stale config cleanup

Please take another look.

Aias00 (dismissed):
## Re-review: [type: bug] Fix stale config cleanup on REFRESH/MYSELF events (#6875)

Re-reviewed at head `5935927`. My previous REQUEST_CHANGES pointed out that `onCommonChanged` removed stale config nodes via `configDataNames.forEach(this::delConfig)`, passing the **bare** data name, while `publishConfig` writes under the `configKeyPrefix + name` key — so the stale nodes were never actually deleted. That is now fixed:

- Line 13: `configDataNames.forEach(name -> delConfig(configKeyPrefix + name));` — the full key (prefix + name) is now passed, matching how `publishConfig` keys the data. Stale config nodes are correctly removed on REFRESH/MYSELF.
- The redundant `configDataNames.size() > changedList.size()` guard is removed (line 10), so cleanup runs whenever there are leftover registered names, which is the intended behavior.

Verified against the head file: every `delConfig` call site (`AbstractNodeDataChangedListener` lines 126, 136, 225, 251, 264, 278) now uses the `configKeyPrefix + ...` form — no bare-name path remains.

Note: the same fix is also present in #7022 (which combines this with the fault-tolerance NPE fixes). This PR is correct on its own; consider coordinating/closing duplicates with #7022 to avoid merge conflicts, but it is safe to approve.

Approving.

Aias00:
@zhang-arvin pls fix conflict

zhang-arvin:

zhang-arvin:
@Aias00 conflict fixed — rebased onto latest master, thanks!

PR #7013, opened 2026-08-31
[type: bug] Fix RewritePlugin placeholder drops first character and StringIndexOutOfBounds (#6883)

### LINKED ISSUE #6883: [BUG] RewritePlugin placeholder branch drops first character of path-variable value + StringIndexOutOfBounds when regex lacks {

zhang-arvin:
Fixes #6883.

## Bug Description
The RewritePlugin has two bugs in its placeholder branch:

1. **First character of path-variable value is dropped**: `rewriteUri.substring(rewriteHandle.getRegex().indexOf("{") + 1)` skips the first character of the actual path-variable value because the `+1` is meant to skip the `{` character in the regex pattern, but it's applied to the rewriteUri which has the actual value at that position.

2. **StringIndexOutOfBoundsException when regex lacks `{`**: The condition only checks `rewriteHandle.getReplace().contains("{")` but not `rewriteHandle.getRegex().contains("{")`. When the replace pattern contains `{` but the regex doesn't, `indexOf("{")` returns -1, causing `substring(-1)` to throw.

## Fix
- Added `rewriteHandle.getRegex().contains("{")` to the guard condition
- Removed the `+1` offset from `rewriteUri.substring()` to preserve the first character


Aias00:
pls add my wechat: aias00

Aias00 (approved):
## Review: Fix RewritePlugin placeholder drops first character (#6883)

**Verdict: APPROVE**

### Analysis

The rewrite placeholder substitution uses `PathMatchUtils.replaceAll`, which performs a **literal** replacement:
```java
public static String replaceAll(final String path, final String regex, final String replacement) {
    return path.replaceAll(Pattern.quote(regex), Matcher.quoteReplacement(replacement));
}
```
The first arg is the template (`getReplace()`), the second is the literal placeholder segment extracted from `getRegex()` (`getRegex().substring(indexOf("{"))`), and the third is the actual value extracted from `rewriteUri`.

**Off-by-one fix (correct):**
- Old: `rewriteUri.substring(getRegex().indexOf("{") + 1)` — dropped the first character of the captured value. E.g. for `rewriteUri = /shenyu/123` with the `{...}` placeholder starting at index `i`, this produced `23` instead of `123`.
- New: `rewriteUri.substring(getRegex().indexOf("{"))` — keeps the full captured segment. This matches the prior fix in #6951 and is consistent.

**New crash guard (defensive, good):**
- Old condition only checked `getReplace().contains("{")`. If `getReplace()` contained `{` but `getRegex()` did **not**, then `getRegex().indexOf("{")` returned `-1`, and `substring(-1)` threw `StringIndexOutOfBoundsException`.
- New condition `getReplace().contains("{") && getRegex().contains("{")` routes the no-placeholder case to the safe `rewriteUri.replaceAll(getRegex(), getReplace())` branch, eliminating the crash.

### Conclusion
Minimal, correct, and aligns with the existing placeholder handling. No behavioral regression for the normal (no-placeholder) path. Approving.

PR #7014, opened 2026-08-31
fix: remove faulty size guard so REFRESH cleans up stale plugin/auth/meta config nodes

### LINKED ISSUE #6874: [BUG] instance_info BaseResultMap omits instance_state mapping — null state on resultMap reads (root of N37 NPEs)

dmsolr:
### What is the purpose of the change

Fixes #6874.

In `AbstractNodeDataChangedListener#onCommonChanged`, the REFRESH/MYSELF branch only removed stale plugin/app-auth/metadata config entries when the old set strictly outnumbered the new one (`configDataNames.size() > changedList.size()`). When a REFRESH replaced the set with equal or larger cardinality but different members (e.g. old=[A,B,C,D], new=[C,D,E,F]), the guard was false, so stale entries A and B were never removed from nacos/apollo/polaris config centers — only orphaned per-entry config nodes remained while the LIST node was overwritten.

This path is reached by `syncAllByNamespaceId(REFRESH, ns)` from `NamespacePluginController` and `ConfigsExportImportController`, which publish PLUGIN/AUTH/META REFRESH events spanning a whole namespace. For app-auth this is security-adjacent, since a deleted app key's config node could linger and be served to gateways that read it before the list update propagates.

### Changes
- Removed the faulty size guard so the stale-entry diff (`removeAll`) always runs.
- Fixed the stale-entry `delConfig` call to use the correctly namespaced key (`configKeyPrefix + name`) instead of the bare entry name, which was also silently preventing cleanup even when the guard passed.
- Added a regression test reproducing the exact scenario from the issue (old=[A,B,C,D] -> new=[C,D,E,F], equal cardinality) verifying stale entries are removed.

### Does this PR introduce a user-facing change?

No.


Aias00 (approved):
## Summary

Fixes #6874 — `AbstractNodeDataChangedListener.onCommonChanged` skipped stale-entry cleanup on REFRESH and deleted config under the wrong key.

## Changes

```java
- if (Objects.nonNull(configDataNames) && configDataNames.size() > changedList.size()) {
+ if (Objects.nonNull(configDataNames)) {
-     configDataNames.forEach(this::delConfig);
+     configDataNames.forEach(staleName -> delConfig(configKeyPrefix + staleName));
  }
```

1. **Removed the faulty size guard** `configDataNames.size() > changedList.size()`. It prevented deletion of stale entries whenever the existing config count was not strictly greater than the incoming change count. Example: existing `[A,B,C,D]` + `LIST=[A,B,C,D]`, then `REFRESH` with `[C,D,E,F]` — old guard `4 > 4` is false, so stale `A`,`B` were never deleted. Removing the guard ensures every name left after `removeAll(changeNames)` (i.e. genuinely stale) is cleaned.
2. **Fixed the delete key** from `delConfig(staleName)` to `delConfig(configKeyPrefix + staleName)`. `getConfigDataNames` returns bare names from `LIST_STR`, so the prefix must be re-applied — matching the `publishConfig(configKeyPrefix + ...)` / `delConfig(configKeyPrefix + ...)` convention used elsewhere (e.g. L126). The old code deleted a key without the prefix (wrong / non-existent) and left the real stale entry behind.

## Verification

- `delConfig(final String dataId)` (L320) takes the full config key; `getConfigDataNames` (L346) returns bare names — prepending `configKeyPrefix` is required and correct.
- Test `testOnPluginChangedRefreshWithEqualCardinalityRemovesStaleEntries` reproduces the equal-cardinality case and asserts `A`/`B` are deleted via `wasDeleted(configKeyPrefix + "A"/"B")`.
- Existing refresh test `testOnSelectorChangedRefreshReplacesPluginSelectorIds` remains valid.

## Conclusion

Two precise, well-tested correctness fixes for REFRESH stale-config cleanup. **Approved.**

PR #7017, opened 2026-08-31
fix: register and clean up API runtime resources(#6716)

### LINKED ISSUE #6716: [BUG] ApiServiceImpl removeRegister/register are dead code; delete() leaves orphan selector/rule/metadata
- severity: Medium
- files: `shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/ApiServiceImpl.java:159-216,220-233`
- description: `removeRegister(ApiDO)` (159-189) and `register(ApiDO)` (191-216) are private methods that are never invoked anywhere. `delete(...)` (220-233) removes rows from `api` and `tag_relation` only; it never calls `removeRegister`, so the associated `selector`, `rule`, and `metadata` rows are not cleaned up. Likewise `create(...)` does not call `register()`.
- impact: Deleting an API leaves orphaned gateway selector/rule/metadata that the runtime may still route on; creating an API does not auto-publish the corresponding selector/rule/metadata as the (now dead) `register()` method intended.
- suggested_fix: Either invoke `removeRegister` from within `delete(...)` and `register()` from `create(...)`, or delete the dead methods.
- confidence: Medium
- related_existing: none.

---

juicewcode:
Fixes #6716

  ## Summary

  - Invoke `register` after a successful API creation so the related selector/rule/metadata data is published.
  - Invoke `removeRegister` after successful API deletion.
  - Apply the same cleanup when deleting APIs by path, HTTP method, and RPC type.
  - Guard registration when required API extension data is missing.
  - Add unit tests for cleanup and non-existent API deletion.

  ## Test

  - Added `testDeleteCleansRegisteredResources` to verify rule, selector, and metadata cleanup.
  - Added `testDeleteWhenApiDoesNotExist` to verify the expected missing-API response.
  - Existing API create/update/find/list tests remain applicable.

Make sure that:


Aias00 (approved):
## Summary

Fixes #6716 — registers API runtime resources (selector/rule/metadata) at creation time and cleans them up on deletion, closing the lifecycle gap where resources were only registered on update.

## Changes

- `create()`: now calls `register(apiDO)` after inserting tag relations so the runtime resources are created up-front.
- `delete(...)` / `deleteByApiPathHttpMethodRpcType(...)`: now call `apis.forEach(this::removeRegister)` / `apiDOs.forEach(this::removeRegister)` after a successful delete to remove the associated selector/rule/metadata.
- `register(...)`: added a null/blank guard:
  ```java
  if (Objects.isNull(ext) || StringUtils.isBlank(apiDO.getContextPath())) {
      return;
  }
  ```
  This prevents an NPE when `apiDO.getExt()` is `null` (Gson `fromJson` returns `null`, then `ext.getXxx()` would throw) and avoids pointless registration when no context path is set.

## Verification

- Confirmed `apis` (L226) and `apiDOs` (L282) are declared in scope before the new `forEach` calls.
- Confirmed `import java.util.Objects;` (L24) and `import org.apache.commons.lang3.StringUtils;` (L23) are present.
- Confirmed `register(...)` (L192) and `removeRegister(...)` (L160) exist.
- Tests: `testDeleteCleansRegisteredResources` asserts `ruleService.deleteByIdsAndNamespaceId`, `selectorService.deleteByNamespaceId`, `metaDataService.deleteByIdsAndNamespaceId` are each invoked once; `testDeleteWhenApiDoesNotExist` asserts the `SYS_API_ID_NOT_EXIST` path.

## Conclusion

Correct and complete resource-lifecycle fix with good regression coverage. **Approved.**

PR #7019, opened 2026-08-31
fix: avoid concurrent modification in base data cache

### LINKED ISSUE #7018: [BUG] ConcurrentModificationException may occur when deleting selector or rule

### Current Behavior

`BaseDataCache` stores selector and rule data in:

- `ConcurrentMap<String, List<SelectorData>>`
- `ConcurrentMap<String, List<RuleData>>`

The map is concurrent, but its values are mutable `ArrayList` instances.

`removeSelectData` and `removeRuleData` modify the published list in place by
calling `removeIf` inside `synchronized (SELECTOR_MAP/RULE_MAP)`.

However, request threads obtain the same list instance through
`obtainSelectorData` and `obtainRuleData`, and iterate it without acquiring the
same lock in `AbstractShenyuPlugin#matchSelector` and `#matchRule`.

Therefore, when Admin synchronizes a selector/rule DELETE event while Bootstrap
is processing requests, the request thread may still throw
`ConcurrentModificationException`.

### Related issue

This is related to #4071 and PR #4155.

PR #4155 serialized cache writers, but readers do not acquire the same lock, so
it does not completely prevent concurrent modification of the shared list.

### Expected Behavior

Selector and rule cache reads should use stable snapshots. Deleting cache data
should create a new list and atomically replace the map value instead of
modifying a list already exposed to request threads.

### Steps To Reproduce

_No response_

### Environment

```markdown
```

### Debug logs

_No response_

### Proposed solution

Use a stable-snapshot/copy-on-write approach in `BaseDataCache`:

1. Obtain the current list inside the existing synchronized block.
2. Create a new list excluding the deleted selector or rule.
3. Atomically replace the map value with the new list.
4. Never structurally modify a list after it has been published to request
   threads.
5. Add regression tests verifying that a previously obtained selector/rule list
   remains unchanged after deletion.

This proposal does not require request threads to acquire a global lock and is
consistent with the existing update path, which already creates and replaces
lists.

TheoLi0905:
Fixes #7018

Publish immutable selector and rule list snapshots, and use atomic ConcurrentMap compute operations instead of modifying cached lists in place.

Tests:
- BaseDataCacheTest and CommonPluginDataSubscriberTest
- LocalPluginControllerTest

Make sure that:

- [x] You have read the contribution guidelines.
- [x] You submit test cases that back your changes.
- [x] Your local test passed ./mvnw clean install -Dmaven.javadoc.skip=true.


Aias00 (approved):
## Summary

Refactors `BaseDataCache` to eliminate `ConcurrentModificationException` by replacing `synchronized (MAP)` blocks with atomic `ConcurrentHashMap.compute` / `computeIfPresent`, and returning **immutable** `List.copyOf(...)` snapshots from `obtainSelectorData` / `obtainRuleData`.

## Verification

- **Thread safety**: `compute`/`computeIfPresent` operate atomically on the single key's bin (better concurrency than the previous global `synchronized(MAP)`), and each rebuilds a fresh local list — no shared mutable list is exposed.
- **Empty → key removed**: `removeSelectData`/`removeRuleData` now return `null` from the remapping function when the result is empty, which removes the map entry (previously left an empty list behind). Tests updated accordingly (`assertNull` instead of `assertThat(...).isEmpty()`).
- **No caller mutates the snapshot** (the key risk of making results immutable): checked every production caller of `obtainSelectorData`/`obtainRuleData` and `getSelectorMap`/`getRuleMap`:
  - `AbstractShenyuPlugin` — iterates only (L93/L111).
  - `ShenyuCacheRepository` — returns the list as-is, no mutation.
  - `LocalPluginController` — read-only / `Optional.orElse(emptyList())`; iterates `ruleDataList` only.
  - `ShenyuControllerEndpoint` — returns the map directly, no mutation.
- **Import cleanup**: `com.google.common.collect.Lists` removed and replaced with `java.util.ArrayList` / `java.util.Objects`; confirmed no remaining `Lists.` usages in the head file.
- **Tests**: added snapshot-stability + immutability (`assertThrows(UnsupportedOperationException.class, () -> snapshot.add(...))`) assertions; existing tests adjusted.

## Conclusion

Clean, well-scoped, behavior-preserving (except the intentional empty-list→null cleanup) concurrency fix with solid test coverage. **Approved.**

Aias00:
pls add my wechat: aias00

PR #7022, opened 2026-09-01
[type:bug] Fix AiProxy switching providers after partial SSE output

### LINKED ISSUE #7021: [BUG] AiProxy switches providers after partial SSE output and mixes stream content
## Description

`AiProxyExecutorService.executeDirectStream` applies retry to the main provider Flux and then unconditionally switches to the fallback provider through `onErrorResume`.

If the main provider emits one or more chunks before failing, those chunks have already been written to the client SSE response. The fallback provider then starts a new completion from the beginning and appends its chunks to the already committed response.

This issue is distinct from #6647:

- #6647 tracks replaying the main upstream after a mid-stream retry.
- This issue tracks switching to a different fallback provider after partial output, producing one SSE response containing content from multiple completions/providers.

Related PR: #6341, reviewed at head `8fa93d7446e052f149d3740bbe7ff2052c375e23`.

## Affected code

- `shenyu-plugin-ai-proxy/.../AiProxyExecutorService.java`: `retryWhen(...).onErrorResume(...fallback...)` wraps the whole streaming Flux.
- `shenyu-plugin-ai-proxy/.../AiProxyPlugin.java`: the Flux is written directly to the committed SSE response.

## Reproduction

An equivalent Reactor sequence:

```text
main provider: emit "main-partial" -> error
retry:         emit "main-partial" -> error
fallback:      emit "fallback-completion"
```

produces:

```text
main-partial, main-partial, fallback-completion
```

Even after #6647 removes the retry replay, an unconditional fallback after the first emitted chunk would still produce:

```text
main-partial, fallback-completion
```

## Impact

- Clients receive a syntactically valid SSE stream with semantically corrupted, mixed completion content.
- Tool calls or structured output can be duplicated or combined across providers.
- Multiple providers may bill for the same client request.
- The response is already committed, so the gateway cannot safely replace earlier chunks.

## Expected behavior

Provider retry and fallback are allowed only before the first response chunk is emitted. After any chunk is sent, an upstream failure must terminate the current stream without starting another provider completion.

## Acceptance criteria

- [ ] Track whether the main stream has emitted a chunk.
- [ ] Allow retry/fallback only before the first emission.
- [ ] After partial emission, propagate/terminate the stream without invoking fallback.
- [ ] Add a regression test where the main provider emits a chunk and then errors while fallback is configured.
- [ ] Assert no duplicated prefix and no fallback chunk is appended after partial output.
- [ ] Retain the existing error-before-first-chunk fallback behavior.
- [ ] Coordinate the retry portion with #6647 rather than implementing two competing fixes.

zhang-arvin:
Fixes #7021

## Problem
When the main AI provider emits partial SSE content and then fails mid-stream, the `executeStream` method uses `retryWhen(Retry.max(1))` to retry the main provider. This causes:
1. The main provider re-starts streaming from the beginning, emitting new content
2. After retries are exhausted, the fallback provider also emits content
3. Result: the client receives mixed content from multiple providers

## Fix
Remove the `retryWhen` from the streaming path. For streaming, retrying after partial content has already been emitted to the client is always problematic. Instead, immediately fall back to the fallback provider on any error.

## Changes
- `shenyu-plugin/shenyu-plugin-ai/shenyu-plugin-ai-proxy/src/main/java/org/apache/shenyu/plugin/ai/proxy/enhanced/service/AiProxyExecutorService.java`: Removed `retryWhen` from `executeStream` method, replaced with direct `onErrorResume` to fallback provider


Aias00 (changes requested):
## Request Changes — two blocking issues

This PR bundles several fixes that overlap with other open PRs (#7008 fault-tolerance NPE guards, #7009 cryptor/sign `removeRule`, #7011 size-guard removal). That consolidation is fine, but it inherits two defects that must be fixed before merge.

### Blocker 1 — `RateLimiterPlugin` does not compile (import missing)

`RateLimiterPlugin.doExecute` adds `if (Objects.isNull(limiterHandle)) { return chain.execute(exchange); }` but the file does **not** import `java.util.Objects` (verified on the head branch: the only `Objects` occurrence is the new usage at line 74, no import line). This is the exact same compile error already flagged on #7008. Add `import java.util.Objects;` to `RateLimiterPlugin.java`.

### Blocker 2 — stale-config cleanup still passes bare names to `delConfig` (incomplete, same as #7011)

In `AbstractNodeDataChangedListener.onCommonChanged`, the REFRESH branch is now:

```java
final List<String> configDataNames = this.getConfigDataNames(configKeyPrefix);
if (Objects.nonNull(configDataNames)) {
    configDataNames.forEach(this::delConfig);
}
```

`getConfigDataNames` (line 347) returns the bare names parsed from the `LIST` node (`configKeyPrefix + LIST_STR`), and `changeNames` are also bare names (`changedList.map(mapperToKey)`). So `delConfig` is invoked with **bare names**. But the actual entries — and the DELETE branch at line 126, and `delConfig` itself — use the full key `configKeyPrefix + name`. As a result, stale entries published at `configKeyPrefix + name` are **not** deleted. This is the same incompleteness as #7011; #7014 already fixed it by calling `delConfig(configKeyPrefix + name)`. Please apply the same prefix fix here:

```java
configDataNames.forEach(name -> delConfig(configKeyPrefix + name));
```

### Non-blocking note — `AiProxyExecutorService` behavior change

`.retryWhen(Retry.max(1))…onErrorResume(NonTransientAiException…)` is replaced with a single `.onErrorResume(throwable -> handleFallbackStream(...))`. This drops the one retry and falls back immediately on any error. That matches the PR goal ("switch providers after partial SSE output"), but it does reduce resilience — please confirm the retry removal is intentional and not needed for transient upstream failures.

### Coordination

Since this PR duplicates #7008 / #7009 / #7011, please coordinate so we don't end up with duplicate or conflicting merges (either close the superseded individual PRs or keep only this consolidated one).

## Conclusion

**Request changes** pending the `Objects` import fix and the `delConfig(configKeyPrefix + name)` prefix fix. The fault-tolerance / cryptor-sign / controller ideas themselves are sound.

zhang-arvin:
- Added missing `import java.util.Objects;` in RateLimiterPlugin.java
- Fixed delConfig to use prefixed key (configKeyPrefix + name) in stale config cleanup

Please take another look.

Aias00 (dismissed):
## Re-review: [type:bug] Fix AiProxy switching providers after partial SSE output

Re-reviewed at head `773b5b5`. My previous REQUEST_CHANGES identified two hard blockers; both are resolved, plus the related fault-tolerance NPE fixes are included:

**1. RateLimiterPlugin missing `import java.util.Objects;` (was a compile error) — FIXED**
`RateLimiterPlugin.java` line 69 adds `import java.util.Objects;` (verified: exactly one occurrence, no duplicate) and a null-handle guard (lines 77-79). Module now compiles.

**2. `onCommonChanged` `delConfig` passed bare name — FIXED**
Line 13: `configDataNames.forEach(name -> delConfig(configKeyPrefix + name));` uses the full key, matching how `publishConfig` writes data, so stale config nodes are actually removed on REFRESH/MYSELF. Verified every `delConfig` call site in the class uses the `configKeyPrefix + ...` form (lines 126, 136, 225, 251, 264, 278) — no bare-name path remains.

**3. Additional fixes bundled in this combined PR**
- `HystrixPlugin` / `Resilience4JPlugin` / `SentinelPlugin`: null-handle guards (safe fall-through to `chain.execute`).
- `AbstractCryptorPluginDataHandler.removeRule` / `SignPluginDataHandler.removeRule`: unconditional `removeHandle` (drops the `Optional.ofNullable(ruleData.getHandle())` guard that skipped cache clearing when handle was null).
- `onSelectorChanged`: null-safe namespace prefix via `StringUtils.defaultString(getNamespaceId(), SYS_DEFAULT_NAMESPACE_ID)`.
- `AiProxyExecutorService.executeStream` (lines 30-43): replaced the `Retry.max(1)` + custom `NonTransientAiException` + `onErrorResume(NonTransientAiException.class, ...)` chain with a single `onErrorResume(throwable -> handleFallbackStream(...))`.

**Non-blocking note on the AiProxy change:** removing the bounded `Retry.max(1)` means the main stream is no longer retried once before falling back — any error (including transient ones) immediately triggers provider switch. For the stated goal ("switching providers after partial SSE output") this is reasonable, but if you want resilience against transient upstream blips you could re-add a small bounded retry before the fallback. Not required for approval.

All blockers resolved. Approving.

Aias00:
@zhang-arvin hi, pls fix conflict

zhang-arvin:

zhang-arvin:
@Aias00 conflict fixed — rebased onto latest master, thanks!

Aias00 (changes requested):

zhang-arvin:
@Copilot Thanks for the review — you're right on both points, and the second one was a real defect I had missed.

**1. Unrelated namespace change removed.** You were correct that the selector key change (`AbstractNodeDataChangedListener`) had nothing to do with AI proxy SSE fallback. It also turned out to be a *duplicate*: it is already carried by its own PR, which I've verified byte-for-byte (identical blob `b8231212` for the file). I've dropped it from this PR entirely — the branch now contains only the `AiProxyExecutorService` change plus tests, so this PR is scoped exclusively to #7021.

**2. The fallback was still invoked after partial output — fixed.** This was a genuine bug, not just a missing test. `handleDirectFallbackStream` unconditionally switched to the fallback provider on *any* error, including an error raised after one or more chunks had already been emitted — which is exactly the `main-partial + fallback-completion` corruption reported in #7021.

Fix: `executeDirectStream` now tracks whether any chunk has been emitted (`AtomicBoolean` set in `doOnNext`). `handleDirectFallbackStream` takes that flag and, if a chunk was already delivered, propagates the error instead of subscribing to the fallback:

```java
if (chunkAlreadyEmitted) {
    LOG.warn("Main direct stream failed after emitting at least one chunk; "
            + "not switching to fallback to avoid mixing partial and fallback output.", throwable);
    return Flux.error(throwable);
}
```

**3. Regression test added**, covering the case you asked for: a main stream that emits a chunk and then errors, with a fallback configured, must emit only the main chunk, propagate the error, and **never subscribe to the fallback** (`verify(fallbackApi, never()).chatCompletionStream(any())`). I also added the complementary case (error with zero chunks emitted still falls back).

**4. One more defect found while testing:** removing the retry wrapper in this PR also removed its `NonTransientAiException` repackaging, which invalidated the existing `testExecuteDirectStreamErrorWithNoFallback` assertion — the suite failed for real (`expected NonTransientAiException, actual RuntimeException`). I confirmed `AiProxyPlugin` only logs on error and does not depend on that exception type, so the production behaviour is fine; the assertion is now aligned with the propagated error type.

`./mvnw -pl shenyu-plugin/shenyu-plugin-ai/shenyu-plugin-ai-proxy test -Dtest=AiProxyExecutorServiceTest` → **Tests run: 8, Failures: 0, Errors: 0**.

@Aias00 @dengliming the branch is rebased on current `master` and conflicts are resolved. Would you mind taking another look?

Aias00 (changes requested):
The core diagnosis is right — and I agree the fallback must not engage after a chunk has been emitted — but I'm keeping this at REQUEST_CHANGES because of how the fix gets there.

**1. Please don't drop `retryWhen` from `executeDirectStream`.**

The PR deletes the entire `retryWhen(Retry.max(1).filter(AiProxyExecutorService::isRetryable)...)` block. That conflates two different concerns: "don't retry/fallback *after* emission" (the actual #7021 bug) and "don't retry at all" (a much broader change). A transient failure that occurs *before* any chunk is emitted is still perfectly safe to retry — nothing has reached the client — and that capability is being removed silently.

Note there is now an approved PR (#7205) that fixes the same issue by keeping the retry and simply gating it:

```java
return Flux.defer(() -> {
    AtomicBoolean emitted = new AtomicBoolean();
    return mainApi.chatCompletionStream(request)
            .doOnNext(chunk -> emitted.set(true))
            .retryWhen(Retry.max(1).filter(error -> !emitted.get() && isRetryable(error)) ...)
            .onErrorResume(error -> emitted.get()
                    ? Flux.error(error)
                    : handleDirectFallbackStream(...));
});
```

I'd rather see this PR aligned with that shape — it fixes the reported corruption *and* preserves retry for the pre-emission case. If you'd prefer to keep this PR minimal, please at least state explicitly in the description that dropping retry is intentional.

**2. `chunkEmitted` should be per-subscription, not per-invocation.**

`final AtomicBoolean chunkEmitted = new AtomicBoolean(false);` is allocated in the method body, outside the returned `Flux`. `chatCompletionStream` returns a cold `Flux` that can be subscribed more than once; if that ever happens, the second subscription starts with the flag already `true` from the first, and fallback would be suppressed incorrectly. Wrapping the body in `Flux.defer(...)` as #7205 does makes the flag subscription-scoped.

**Everything else looks good:**
- The `emitted ? Flux.error(throwable) : fallback` decision, with a `LOG.warn` explaining why the fallback was skipped, is the right behaviour and good for operability.
- The comment above the flag clearly states the rationale for future readers.
- Both tests are meaningful and cover both sides: `testExecuteDirectStreamDoesNotFallbackAfterChunkEmitted` asserts the fallback is *never* subscribed after one chunk, and `testExecuteDirectStreamFallbackAfterErrorWithoutChunks` asserts fallback still works when nothing was emitted. That second test is important — it's the one that catches an over-broad fix.
- `isRetryable` is still used elsewhere in the class (line 91), so removing it from this path doesn't leave dead code.

Address items 1 and 2 and I'll approve immediately.

PR #7023, opened 2026-09-01
fix: reject AI proxy API key creation for non-existent selectors(#6715)

### LINKED ISSUE #6715: [BUG] AiProxyApiKeyController.create does not enforce selector existence, allowing orphan proxy-api-key records
- severity: Medium
- files: `shenyu-admin/src/main/java/org/apache/shenyu/admin/controller/AiProxyApiKeyController.java:75-88`; `shenyu-admin/src/main/java/org/apache/shenyu/admin/service/impl/AiProxyApiKeyServiceImpl.java:78-110`
- description: In `create` (line 78), `selectorMapper.selectById(selectorId)` is fetched but only used to *override* `dto.namespaceId` when the selector exists; if `selector` is null, execution continues and `aiProxyApiKeyService.create(dto, selectorId)` is still called. The service only checks `StringUtils.isBlank(selectorId)`, never verifying the selector exists. Inconsistent with the `list` endpoint (line 109-112) which explicitly rejects a missing selector.
- impact: Orphan AI proxy key records referencing deleted/never-existed selectors; `realKeyResolver` lookups return null real keys, confusing gateway routing of AI traffic.
- suggested_fix: Return `AdminConstants.ID_NOT_EXIST` when `selector` is null (as `list` does).
- confidence: High
- related_existing: none.

---

juicewcode:
Fixes #6715
## Summary

  - Reject AI proxy API key creation when the referenced selector does not exist.
  - Return `AdminConstants.ID_NOT_EXIST` before invoking the service.
  - Preserve selector-based namespace derivation for valid selectors.
  - Add controller unit tests for both valid and invalid selector cases.

  ## Test

  - Added `AiProxyApiKeyControllerTest`.
  - Covers:
    - Missing selector returns `ID_NOT_EXIST` and does not call the create service.
    - Existing selector overwrites the DTO namespace and delegates creation.
  - Tests were not executed, per task requirements; static checks were completed.

Make sure that:

### DIFF
--- a/shenyu-admin/src/main/java/org/apache/shenyu/admin/controller/AiProxyApiKeyController.java
+++ b/shenyu-admin/src/main/java/org/apache/shenyu/admin/controller/AiProxyApiKeyController.java
@@ -76,9 +76,10 @@ public class AiProxyApiKeyController implements PagedController<ProxyApiKeyQuery
             @Valid @RequestBody final ProxyApiKeyDTO dto) {
         // derive namespaceId from selector to avoid mismatch
         final SelectorDO selector = selectorMapper.selectById(selectorId);
-        if (Objects.nonNull(selector)) {
-            dto.setNamespaceId(selector.getNamespaceId());
+        if (Objects.isNull(selector)) {
+            return ShenyuAdminResult.error(AdminConstants.ID_NOT_EXIST);
         }
+        dto.setNamespaceId(selector.getNamespaceId());
         int rows = aiProxyApiKeyService.create(dto, selectorId);
         if (rows > 0) {
             final ProxyApiKeyVO vo = aiProxyApiKeyService.findById(dto.getId());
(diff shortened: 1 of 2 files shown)


Aias00 (approved):
## Summary

Good catch — the original code only derived `namespaceId` inside the `Objects.nonNull(selector)` branch but still proceeded to call the service with a potentially `null` selector context, which could create an API key bound to the wrong/empty namespace or NPE downstream. The fix correctly rejects the request up-front when the referenced selector does not exist.

## Verification

- `AdminConstants.ID_NOT_EXIST = "ID NOT EXIST!"` exists; `ShenyuAdminResult.error(...)` propagates it as the message, so the new test assertion `assertEquals(AdminConstants.ID_NOT_EXIST, result.getMessage())` is valid.
- `SelectorDO.builder().namespaceId(...).build()` is valid — `SelectorDO` has a `namespaceId` field + builder.
- `ProxyApiKeyDTO` has `getNamespaceId()`/`setNamespaceId(...)`.
- `Objects` was already imported in `AiProxyApiKeyController` (the old code used `Objects.nonNull`), so the switch to `Objects.isNull` compiles.
- The new `AiProxyApiKeyControllerTest` covers both paths: reject-when-selector-missing and derive-namespace-and-delegate-when-present.

## Conclusion

Minimal, correct defensive fix with regression coverage. **Approve.**

PR #7024, opened 2026-09-01
[type:bug] Fix AiProxy forwarding internal fallbackConfig to upstream (#7020)

### LINKED ISSUE #7020: [BUG] AiProxy forwards internal fallbackConfig and credentials to primary upstream
## Description

PR #6341 changes AiProxy to build Spring AI `ChatCompletionRequest` directly from the raw client JSON. ShenYu's request-only `fallbackConfig` field is not removed before conversion.

Spring AI 1.1.2 stores unknown request fields in `ChatCompletionRequest.extraBody` through `@JsonAnySetter` and serializes them through `@JsonAnyGetter`. As a result, the gateway forwards the internal fallback configuration to the primary model provider before fallback is needed.

## Affected code

- `shenyu-plugin-ai-common/.../OpenAiProtocolAdapter.java`: raw request tree is converted without removing `fallbackConfig`.
- `shenyu-plugin-ai-proxy/.../AiProxyConfigService.java`: the same field is intentionally parsed as ShenYu dynamic fallback metadata.
- `shenyu-plugin-ai-proxy/.../AiProxyPlugin.java`: the resulting request is sent through `OpenAiApi` for both streaming and non-streaming calls.

Related PR: #6341, reviewed at head `8fa93d7446e052f149d3740bbe7ff2052c375e23`.

## Reproduction

Given:

```json
{
  "model": "main",
  "messages": [{"role": "user", "content": "hi"}],
  "fallbackConfig": {
    "baseUrl": "https://fallback.example",
    "apiKey": "fallback-secret",
    "model": "fallback-model"
  }
}
```

A local capture of the actual `OpenAiApi.chatCompletionEntity` HTTP request body contained both:

```json
"fallbackConfig": {
  "baseUrl": "https://fallback.example",
  "apiKey": "fallback-secret",
  "model": "fallback-model"
}
```

and the same object under `extra_body`.

## Impact

- Fallback provider credentials and routing metadata are disclosed to the primary provider.
- Strict OpenAI-compatible providers may reject requests containing ShenYu-only fields.
- Requests using dynamic fallback can fail before the fallback path is triggered.

## Expected behavior

Gateway-only metadata must never be serialized into upstream model requests. Dynamic fallback resolution should continue to use the field internally.

## Acceptance criteria

- [ ] Strip `fallbackConfig` and any other ShenYu-only request fields before creating `ChatCompletionRequest`.
- [ ] Apply the sanitization to both primary and fallback request construction.
- [ ] Add a regression test that captures or serializes the outbound request and asserts `fallbackConfig`, `baseUrl`, and the fallback API key are absent.
- [ ] Verify dynamic fallback still resolves and executes with its configured provider/model/key.
- [ ] Cover streaming and non-streaming request paths.

zhang-arvin:
Fixes #7020

## What changes were proposed in this pull request?

AiProxy's `extractPrompt` method forwards the entire request body to the upstream AI provider when no `prompt` or `content` field is found at the top level. This includes the internal `fallbackConfig` field with credentials (provider, apiKey, etc.), leaking sensitive configuration to the upstream.

## Changes

- Modified `AiProxyConfigService.extractPrompt()` to strip the `fallbackConfig` field from the JSON body before forwarding to upstream
- Updated test `testExtractPromptWithFallbackConfigButNoPromptOrContent` to verify the fix

## How was this patch tested?

- Unit tests: `AiProxyConfigServiceTest` (19/19 pass)
- Verified that `fallbackConfig` is removed from the forwarded body when present alongside `messages`/other fields

### DIFF
--- a/shenyu-plugin/shenyu-plugin-ai/shenyu-plugin-ai-proxy/src/main/java/org/apache/shenyu/plugin/ai/proxy/enhanced/service/AiProxyConfigService.java
+++ b/shenyu-plugin/shenyu-plugin-ai/shenyu-plugin-ai-proxy/src/main/java/org/apache/shenyu/plugin/ai/proxy/enhanced/service/AiProxyConfigService.java
@@ -18,6 +18,7 @@
 package org.apache.shenyu.plugin.ai.proxy.enhanced.service;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import org.apache.shenyu.common.dto.convert.rule.AiProxyHandle;
 import org.apache.shenyu.common.utils.JsonUtils;
 import org.apache.shenyu.common.utils.Singleton;
@@ -122,6 +123,11 @@ public class AiProxyConfigService {
             }
             if (jsonNode.has("content")) {
                 return jsonNode.get("content").asText();
+            }
+            // Strip internal fallbackConfig from body before forwarding to upstream
+            if (jsonNode.has(FALLBACK_CONFIG) && jsonNode.isObject()) {
+                ((ObjectNode) jsonNode).remove(FALLBACK_CONFIG);
+                return jsonNode.toString();
             }
         } catch (Exception e) {
             // ignore parsing errors and fall back to original body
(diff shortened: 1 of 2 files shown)


Aias00 (approved):
## Summary

Good security fix that closes a credential-leak path to the upstream AI provider.

## What is correct

- When the request body has no top-level `prompt`/`content` (i.e., the entire body is forwarded to the upstream), the internal `fallbackConfig` field — which can contain provider/apiKey credentials — was previously leaked. The fix strips it via `((ObjectNode) jsonNode).remove(FALLBACK_CONFIG)` before forwarding.
- Verified in the head file: `jsonNode` is freshly parsed per call (`JsonUtils.toJsonNode(requestBody)`), so the in-place `remove` mutation cannot affect any shared state. The `jsonNode.isObject()` guard protects the `ObjectNode` cast, and the `FALLBACK_CONFIG` constant is defined (line 39).
- The `prompt`/`content` branches return only those extracted fields, so they never forwarded `fallbackConfig` in the first place — the leak was exclusively in the full-body forwarding path, which is the one now fixed.
- Test updated to assert `fallbackConfig` is stripped while `messages` is preserved.

## Verdict

Approve.

## Non-blocking note

If a top-level JSON array body (unusual for AI request shapes) were to carry `fallbackConfig`, the `has()`/`isObject()` checks would skip stripping. Out of scope for normal request shapes, just flagging the edge.

Issue #6603, opened 2026-08-02
[Task] apache-rat-plugin configured but never executed in CI + excldue typo


Aias00:
## Description
rat:check is bound to `verify` but CI stops at `test`, so it's dead. Separately `pom.xml:828` misspells `exclude` as `excldue` — Maven silently ignores it, so uppercase `.MD` files would be flagged as unlicensed if rat ever ran.

## Location
```
pom.xml:791-855 (rat bound to verify), pom.xml:828 (<excldue>**/*.MD</excldue>)
.github/workflows/ci.yml:125-137 (clean test — never reaches verify)
```

## Impact
License-file policy unenforced in PR CI; latent typo.

## Suggested fix
Run `mvn verify` (or `rat:check` explicitly) in `ci.yml`, or remove the dead rat config; fix `excldue`->`exclude`.

## Related existing issue(s)
#6083 is checkstyle-pom-license — this is a different mechanism (apache-rat-plugin) and the typo is unique.

_Identified during the 2026-08-02 audit; full list in [`docs/issue-candidates-2026-08-02.md`](docs/issue-candidates-2026-08-02.md)._

PR #7028, opened 2026-09-02
Fix excldue typo in apache-rat exclude patterns

tanvir-ux:
## What
Fixed `<excldue>` → `<exclude>` in the apache-rat plugin config in `pom.xml` and `shenyu-e2e/pom.xml`.

Maven silently ignores unknown tags, so `**/*.MD` was never excluded from license checks.

## Why
Part of #6603 — this PR only fixes the typo; I left the CI verify/rat:check change out so this stays a small, safe diff.

## Test
Diff-only change to POM exclude tags; no runtime behavior change.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -825,7 +825,7 @@
                         <exclude>**/.gitpod.yml</exclude>
                         <!-- document files -->
                         <exclude>**/*.md</exclude>
-                        <excldue>**/*.MD</excldue>
+                        <exclude>**/*.MD</exclude>
                         <!-- wasm build files -->
                         <exclude>**/Cargo.lock</exclude>
                         <exclude>**/go.sum</exclude>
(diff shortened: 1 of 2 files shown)


Aias00 (approved):
## Summary

Trivial but correct fix: `<excldue>` → `<exclude>` in the Apache RAT license-check plugin configuration, in both the root `pom.xml` and `shenyu-e2e/pom.xml`.

## What is correct

- Maven silently ignores unknown configuration tags, so the original `<excldue>**/*.MD</excldue>` was a no-op — `**/*.MD` files were never excluded from RAT license checks. Restoring `<exclude>` makes the intended exclusion effective again.
- Change is purely declarative (POM exclude tags); no runtime/compiled-behavior change, no functional risk.
- Scope is intentionally small (the author explicitly left the CI `rat:check` enablement out), which keeps this a safe, low-risk diff.

## Verdict

Approve. No blocking issues.

## Non-blocking suggestions

- Consider following up in #6603 with the actual `rat:check` CI enablement so the typo can't silently regress again.

PR #7029, opened 2026-09-03
fix: invoke RedisRateLimiter error callback/logging inside onErrorResume (#6645)

### LINKED ISSUE #6645: [BUG] RedisRateLimiter silently swallows Redis errors and fail-opens (dead doOnError)
- severity: High
- files: `shenyu-plugin/shenyu-plugin-fault-tolerance/shenyu-plugin-ratelimiter/src/main/java/org/apache/shenyu/plugin/ratelimiter/executor/RedisRateLimiter.java:63-75`
- description: `resultFlux.onErrorResume(throwable -> Flux.just(Arrays.asList(1L, -1L)))` converts any Redis error into a normal emission of `{allowed=1, tokensLeft=-1}`. The downstream `.reduce().map()` completes successfully, so the trailing `.doOnError(...)` (which logs and calls the callback) is unreachable — `onErrorResume` already absorbed the error. Net effect: on Redis outage every request is allowed (fail-open) AND nothing is logged.
- impact: Complete, invisible bypass of rate limiting during Redis outages; operators get no log signal.
- suggested_fix: Move logging into the `onErrorResume` lambda; make fail-open vs fail-closed an explicit, logged policy.
- confidence: High
- related_existing: none — #5376/#5340/#5021 are different.

---

dmsolr:
Fixes #6645

### Root cause
In `RedisRateLimiter.isAllowed`, `onErrorResume` absorbed every Redis error into a normal `Flux.just(1L, -1L)` emission, so the trailing `.doOnError(...)` (which logged the error and invoked the rate limiter algorithm's `callback`, e.g. cleanup of optimistically-written keys in `ConcurrentRateLimiterAlgorithm`) was unreachable dead code. On a Redis outage, every request was silently fail-opened with no log signal and no cleanup callback.

### Fix
Move the algorithm callback invocation and error logging into the `onErrorResume` lambda itself, so they actually execute when a Redis error occurs. Removed the now-dead `.doOnError(...)`.

### Testing
- Added `allowedThrowableInvokesAlgorithmCallbackTest` in `RedisRateLimiterTest`, using the `concurrent` algorithm, asserting the algorithm's error callback (Redis zset cleanup) is invoked when the Redis call errors. This test fails against the old code and passes after the fix.
- `./mvnw -pl shenyu-plugin/shenyu-plugin-fault-tolerance/shenyu-plugin-ratelimiter -am test`: checkstyle 0 violations, full reactor build success, all tests pass (including 8/8 in `RedisRateLimiterTest`).

- [x] You have read the contribution guidelines.
- [x] Your local test passed `./mvnw clean install -Dmaven.javadoc.skip=true` (module-scoped).

### DIFF
--- a/shenyu-plugin/shenyu-plugin-fault-tolerance/shenyu-plugin-ratelimiter/src/main/java/org/apache/shenyu/plugin/ratelimiter/executor/RedisRateLimiter.java
+++ b/shenyu-plugin/shenyu-plugin-fault-tolerance/shenyu-plugin-ratelimiter/src/main/java/org/apache/shenyu/plugin/ratelimiter/executor/RedisRateLimiter.java
@@ -60,7 +60,13 @@ public class RedisRateLimiter {
         List<String> keys = rateLimiterAlgorithm.getKeys(id);
         List<String> scriptArgs = Stream.of(replenishRate, burstCapacity, Instant.now().getEpochSecond(), requestCount).map(String::valueOf).collect(Collectors.toList());
         Flux<List<Long>> resultFlux = Singleton.INST.get(ReactiveRedisTemplate.class).execute(script, keys, scriptArgs);
-        return resultFlux.onErrorResume(throwable -> Flux.just(Arrays.asList(1L, -1L)))
+        // the error is absorbed right here, so callback/logging must happen in this lambda, not in a downstream doOnError
+        return resultFlux
+                .onErrorResume(throwable -> {
+                    rateLimiterAlgorithm.callback(rateLimiterAlgorithm.getScript(), keys, scriptArgs);
+                    LOG.error("Error occurred while judging if user is allowed by RedisRateLimiter, fail-open and allow the request:{}", throwable.getMessage());
+                    return Flux.just(Arrays.asList(1L, -1L));
+                })
                 .reduce(new ArrayList<Long>(), (longs, l) -> {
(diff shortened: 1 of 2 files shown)


Aias00 (approved):
Thanks @dmsolr! 

The root-cause analysis is correct: `onErrorResume` absorbs the Redis error before it can reach the trailing `.doOnError(...)`, so the algorithm callback and error logging were unreachable dead code — on a Redis outage every request silently fail-opened.

The fix is sound:

- Moving `rateLimiterAlgorithm.callback(script, keys, scriptArgs)` and the error log into the `onErrorResume` lambda makes both actually execute; the argument is equivalent to the old `rateLimiterAlgorithm.getScript()`.
- Wrapping the callback in try/catch is a good defensive addition — a callback failure (e.g. Redis still down during zset cleanup) can no longer break the fallback emission.
- Fail-open semantics (`allowed=true, tokensRemaining=-1`) are unchanged, so this is purely a correctness/observability fix with no behavioral regression.
- The new test `allowedThrowableInvokesAlgorithmCallbackTest` fails on the old code and passes on the new one, and its mocks match `ConcurrentRateLimiterAlgorithm#callback` (which calls `opsForZSet().remove(...).subscribe()` — the mock interaction is recorded synchronously, so `verify` is reliable).

LGTM, approving.

PR #7033, opened 2026-09-05
fix: make shenyu-examples-mcp work normally

### LINKED ISSUE #7032: [BUG] shenyu-examples-mcp can't find healthy upstream

### Current Behavior

now it will return error that can't find healthy upstream if you call the api in shenyu-examples-mcp by mcp client

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
ShenYu version(s):latest
```

### Debug logs

_No response_

### Anything else?

i will fix it

hengyuss:
Fixes: #7032 

Make sure that:

### DIFF
--- a/shenyu-examples/pom.xml
+++ b/shenyu-examples/pom.xml
@@ -39,7 +39,7 @@
         <docker-maven-plugin.version>0.40.1</docker-maven-plugin.version>
         <maven-checkstyle-plugin.version>3.4.0</maven-checkstyle-plugin.version>
         <swagger.version>2.6.0</swagger.version>
-        <shenyu.client.java.version>2.7.0.3</shenyu.client.java.version>
+        <shenyu.client.java.version>2.7.2-SNAPSHOT</shenyu.client.java.version>
     </properties>
 
     <modules>
--- a/shenyu-client/shenyu-client-core/src/main/java/org/apache/shenyu/client/core/disruptor/ShenyuClientRegisterEventPublisher.java
+++ b/shenyu-client/shenyu-client-core/src/main/java/org/apache/shenyu/client/core/disruptor/ShenyuClientRegisterEventPublisher.java
@@ -27,8 +27,6 @@ import org.apache.shenyu.disruptor.provider.DisruptorProvider;
 import org.apache.shenyu.register.client.api.ShenyuClientRegisterRepository;
 import org.apache.shenyu.register.common.type.DataTypeParent;
 
-import java.util.Objects;
-
 /**
  * The type shenyu client register event publisher.
  */
@@ -37,6 +35,8 @@ public class ShenyuClientRegisterEventPublisher {
     private static final ShenyuClientRegisterEventPublisher INSTANCE = new ShenyuClientRegisterEventPublisher();
 
     private volatile DisruptorProviderManage<DataTypeParent> providerManage;
+
+    private boolean initialized;
 
     /**
      * Get instance.
@@ -53,7 +53,7 @@ public class ShenyuClientRegisterEventPublisher {
      * @param shenyuClientRegisterRepository shenyuClientRegisterRepository
      */
     public synchronized void start(final ShenyuClientRegisterRepository shenyuClientRegisterRepository) {
-        if (Objects.nonNull(providerManage)) {
+        if (initialized) {
             return;
         }
         RegisterClientExecutorFactory factory = new RegisterClientExecutorFactory();
@@ -64,6 +64,7 @@ public class ShenyuClientRegisterEventPublisher {
         DisruptorProviderManage<DataTypeParent> newProviderManage = new DisruptorProviderManage<>(factory);
         newProviderManage.startup();
         providerManage = newProviderManage;
+        initialized = true;
     }
 
     /**
--- a/shenyu-client/shenyu-client-core/src/main/java/org/apache/shenyu/client/core/disruptor/ShenyuClientRegisterEventPublisher.java
+++ b/shenyu-client/shenyu-client-core/src/main/java/org/apache/shenyu/client/core/disruptor/ShenyuClientRegisterEventPublisher.java
@@ -27,6 +27,8 @@ import org.apache.shenyu.disruptor.provider.DisruptorProvider;
 import org.apache.shenyu.register.client.api.ShenyuClientRegisterRepository;
 import org.apache.shenyu.register.common.type.DataTypeParent;
 
+import java.util.Objects;
+
 /**
  * The type shenyu client register event publisher.
  */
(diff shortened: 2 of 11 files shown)


hengyuss:
Should I revert the version in the shenyu-client parent POM to 2.7.0.3 while allowing shenyu-examples-mcp to use 2.7.2-snapshot separately? 
If examples depend on multiple shenyu-clients, issues may arise where upstream services fail to register successfully. So I change ther the version in the shenyu-client parent POM to 2.7.2-snapshot for the fix can apply.
 @dengliming

dengliming:
@hengyuss Thanks for the explanation. I understand the motivation for aligning the client versions to avoid registration issues when multiple `shenyu-client` versions are involved.

My concern is that changing the shared `shenyu.client.java.version` to `2.7.2-SNAPSHOT` affects all examples, not just `shenyu-examples-mcp`, and introduces a dependency on an unreleased snapshot version for clean standalone builds. Since the examples README still points to the published `2.7.0.3` artifacts, this could make the repository harder to build for other contributors.

Would it be possible to keep the parent version on the released artifact and apply a more scoped override for `shenyu-examples-mcp`, or update the related build/documentation together if the whole examples module now requires the snapshot?

Aias00 (dismissed):

hengyuss:
I think update the related build/documentation is a good choice and I will do it

Aias00 (changes requested):
## Review (PMC Aias00) — `fix: make shenyu-examples-mcp work normally`

**Decision: REQUEST CHANGES**

The core diagnosis is excellent and clearly the right one — but one change in here breaks standalone builds of the examples.

### 🔴 Blocking — examples must not default to an unreleased SNAPSHOT

```xml
-  <shenyu.client.java.version>2.7.0.3</shenyu.client.java.version>
+  <shenyu.client.java.version>2.7.2-SNAPSHOT</shenyu.client.java.version>
```

`shenyu-examples` is a **standalone reactor** (`mvn -f shenyu-examples/pom.xml ...`) whose client artifacts are external, published from `apache/shenyu-client-java`. `2.7.2-SNAPSHOT` is not on Maven Central and is not built by this repository's reactor, so pinning it as the *default* means a clean checkout can no longer build any example — not just the MCP one. The README you added even says so:

> "A clean standalone build cannot assume that these unreleased artifacts are available from Maven Central."

That is exactly the property we must not ship as a default. Apache examples have to build out of the box; new contributors and release managers rely on it.

Please revert the default to `2.7.0.3` and keep `2.7.2-SNAPSHOT` as an **explicit override** only:
```bash
mvn -f shenyu-examples/pom.xml -pl shenyu-examples-mcp -am \
  -Dshenyu.client.java.version=2.7.2-SNAPSHOT -DskipTests
```
The README section you wrote is genuinely useful — keep it, just reworded as "override" rather than "default".

### 🟠 Should fix

2. **This duplicates PR #7082 and will conflict.** Both PRs make `ShenyuClientRegisterEventPublisher#start()` idempotent, with different implementations (#7082 guards on `providerManage != null`; this one adds a separate `initialized` flag). They touch the same lines of the same file. Please drop the publisher change from this PR (or close one of the two) so the idempotency fix lands once, cleanly.
3. **`shenyu-spring-boot-starter-client-mcp/.../META-INF/spring.factories` is dead configuration.** The project is on Spring Boot **3.3.1**, where `EnableAutoConfiguration` in `spring.factories` is no longer read — only `META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports` is. Harmless, but it is noise that suggests Boot 2 support we do not have. I would drop it.

### ✅ The actual fix — nice find

Adding the missing auto-configuration registration is the real reason the MCP example never worked:

```
+META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports
+org.apache.shenyu.springboot.starter.client.mcp.ShenyuMcpClientConfiguration
```

Without this file, Spring Boot 3 never loaded `ShenyuMcpClientConfiguration`, so nothing registered. That is a one-file root-cause fix, and it explains the symptom precisely. Good catch.

### Also good

4. `ShenyuToolCallback` forcing `shenyuContext.setRpcType(RpcTypeEnum.HTTP.getName())` with a clear comment is reasonable — MCP tool invocation is an internal HTTP call, so the response plugin must pick an HTTP `MessageWriter`. It is covered by `testToolCallUsesHttpRpcType`. One caution: it now overrides the metadata-driven rpcType **unconditionally**, including when `MetaData` is present and enabled. If MCP ever fronts a non-HTTP downstream, this will need revisiting — worth a `TODO` or a narrower condition.
5. `start()` publishing `providerManage` only after `startup()` succeeds is correct (avoids exposing a half-started disruptor). `initialized` is only read/written under the `synchronized` method, and `providerManage` is `volatile`, so the visibility is handled.

Revert the SNAPSHOT default and this is good to go.

shenyu-examples/pom.xml
```diff
@@ -39,7 +39,7 @@
         <docker-maven-plugin.version>0.40.1</docker-maven-plugin.version>
         <maven-checkstyle-plugin.version>3.4.0</maven-checkstyle-plugin.version>
         <swagger.version>2.6.0</swagger.version>
-        <shenyu.client.java.version>2.7.0.3</shenyu.client.java.version>
+        <shenyu.client.java.version>2.7.2-SNAPSHOT</shenyu.client.java.version>
```
Aias00, line 42: 🔴 **Blocking — this must stay a released version.**

`shenyu-examples` is a standalone reactor (`mvn -f shenyu-examples/pom.xml`) and these client artifacts come from `apache/shenyu-client-java`. `2.7.2-SNAPSHOT` is not on Maven Central and is not produced by this repo's reactor, so making it the default means a clean checkout can no longer build *any* example.

Please revert to `2.7.0.3` here and pass `2.7.2-SNAPSHOT` as an explicit `-Dshenyu.client.java.version=...` override in the MCP build docs. Apache examples have to build out of the box.

Aias00 (approved):
## Review (updating my earlier request-changes)

The real root cause I pointed at is now fixed: `shenyu-spring-boot-starter-client-mcp` was missing `META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports`, so under Boot 3.3.1 its configuration class was never loaded. That file is now added, and `ShenyuToolCallback.configureShenyuContext` correctly forces `rpcType = HTTP` so the MCP response selects an HTTP `MessageWriter` (MCP tools are invoked through an internal HTTP request) instead of an MCP writer. `ShenyuToolCallbackTest` covers the new behaviour.

Soft (non-blocking) suggestion: `shenyu.client.java.version` still defaults to `2.7.2-SNAPSHOT` in `shenyu-examples/pom.xml`. The build currently resolves it, but before the 2.7.2 cut please pin it to a released `shenyu-client-java` version so the examples reactor builds for users without snapshot-repo access. Not a merge blocker.

CI's `e2e-websocket` failure (and the cancelled siblings) is infra/flaky and unrelated to MCP. Approving with the version note above.

PR #7035, opened 2026-09-05
[ISSUE #7034] Fix transient cache gaps during websocket refresh.

### LINKED ISSUE #7034: [BUG] Websocket refresh temporarily removes cached plugin, selector and rule data

### Current Behavior

WebSocket REFRESH and MYSELF handling removes the received Plugin, Selector and Rule entries from the cache, then adds them back one by one.

Request threads read these caches concurrently. Between removal and re-subscription, valid configuration is temporarily unavailable: a plugin may be skipped, or selector/rule matching may fail.

This can happen during synchronization after a WebSocket reconnect or an explicit refresh, even when the configuration has not changed.

### Expected Behavior

Refreshing unchanged configuration should not temporarily make it unavailable to requests.

The existing cache should remain readable while the refreshed batch is being built. The completed batch should then become visible in one publication.

### Steps To Reproduce

1. Configure WebSocket data synchronization and a divide route with one matching selector and rule. Confirm that requests succeed.
2. Trigger synchronization by reconnecting the gateway to Admin.
3. For deterministic reproduction, pause SelectorDataHandler.doRefresh() immediately after refreshSelectorDataSelf(dataList) and before the subscription loop.
4. Send a request to the configured route. The selector is temporarily absent from BaseDataCache, so matching fails.
5. Resume synchronization. The selector is added back and matching succeeds again.

Plugin and Rule refresh handlers contain the same remove-then-add interval.

### Environment

```markdown
ShenYu version(s): 2.7.2-SNAPSHOT
Code baseline: master at 7273f23b7, before the proposed fix
Data synchronization: WebSocket
```

### Debug logs

_No response_

### Anything else?

The proposed fix is limited to removing the temporary cache gap. It preserves entries absent from the received batch and keeps the existing empty-batch behavior, since REFRESH can also contain single-plugin synchronization data.

TheoLi0905:
Fixes #7034

WebSocket refresh currently removes cached entries before adding the received data back. Concurrent requests can therefore observe missing configuration even when that configuration has not changed.

This change builds the refreshed batch in a copy of the existing map and publishes it through a single volatile reference assignment. Selector and Rule batches use new lists, so the published lists remain unchanged during construction.

Missing entries and empty batches retain their existing behavior. Default batch methods preserve the legacy callbacks for custom subscribers. No protocol changes are introduced.

Make sure that:

### DIFF
--- a/shenyu-plugin/shenyu-plugin-base/src/main/java/org/apache/shenyu/plugin/base/cache/BaseDataCache.java
+++ b/shenyu-plugin/shenyu-plugin-base/src/main/java/org/apache/shenyu/plugin/base/cache/BaseDataCache.java
@@ -40,17 +40,17 @@ public final class BaseDataCache {
     /**
      * pluginName -> PluginData.
      */
-    private static final ConcurrentMap<String, PluginData> PLUGIN_MAP = Maps.newConcurrentMap();
+    private static volatile ConcurrentMap<String, PluginData> pluginMap = Maps.newConcurrentMap();
 
     /**
      * pluginName -> SelectorData.
      */
-    private static final ConcurrentMap<String, List<SelectorData>> SELECTOR_MAP = Maps.newConcurrentMap();
+    private static volatile ConcurrentMap<String, List<SelectorData>> selectorMap = Maps.newConcurrentMap();
 
     /**
      * selectorId -> RuleData.
      */
-    private static final ConcurrentMap<String, List<RuleData>> RULE_MAP = Maps.newConcurrentMap();
+    private static volatile ConcurrentMap<String, List<RuleData>> ruleMap = Maps.newConcurrentMap();
 
     private BaseDataCache() {
     }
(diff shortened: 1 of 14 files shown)


dmsolr:
Request changes: the refresh implementation in BaseDataCache can lose concurrent incremental updates. refreshPluginData/refreshSelectorData/refreshRuleData copy the current map, build a replacement, and then publish it via a volatile assignment without synchronizing with cache*Data or remove*Data. If an update or deletion occurs while the replacement is being built, it writes to the old map and is then overwritten by the refresh publication, causing an update to be lost or deleted data to reappear. Please protect refresh and incremental mutations with a common lock or versioned CAS/retry, and add a concurrency test covering an update/deletion racing with refresh.

Aias00 (approved):

TheoLi0905:
Updated the PR branch using GitHub's "Update branch → Rebase" option. No intentional code changes were made.

PR #7043, opened 2026-09-08
[type:perf] kafka logging: flush producer once per batch instead of p…

renxf0319:
### What this PR do
Optimize logging‑kafka plugin performance:
- Move producer.flush() out of record loop, flush once per batch logs instead of per record.
- Avoid frequent blocking flush, restore kafka producer batch capability.
- Minor optimize: extract config map outside loop, add topic blank defensive check, adjust log level.

### Background
The original code calls producer.flush() after every record send, which disables Kafka batch mechanism and seriously degrades gateway throughput.


renxf0319:
已补充异常捕获逻辑

Aias00 (approved):

renxf0319:
Hi, could you please help trigger the CI workflow for this PR? Thanks a lot.

renxf0319:
Hi @Aias00, thanks for the earlier review.

I rebuilt the branch on top of the latest apache:master as a single commit (61c2656)
to resolve the merge conflict, which unfortunately dismissed your previous approval.
Sorry for the churn.

The change is functionally the same as before:
- producer.flush() is called once per batch instead of once per record;
- the batch-level flush is wrapped in try/catch so a KafkaException no longer escapes
  consume0() and makes AbstractLogCollector discard the whole drained batch;
- unit tests added for the once-per-batch flush and the flush failure path.

Local verification: ./mvnw -pl shenyu-plugin/shenyu-plugin-logging/shenyu-plugin-logging-kafka -am test
→ Tests run: 4, Failures: 0, Errors: 0 (checkstyle passed too).

Could a maintainer please approve the pending workflow runs and take another look? Thanks!

Aias00 (approved):
## Review: kafka logging — flush producer once per batch instead of per record

**Verdict: APPROVED**

### What this PR does
`KafkaLogCollectClient.consume0` previously called `producer.flush()` inside the per-record `catch` block, which meant the flush was only attempted when a record failed and could be skipped/duplicated across the loop. This moves `producer.flush()` **once, after the batch loop**, and wraps it in its own `try/catch` so a flush failure is logged without silently dropping the batch or masking the per-record errors.

### Tests
- `testConsume0FlushesOncePerBatch` — asserts exactly one `flush()` call per batch.
- `testConsume0HandlesFlushFailure` — asserts a flush exception is handled and doesn't abort the consume path.

Both target the actual regression (flush frequency + failure handling). Clean, low-risk reliability/perf improvement with no behaviour change for the happy path.

### Notes (non-blocking)
- `gh pr checks` currently reports "no checks" on the head branch (likely a force-push reset the runs); the Java build/unit jobs were green in the prior run. A re-run before merge is always good practice.

Nice, focused fix — thanks.

Issue #6626, opened 2026-08-03
[BUG] FileSizeFilter leaks pooled DataBuffer on oversized-file rejection path


Aias00:
- Severity: High
- Location:
`shenyu-web/src/main/java/org/apache/shenyu/web/filter/FileSizeFilter.java:76-84` (rejection path, no release) vs `:98` (`.doFinally` release only on success path)
- 
Description:
When the size check at line 78 rejects the upload (`dataBuffer.capacity() > Constants.BYTES_PER_MB * fileMaxSize`), the lambda returns `WebFluxResultUtils.result(exchange, error)` at line 84 **without releasing `dataBuffer`**. The `doFinally(signalType -> DataBufferUtils.release(dataBuffer))` is only on the success-path `bodyInsert.insert(...).then(...).doFinally(...)` chain (line 98), not the rejection branch.
- 
Impact:
Every rejected oversized multipart upload leaks one pooled `DataBuffer` (up to `fileMaxSize` MB of Netty direct memory). An attacker sending oversized multipart bodies can exhaust the pooled direct-memory arena, causing `OutOfDirectMemoryError` with no GC recovery.
- 
Suggested fix:
Add `.doFinally(signalType -> DataBufferUtils.release(dataBuffer))` to the rejection-path Mono, or restructure so both paths share a single outer `doFinally`.
- 
Confidence: High
- Related existing: #4505 (CLOSED, the original leak fix) — that fix only patched the success path and left the rejection path leaking. This is an incomplete-fix regression, not a dup.

---
_Identified during the 2026-08-02 deep re-scan; full list in [`docs/scan2-2026-08-02/00-consolidated-critical-high.md`](docs/scan2-2026-08-02/00-consolidated-critical-high.md)._

PR #4507, opened 2023-03-28
[Bugfix#4505] FileSizeFilter make memory leak

847850277:
Dear community
1、Corrected `maxInMemorySize` to `- 1` to resolve the`org.springframework.core.io.buffer.DataBufferLimitException: Exceeded limit on max bytes to buffer` error.
2、Finally, the DataBuffer object was released.

Make sure that:

- [ ] You have read the [contribution guidelines](https://shenyu.apache.org/community/contributor-guide).
- [ ] You submit test cases (unit or integration tests) that back your changes.
- [ ] Your local test passed `./mvnw clean install -Dmaven.javadoc.skip=true`.

PR #7061, opened 2026-09-14
fix: bound multipart buffering in FileSizeFilter to prevent memory DoS (#6627)

### LINKED ISSUE #6627: [BUG] FileSizeFilter maxInMemorySize(-1) buffers entire request body in memory before size check (DoS)
- Severity: Medium
- Location:
`shenyu-web/src/main/java/org/apache/shenyu/web/filter/FileSizeFilter.java:62-64` (constructor sets `-1`), `:76` (`bodyToMono(DataBuffer.class)` reads entire body), `:78` (size check fires only after full buffering)
- 
Description:
The constructor sets `maxInMemorySize(-1)` (unlimited). `serverRequest.bodyToMono(DataBuffer.class)` at line 76 reads the **entire** multipart body into a single in-memory `DataBuffer` with no size cap. The size check at line 78 only fires **after** the full body is already buffered. An attacker can send a 10 GB multipart body; the gateway buffers all 10 GB, then rejects it as "too large" — but the OOM damage is already done.
- 
Impact:
Memory-exhaustion DoS. A single malicious request can crash the gateway process.
- 
Suggested fix:
Set `maxInMemorySize` to a bounded value (e.g. `fileMaxSize * BYTES_PER_MB`) so the codec rejects oversized bodies during buffering, or switch to a streaming size-check.
- 
Confidence: High
- Related existing: #4405 / #4505 — those are about the API and the leak; the DoS aspect of `-1` is not in either.

---

wy471x:
Make sure that:

## Summary

Fixes #6627.

`FileSizeFilter` put no bound on what it buffered: the constructor built its codecs with `maxInMemorySize(-1)`, so `serverRequest.bodyToMono(DataBuffer.class)` joined the **entire** multipart body in memory and the size check only ran after that. A client could stream an arbitrarily large multipart body and have the gateway buffer all of it before the request was finally rejected — a memory-exhaustion DoS (one request can take the process down).

The bound had been removed on purpose in #4507 ("Corrected `maxInMemorySize` to `-1` to resolve the `DataBufferLimitException: Exceeded limit on max bytes to buffer` error"), because `DataBufferLimitException` was not handled and broke the existing rejection response. This PR re-introduces the bound **and** handles that exception, which is what makes it safe.

### Changes:

1. `FileSizeFilter` (`FileSizeFilter.java:65`, `:73`) — new `maxInMemorySize` field, computed as `fileMaxSize * Constants.BYTES_PER_MB` (saturating at `Integer.MAX_VALUE` to avoid `int` overflow) and passed to `configurer.defaultCodecs().maxInMemorySize(...)`, so the codec aborts while buffering as soon as the configured limit is crossed instead of buffering the whole body first.
2. `FileSizeFilter.filter` (`FileSizeFilter.java:111`) — `onErrorResume(DataBufferLimitException.class, ...)` maps the codec limit trip to the same rejection response the filter already produced (`400` + `ShenyuResultEnum.PAYLOAD_TOO_LARGE`), preserving the existing API behavior for oversized uploads.
3. `FileSizeFilter.filter` (`FileSizeFilter.java:83`) — a non-positive `shenyu.file.max-size` now rejects the multipart request before reading the body at all, so no configuration is left with unbounded buffering (previously such a request was buffered in full and only then rejected).
4. `FileSizeFilter.filter` (`FileSizeFilter.java:90`) — the rejection branch now releases the joined pooled `DataBuffer` before returning; the existing `.doFinally` release only covered the success path, so every oversized upload leaked one pooled buffer (the leak tracked in #6626).
5. `FileSizeFilter.payloadTooLarge` (`FileSizeFilter.java:125`) / `maxInMemorySize` (`FileSizeFilter.java:141`) — the three rejection sources now share one response/logging path instead of duplicating the status + error construction.

### Test Cases:

- `FileSizeFilterTest#testFilterRejectsOversizedBodyWhileBuffering` (`FileSizeFilterTest.java:128`) — sends 4 × 512 KB against a 1 MB limit and asserts the request is rejected with `400` while the body is *not* buffered in full (only the buffers consumed until the limit trips) and the filter chain is never called. With the previous `maxInMemorySize(-1)` the same test fails with `consumed buffers: 4`, i.e. it pins the vulnerability.
- Existing `FileSizeFilterTest#testFilter` / `#testDecorate` are unchanged and still pass, including the `fileMaxSize = -1` rejection case.

## Verification

- `./mvnw clean install -Dmaven.javadoc.skip=true` on JDK 21: whole reactor passes, with the one pre-existing order-dependent test excluded (`-Dtest='!DubboReconcilerTest' -DfailIfNoTests=false`). `org.apache.shenyu.k8s.DubboReconcilerTest` shares the static `IngressCache` with `WebSocketReconcilerTest` / `DivideIngressReconcilerTest` (all use `mockedNamespace/mockedIngress`) and fails purely on test execution order — it passes in isolation and reproduces the same failure on unmodified `master`; `shenyu-kubernetes-controller` is unrelated to this change.
- `shenyu-web` module: 82 tests, 0 failures.
- `checkstyle:check` — 0 violations; `mvn validate` (checkstyle + Apache RAT) passes.

Note: the release in change 4 also fixes the pooled-buffer leak tracked in #6626, since it is the same rejection path.

@Aias00, could you please help review this PR? Thank you!

close #6627


Aias00 (approved):

PR #7062, opened 2026-09-14
fix: make upstream lists thread-safe (#6857)

### LINKED ISSUE #6857: [BUG] UpstreamCacheManager.findUpstreamListBySelectorId returns live ArrayList mutated by health-check thread without reader sync
## Description
`findUpstreamListBySelectorId` returns `task.getHealthyUpstream().get(selectorId)` — a direct reference to the live `ArrayList` stored in `UpstreamCheckTask.healthyUpstream` (a `ConcurrentMap` of plain `ArrayList` values). The health-check thread modifies this same list under `synchronized(lock)`: `putToMap` calls `list.add(upstream)` and `removeFromMap` calls `list.remove(upstream)`. However, the readers (`DividePlugin:94`, `WebSocketPlugin:92`, `ApacheDubboGrayLoadBalance:57`, `DefaultRetryStrategy:109`) iterate and index the returned list on request threads WITHOUT acquiring `lock`. `ArrayList` is not thread-safe: a concurrent `add` (which may trigger `Arrays.copyOf` resize) or `remove` (which shifts elements via `System.arraycopy`) while a reader calls `size()` then `get(i)` can produce `IndexOutOfBoundsException`, null reads, or stale/inconsistent list state.

## Location
- `shenyu-loadbalancer/src/main/java/org/apache/shenyu/loadbalancer/cache/UpstreamCacheManager.java:128`
- `shenyu-loadbalancer/src/main/java/org/apache/shenyu/loadbalancer/cache/UpstreamCheckTask.java:289-292,305-308,355`
- `shenyu-plugin/shenyu-plugin-proxy/shenyu-plugin-divide/src/main/java/org/apache/shenyu/plugin/divide/DividePlugin.java:94`

## Impact
Occasional `IndexOutOfBoundsException` or NPE in the load balancer during a health-check cycle, causing intermittent 500 errors for proxied requests. Also possible: load balancer sees a partially-updated list (missing or phantom upstreams), routing to a stale or non-existent endpoint.

## Suggested fix
Either (a) return a snapshot copy from `findUpstreamListBySelectorId` (`new ArrayList<>(list)` under `synchronized(lock)`), or (b) change the `healthyUpstream` values to `CopyOnWriteArrayList` so readers get a consistent snapshot. Option (b) is lower overhead for the read-heavy path.

## Related existing
None — distinct from #6570 (BaseDataCache.removeSelectData/removeRuleData mutate live list in place) which covers `shenyu-common`'s selector/rule cache, and distinct from N35 (admin-side `UpstreamCheckService.fetchUpstreamData` LinkedList in `shenyu-admin`). This covers the gateway-side `UpstreamCacheManager`/`UpstreamCheckTask` in `shenyu-loadbalancer`.

juicewcode:
Fixes #6857

## Summary

  - Replace mutable upstream `ArrayList` instances with `CopyOnWriteArrayList`.
  - Prevent request threads from observing unsafe concurrent modifications made by health-check threads.
  - Preserve the existing synchronization for compound upstream update operations.

## Test

  - Add a regression test verifying that upstream lists use `CopyOnWriteArrayList`.
  - Verify that an existing iterator retains a stable snapshot when an upstream is added.
  - Verify that subsequent reads observe the newly added upstream.

Make sure that:


Aias00 (approved):

PR #7070, opened 2026-09-16
fix(sdk): avoid sharing client proxies across contexts

### LINKED ISSUE #6856: [BUG] ShenyuClientProxyFactory caches proxies in static ConcurrentMap keyed by apiClass — stale proxy + ApplicationContext leak across restarts
## Description
`PROXY_CACHE` is a `static` `ConcurrentMap<Class<?>, Object>` that is never cleared. `createProxy` does a `containsKey`-then-`get` fast path, else `synchronized(apiClass)` builds a `ShenyuClientInvocationHandler` (which captures the `ApplicationContext` and the per-bean `ShenyuClientFactoryBean` and resolves `applicationContext.getBean(ShenyuSdkClient.class)` at construction) and stores it in the static map. The first `ApplicationContext` to build a proxy for a given interface wins; every subsequent context (Spring Boot devtools restart, repeated `@SpringBootTest` contexts) receives the cached proxy bound to the now-closed previous context.

## Location
- `shenyu-sdk/shenyu-sdk-spring/src/main/java/org/apache/shenyu/sdk/spring/proxy/ShenyuClientProxyFactory.java:36,51-61`
- `shenyu-sdk/shenyu-sdk-spring/src/main/java/org/apache/shenyu/sdk/spring/proxy/ShenyuClientInvocationHandler.java:65-74,99-114`

## Impact
After a context restart the proxy's captured `ShenyuSdkClient` and `ApplicationContext` belong to the dead context (broken/leaked HTTP client, stale url/path from the old `ShenyuClientFactoryBean`). The static cache also holds a strong reference to the closed `ApplicationContext` and all beans it transitively captured, preventing GC of closed contexts → memory growth across restarts.

## Suggested fix
Key the cache per-context (or drop the static cache and rely on `FactoryBean.isSingleton()` + Spring's own singleton cache); alternatively clear `PROXY_CACHE` on context close / make it an instance field wired to the registrar.

## Related existing
None — distinct from #6787 (static URIS list in register-client `ShenyuClientURIExecutorSubscriber`, different class/module) and #6729 (GW-SPI-7, ExtensionLoader pinned ClassLoader).

dengliming:
## Summary
- remove the static API-class proxy cache from `ShenyuClientProxyFactory`
- let Spring `FactoryBean` singleton scoping manage proxies per application context
- add a regression test proving separate contexts receive separate proxies

## Testing
- `mvn -pl shenyu-sdk/shenyu-sdk-spring -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=ShenyuClientProxyFactoryTest -DfailIfNoTests=false test`

Fixes #6856


Aias00 (approved):

PR #7081, opened 2026-09-16
fix(admin): persist app auth apply updates

### LINKED ISSUE #6540: [BUG] AppAuthServiceImpl.applyUpdate never persists open/enabled/phone/extInfo/userId and gates path-saving on the stale DB open
## Description
`applyUpdate` finds `appAuthDO` by appKey, then conditionally saves `authParam` and `authPath`, and publishes a `CREATE` event built from the *unchanged* entity. It never calls any mapper update, so the DTO's `open`/`enabled`/`phone`/`extInfo`/`userId` are dropped. Worse, the path-save guard uses the **DB** value `appAuthDO.getOpen()` (line 161), while validation (`hasMissingPathsWhenOpen`) uses the **DTO** value `authApplyDTO.getOpen()`. If the DB row has `open=false` and the client applies with `open=true` + paths, validation passes but paths are not saved, and the event carries `open=false`. The `applyCreate` path (built via `AppAuthDO.create(AuthApplyDTO)`) persists all those fields, confirming the asymmetry is a bug.

## Location
```
shenyu-admin/.../service/impl/AppAuthServiceImpl.java:145-178
applyCreate path builds DO via AppAuthDO.create
```

## Impact
Enabling path-level auth via the apply-update flow silently fails; published sync data is stale.

## Suggested fix
Persist the changed fields (`appAuthMapper.updateSelective(...)` built from the DTO), and use `authApplyDTO.getOpen()` for the path-save branch.

## Related existing issue(s)
#6279 (closed) was about the open/enable toggle UI symptom; this is the underlying persistence gap in `applyUpdate`, distinct.

dengliming:
## Summary
- persist mutable application authorization fields during apply updates
- decide path replacement from the requested open state instead of stale database state
- publish synchronization data from the updated entity
- add regression coverage for enabling path authorization on an existing closed application

## Testing
- `JAVA_TOOL_OPTIONS=-javaagent:<byte-buddy-agent> mvn -pl shenyu-admin -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=AppAuthServiceTest -DfailIfNoTests=false test`

Fixes #6540


Aias00 (approved):

PR #7082, opened 2026-09-16
fix(client): make register publisher startup idempotent

### LINKED ISSUE #6541: [BUG] ShenyuClientRegisterEventPublisher.start() is not idempotent: leaked Disruptor threads + duplicate heartbeat schedulers
## Description
`ShenyuClientRegisterEventPublisher` is a singleton, but `start()` is invoked once per client listener ctor (+ once for the `publisher` bean + a Tars duplicate). Each call constructs a brand-new `DisruptorProviderManage` with its own consumer thread and its own `ShenyuClientURIExecutorSubscriber`. The subscriber's `URIS` list is `static` (shared), while each subscriber's heartbeat `executor` is an instance field scheduled in its constructor. With N `start()` calls there are N heartbeat schedulers all iterating the same shared `URIS` list.

## Location
```
shenyu-client-core/.../AbstractContextRefreshedEventListener.java:139
shenyu-client-tars/.../TarsServiceBeanEventListener.java:80
shenyu-client-autoconfig/.../ClientRegisterConfiguration.java:99-104
ShenyuClientRegisterEventPublisher.java:53-61
ShenyuClientURIExecutorSubscriber.java:50 (URIS is static)
```

## Impact
Duplicate heartbeat traffic to the admin for every registered URI (2x for Tars alone), and orphaned Disruptor consumer threads for the process lifetime.

## Suggested fix
Make `start()` idempotent (guard with an `AtomicBoolean`/`started` flag), and remove the redundant `publisher.start(...)` call in `TarsServiceBeanEventListener` line 80.

## Related existing issue(s)
#6487 covers heartbeat scheduler *stop* after one full failure; this is about *duplicate* schedulers from non-idempotent start, distinct.

dengliming:
## Summary
- serialize register publisher initialization and reuse the first started provider manager
- publish the provider manager only after startup completes
- remove the redundant Tars listener startup call
- add regression coverage for repeated startup

## Testing
- `JAVA_TOOL_OPTIONS=-javaagent:<byte-buddy-agent> mvn -pl shenyu-client/shenyu-client-core,shenyu-client/shenyu-client-tars -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=ShenyuClientRegisterEventPublisherTest -DfailIfNoTests=false test`

Fixes #6541


Aias00 (changes requested):

shenyu-client/shenyu-client-core/src/main/java/org/apache/shenyu/client/core/disruptor/ShenyuClientRegisterEventPublisher.java
```diff
@@ -50,14 +52,49 @@ public static ShenyuClientRegisterEventPublisher getInstance() {
...
+        }
+    }
+
+    /**
+     * Create URI subscriber.
+     *
+     * @param repository register repository
+     * @return URI subscriber
+     */
+    protected ShenyuClientURIExecutorSubscriber createUriSubscriber(final ShenyuClientRegisterRepository repository) {
```
Aias00, line 86: Non-blocking design nit: these two `protected` factories exist only so `ShenyuClientRegisterEventPublisherTest` can inject mocks. They widen the API surface of a public singleton (any subclass anywhere can now override disruptor construction).

Options if you want to keep production API tight:
- make them package-private and put the test in the same package (`org.apache.shenyu.client.core.disruptor`), or
- drop them and use `Mockito.mockConstruction(DisruptorProviderManage.class, ...)` in the test.

Not a blocker — the behaviour is correct either way.

shenyu-client/shenyu-client-core/src/main/java/org/apache/shenyu/client/core/disruptor/ShenyuClientRegisterEventPublisher.java
```diff
@@ -50,14 +52,49 @@ public static ShenyuClientRegisterEventPublisher getInstance() {
...
+        factory.addSubscribers(uriSubscriber);
         factory.addSubscribers(new ShenyuClientApiDocExecutorSubscriber(shenyuClientRegisterRepository));
         factory.addSubscribers(new ShenyuClientMcpExecutorSubscriber(shenyuClientRegisterRepository));
-        providerManage = new DisruptorProviderManage<>(factory);
-        providerManage.startup();
+        DisruptorProviderManage<DataTypeParent> manage = createProviderManage(factory);
+        try {
+            manage.startup();
+            uriSubscriber.start();
+            providerManage = manage;
```
Aias00, line 69: Good — assigning `providerManage` only after both `startup()` and `uriSubscriber.start()` succeed is what makes the failure path retryable (the field stays null, so the next `start()` builds a fresh manager).

One thing worth a line of Javadoc: once this field is set, a later `start()` with a *different* `ShenyuClientRegisterRepository` is silently ignored. That is the intended idempotency, but it is surprising to a reader who passes a repository and sees it discarded.

Aias00 (approved):
## Review (re-review, PMC Aias00) — `fix(client): make register publisher startup idempotent`

**Decision: APPROVE**

### Correction to my earlier review
My previous `REQUEST_CHANGES` said "this PR does not compile". That was wrong, and I apologise for the noise: I read the red CI at face value. Pulling the actual job log for `build (21, ubuntu-latest)` shows it ends with `##[error]The operation was canceled.` — every `build`/`build-docker-images`/`k8s-examples-http` job in that run was *cancelled*, not failed, and no module reported a compilation or test failure (`check-license-header` and `changes` in the same run are green). I have re-triggered the cancelled jobs so we get a clean signal. The code is fine.

### Why the change is correct
- **The redundant Tars startup is genuinely redundant.** `TarsServiceBeanEventListener extends AbstractContextRefreshedEventListener`, and `AbstractContextRefreshedEventListener`'s constructor already calls `publisher.start(shenyuClientRegisterRepository)` (line 139). The extra call in the Tars subclass constructor invoked `start()` a second time on the same singleton, which built a **second** `DisruptorProviderManage` and overwrote `providerManage` — leaking the first ring buffer and its `shenyu_disruptor_consumer_*` / `shenyu_disruptor_provider_*` threads. That is exactly #6541. Removing it is right.
- **Idempotency.** `start()` is now `synchronized` and returns early when `providerManage != null`, so any number of client listeners (tars + http + dubbo in one app, etc.) share one disruptor. `providerManage` is `volatile`, so `publishEvent()`/`getProviderManage()` readers see a safely published, fully started instance.
- **Publish-after-startup + rollback.** `providerManage` is assigned only *after* `manage.startup()` and `uriSubscriber.start()` succeed. On a `RuntimeException` the URI subscriber and the `DisruptorProvider` are shut down and the exception is rethrown, leaving `providerManage` null so a retry can actually succeed. This closes the "failed startup leaks heartbeat schedulers" gap.
- **Heartbeat moved out of the constructor.** Scheduling the 30s/10s heartbeat from `ShenyuClientURIExecutorSubscriber`'s constructor meant a scheduler thread was created even when startup subsequently failed. Moving it to `start()` behind an `AtomicBoolean` guard fixes that, with no behaviour change on the happy path. The existing `ShenyuClientURIExecutorSubscriberTest` constructs the subscriber directly and only exercises `getType()`/`executor(...)`, so it is unaffected.
- `executor()` now calls `shutdown()` instead of `if (!executor.isTerminated()) { executor.shutdown(); }` — `ScheduledThreadPoolExecutor#shutdown()` is idempotent, so this is equivalent.

### Non-blocking suggestions
1. **`createUriSubscriber(...)` / `createProviderManage(...)` are test-only seams in production code.** They widen the API of a public singleton purely so the test can inject mocks. Consider keeping them package-private with a `@VisibleForTesting`-style comment, or use Mockito's `mockConstruction(...)`, so the extension point does not become part of the supported API surface.
2. **Document the silent no-op.** After the first successful `start()`, a later `start()` with a *different* `ShenyuClientRegisterRepository` is silently ignored. That is the intended idempotency, but a line of Javadoc on `start()` would save the next reader the same question.
3. `start()` catches `RuntimeException` only; an `Error` from `startup()` would leave the disruptor running with `providerManage` null. Practically never happens — mentioning it only for completeness.

### CI status
I re-triggered the cancelled jobs; they are still sitting in `queued` (Apache runner pool is saturated), so the status may stay yellow/red for a while through no fault of this PR. From the log evidence and the symbol checks above I am satisfied the code compiles and the tests are sound, so I am approving — but please make sure a clean run lands before merge.

Thanks for the fix, and sorry again for the earlier mis-diagnosis.

PR #7083, opened 2026-09-16
fix(httpclient): apply deduplicated WebClient headers

### LINKED ISSUE #6543: [BUG] WebClient request-header deduplication is a no-op (feature broken on the WebClient path)
## Description
`doRequest` builds a deduplicated copy of the request headers (`httpHeaders`, lines 62-63) via `duplicateHeaders(..., REQ_UNIQUE_HEADER)`. But the outbound request is then populated from `exchange.getRequest().getHeaders()` — the **original, non-deduplicated** headers — via `headers.addAll(exchange.getRequest().getHeaders())` (line 67). The `httpHeaders` variable is computed and thrown away; it is never used to build the request. The sibling `NettyHttpClientPlugin.java:72` correctly applies the deduped copy via `httpHeaders.forEach(headers::set)`.

## Location
```
shenyu-plugin-httpclient/.../WebClientPlugin.java:62-67
NettyHttpClientPlugin.java:72 (correct)
```

## Impact
The "unique header" feature is silently inactive on the WebClient path while it works on the Netty path — inconsistent behavior depending on which HTTP client is configured. Upstream servers that reject duplicate headers (or pick the wrong value) will misbehave, and the configured retain-first/last/unique policy is ignored.

## Suggested fix
In the `WebClientPlugin` headers lambda, replace `headers.addAll(exchange.getRequest().getHeaders())` with `httpHeaders.forEach(headers::set)` (mirroring `NettyHttpClientPlugin`).

## Related existing issue(s)
#6255/#6254 (closed) were about request-template custom headers; this is the dedup result being discarded entirely, distinct.

dengliming:
## Summary
- copy inbound WebClient headers into a mutable collection before deduplication
- apply the deduplicated header collection to the outbound request
- add regression coverage for the retain-last request-header strategy

## Testing
- `JAVA_TOOL_OPTIONS=-javaagent:<byte-buddy-agent> mvn -pl shenyu-plugin/shenyu-plugin-httpclient -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=WebClientPluginTest -DfailIfNoTests=false test`

Fixes #6543


Aias00 (approved):

PR #7084, opened 2026-09-16
fix(register): clean up exhausted failback retries

### LINKED ISSUE #6545: [BUG] Failback retry exhaustion leaves a permanently stuck entry that silently swallows all future failed registrations for the same key
## Description
`FailureRegistryTask` is constructed with `retryCount = 18`, so `retryLimit = (18 < 0) = false`. After 18 consecutive failures, `tickCount` reaches 19; the guard `!retryLimit && tickCount > retryCount` is true, so `run()` logs "task over" and returns **without** calling `remove(key)` and **without** rescheduling. The `Holder` for that key stays in `concurrentHashMap` forever. `addToFail` then short-circuits on the stuck entry: `if (Objects.nonNull(oldObj)) { return; }` — no new `FailureRegistryTask` is ever scheduled again for that key.

## Location
```
shenyu-common/.../timer/AbstractRetryTask.java:107-110 (exhaustion branch — returns without cleanup)
shenyu-register-client-api/.../retry/FailureRegistryTask.java:54-58 (doRetry — only removes on success)
shenyu-register-client-api/.../FailbackRegistryRepository.java:175-184 (addToFail — early-returns when key present)
```

## Impact
After ~3 minutes (18 x 10s) of sustained failure the admin becomes reachable again, but the client never re-registers that URI/metadata because every subsequent failure for the same key hits the stuck entry. Also a memory leak (`concurrentHashMap` grows with dead entries).

## Suggested fix
In the `AbstractRetryTask.run()` exhaustion branch, call a cleanup hook (e.g. `onGiveUp(key)` overridden by `FailureRegistryTask` to call `registerRepository.remove(key)`) before returning, so future failures can schedule a fresh task. Alternatively, have `doRetry` remove the entry in a `finally` and re-add on next failure.

## Related existing issue(s)
None

dengliming:
## Summary
- add a retry-exhaustion hook to the common retry task
- remove exhausted failure registrations from the failback repository
- allow later failures for the same key to schedule a fresh retry task
- add regression coverage for cleanup after all 18 attempts fail

## Testing
- `JAVA_TOOL_OPTIONS=-javaagent:<byte-buddy-agent> mvn -pl shenyu-register-center/shenyu-register-client/shenyu-register-client-api -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=FailureRegistryTaskTest -DfailIfNoTests=false test`

Fixes #6545


Aias00 (approved):

PR #7087, opened 2026-09-16
fix(response): avoid resubscribing response bodies

### LINKED ISSUE #6724: [BUG] releaseIfNotConsumed re-subscribes to a single-use response-body source, so error/cancel release leaks in-flight ByteBufs
- Severity: Medium
- Location:
`shenyu-plugin/shenyu-plugin-response/src/main/java/org/apache/shenyu/plugin/response/strategy/WebClientMessageWriter.java:84` + `:124-126`; `shenyu-plugin/shenyu-plugin-response/src/main/java/org/apache/shenyu/plugin/response/strategy/NettyClientMessageWriter.java:81` + `:97-99`
- 
Description:
Both writers, on `response.writeWith(body)` error, call `releaseIfNotConsumed(body, error)` whose body is `dataBufferDody.map(DataBufferUtils::release).then(Mono.error(ex))`. That operator **re-subscribes** to the same `body` `Flux<DataBuffer>` to drain/release remaining buffers. But both body sources are single-subscription: WebClient's `fluxResponseEntity.getBody()` and Netty's `connection.inbound().receive().retain()`. A second subscription errors or emits nothing, so the buffers already emitted-but-not-yet-released when the write failed are never reached by `DataBufferUtils::release`. They leak. On the Netty path the second-subscription `IllegalStateException` can replace the original error signal.
- 
Impact:
Pooled direct-memory ByteBuf leak on the response error/cancel path (client abort during streamed response). Compounds PERF-27/28.
- 
Suggested fix:
Track emitted buffers via `DataBufferUtils.track(...)` or release-on-complete/cancel inside the same single subscription (`body.doOnDiscard(DataBuffer.class, DataBufferUtils::release)` + `doFinally`), instead of re-subscribing to the single-use source.
- 
Confidence: Medium
- Related existing: #6413 is the *request*-body single-use retry bug (different stream and trigger); this is the *response*-body release-on-error re-subscription.

---

## D. Plugin lifecycle / SPI (8 findings)

---

dengliming:
Fixes #6724

Response writers now attach discard cleanup to the original body stream instead of subscribing to single-use response bodies a second time after an error. The Netty connection is also disposed when response writing fails.

The regression test verifies that an erroring WebClient response body is subscribed exactly once.

Test:
- `mvn -pl shenyu-plugin/shenyu-plugin-response -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=WebClientMessageWriterTest,NettyClientMessageWriterTest -DfailIfNoTests=false test`


Aias00 (approved):

PR #7088, opened 2026-09-16
fix(web): reload replaced extension jars

### LINKED ISSUE #6725: [BUG] Replaced ext-lib JAR is never reloaded (version check is dead code)
- Severity: Medium-High
- Location:
`shenyu-web/src/main/java/org/apache/shenyu/web/loader/ShenyuExtPathPluginJarLoader.java:54-56` (`if (pluginJarName.contains(absolutePath)) { continue; }`); `shenyu-web/src/main/java/org/apache/shenyu/web/loader/ShenyuPluginClassLoader.java:212-219` (`compareVersion` — zero call sites)
- 
Description:
`loadExtendPlugins` keys only on the file's absolute path. If the path is already known, the jar is silently skipped — content/version is never re-examined. `ShenyuPluginClassLoader.compareVersion(...)` would detect a version change but has zero call sites.
- 
Impact:
Operator replaces `ext-lib/foo.jar` with a new build (same filename) and expects hot reload. The scheduler rescans, sees the path known, silently ignores. Gateway keeps running the old plugin indefinitely. Only delete-then-re-add triggers reload.
- 
Suggested fix:
Track `(path, version)` or `(path, lastModified/hash)`; when path exists but version differs, `removePluginClassLoader(path)` and re-parse. Wire `compareVersion` into this check.
- 
Confidence: High
- Related existing: none

---

dengliming:
Fixes #6725

The extension path scanner now parses jars already known by path and compares their versions with the active plugin class loader. When a replacement has a different version, the old loader is closed and the replacement is returned for loading.

The regression test replaces a jar at the same path and verifies that the new version is discovered while an unchanged version is skipped.

Test:
- `mvn -pl shenyu-web -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=ShenyuExtPathPluginJarLoaderTest,ShenyuPluginClassLoaderHolderTest -DfailIfNoTests=false test`


Aias00 (approved):

PR #7091, opened 2026-09-16
fix(web): load plugin inner classes

### LINKED ISSUE #6728: [BUG] Plugin inner classes ($ entries) cannot be resolved by the plugin classloader
- Severity: Medium
- Location:
`shenyu-web/src/main/java/org/apache/shenyu/web/loader/PluginJarParser.java:66` (`!entryName.contains("$")`); `shenyu-web/src/main/java/org/apache/shenyu/web/loader/ShenyuPluginClassLoader.java:118-139` (`findClass`)
- 
Description:
`parseJar` excludes any entry containing `$` from `clazzMap` and stores it in `resourceMap`. `findClass` only consults `classCache` and `clazzMap`; it never reads `resourceMap`. So `Outer$Inner` is unfindable: `ability("Outer$Inner")` returns true → delegates to `getParent().loadClass(...)` → app classloader cannot see plugin classes → `ClassNotFoundException`.
- 
Impact:
Any uploaded/ext plugin referencing a nested/inner class (Builder pattern, nested enums, synthetic accessors) throws `NoClassDefFoundError` on first use at runtime; the plugin effectively can't use inner classes.
- 
Suggested fix:
Put `$`-bearing `.class` entries into `clazzMap` too, or have `findClass` fall back to `resourceMap` for names containing `$`.
- 
Confidence: Medium
- Related existing: none

---

dengliming:
Fixes #6728

PluginJarParser now stores every class entry in the plugin class map, including nested and synthetic classes whose names contain `$`. Those classes are therefore visible to ShenyuPluginClassLoader when an outer plugin class references them.

The regression test verifies that both outer and inner class entries are parsed as classes rather than resources.

Test:
- `mvn -pl shenyu-web -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=PluginJarParserTest -DfailIfNoTests=false test`


Aias00 (approved):

PR #7095, opened 2026-09-16
fix(sync): isolate websocket handler instances

### LINKED ISSUE #6732: [BUG] WebsocketDataHandler stores handlers in a static mutable EnumMap mutated by every constructor
- Severity: Low-Medium
- Location:
`shenyu-sync-data-center/shenyu-sync-data-websocket/src/main/java/org/apache/shenyu/plugin/sync/data/websocket/handler/WebsocketDataHandler.java:35` (`private static final EnumMap<ConfigGroupEnum, DataHandler> ENUM_MAP`), populated in constructor `:50-57`, read at `:68`
- 
Description:
`ENUM_MAP` is `static`, shared across all `WebsocketDataHandler` instances. `WebsocketSyncDataService` creates one `ShenyuWebsocketClient` (and thus one `WebsocketDataHandler`) per URL in `shenyu.sync.websocket.urls` (multiple for HA); `masterCheck`/recreate creates more. Each constructor overwrites the shared map; the last-constructed instance's handlers win for *all* clients.
- 
Impact:
In standard single-subscriber deployment the injected subscribers are the same shared Spring beans, so the clobbering is masked. But it is fundamentally wrong (mutable static state initialized in a constructor) and becomes a live cross-talk defect for any multi-tenant or test wiring that supplies per-client subscriber lists.
- 
Suggested fix:
Make `ENUM_MAP` a non-static instance field.
- 
Confidence: High (as a defect), Low (production impact in standard config)
- Related existing: none

---

## F. Header/cookie/body manipulation plugins (4 findings)

---

dengliming:
Fixes #6732

WebsocketDataHandler now owns an instance-level EnumMap instead of mutating one static map from every constructor. Multiple WebSocket clients therefore retain the subscribers they were created with and cannot overwrite each other.

The regression test constructs two handlers with different plugin subscribers and verifies that invoking the first handler does not dispatch to the second subscriber.

Test:
- `mvn -pl shenyu-sync-data-center/shenyu-sync-data-websocket -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=WebsocketDataHandlerTest -DfailIfNoTests=false test`


Aias00 (approved):

PR #7099, opened 2026-09-16
fix(tars): avoid busy spin during proxy initialization

### LINKED ISSUE #6737: [BUG] initPrx busy-spins on lock contention burning CPU
- severity: Medium
- files: `shenyu-plugin/shenyu-plugin-proxy/shenyu-plugin-rpc/shenyu-plugin-tars/src/main/java/org/apache/shenyu/plugin/tars/cache/ApplicationConfigCache.java:189-206`
- description: `initPrx` is a `while(true)` loop that calls `tryLockedLoadMetaData(...)` whenever `prxClassCache.get(path)` is null. `tryLockedLoadMetaData` uses `LOCK.tryLock()`; if the lock is held, it returns immediately, leaving `prxClass` null, so the loop re-enters with no `Thread.yield()`/backoff.
- impact: Under concurrent tars metadata registration, worker threads spin-wait (100% CPU) for the duration of ByteBuddy class generation.
- suggested_fix: Acquire the lock blockingly inside the loop, or back off with `Thread.yield()` when `tryLock()` fails.
- confidence: Medium
- related_existing: none

---

dengliming:
Replace the `tryLock` retry loop used by Tars proxy initialization with blocking lock acquisition and a cache recheck under the lock. Contending initialization threads now park instead of consuming CPU, while the double check prevents duplicate proxy class generation.

A concurrency regression test verifies that a contending initializer queues on the lock and completes after it is released.

Test: `mvn -pl shenyu-plugin/shenyu-plugin-proxy/shenyu-plugin-rpc/shenyu-plugin-tars -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=ApplicationConfigCacheTest -DfailIfNoTests=false test` (with the module test skip temporarily disabled; 6 tests passed)

Fixes #6737


Aias00 (approved):

PR #7102, opened 2026-09-17
fix(sync): process zookeeper delete events

### LINKED ISSUE #6611: [BUG] Zookeeper callback early-returns on all NODE_DELETED events — every config deletion is silently dropped
- Severity: Critical
- Location:
`shenyu-sync-data-center/shenyu-sync-data-zookeeper/src/main/java/org/apache/shenyu/sync/data/zookeeper/ZookeeperSyncDataService.java:86` (guard), `:102-112` (dead `NODE_DELETED` switch branch)
- 
Description:
The `CuratorCacheListener` lambda (`(type, oldData, data)`) has an early guard `if (Objects.isNull(data) || Objects.isNull(data.getData())) { return; }`. Per the Curator 5.7.0 `CuratorCacheListener` contract (verified: the lambda maps to `event(Type, ChildData oldData, ChildData data)` where `data` is "the new data or null"), for `NODE_DELETED` the `data` parameter is `null` because the node no longer exists. Every delete event hits the early return and never reaches the `switch(type)` block. The `case NODE_DELETED: eventType = EventType.DELETE;` branch is dead code. No plugin/selector/rule/auth/metadata/proxy-selector deletion is ever processed on the ZK sync path.
- 
Impact:
When any config item is deleted in admin (ZK sync backend), the gateway never removes it from its local cache. Deleted plugins, selectors, rules, app-auths, metadata, proxy-selectors remain active indefinitely — stale routing, phantom rules, security exposure (deleted auth keys still honored).
- 
Suggested fix:
Before the null-data guard, check `type == NODE_DELETED` and use `oldData` (carries the path and last-known data) to dispatch `event(configNamespace, oldData.getPath(), null, registerPath, EventType.DELETE)`. Only guard on `data` for PUT events.
- 
Confidence: High
- Related existing: none — #6526/#6525 are in `shenyu-registry` (discovery registry), not `shenyu-sync-data-zookeeper` (config sync). Different component.

---

dengliming:
## Summary
- read the last-known ChildData for Curator NODE_DELETED callbacks
- dispatch delete events without requiring new node data
- keep create and change callbacks guarded against missing payloads
- add a listener-level regression test for plugin deletion

## Testing
- `mvn -pl shenyu-sync-data-center/shenyu-sync-data-zookeeper -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=ZookeeperSyncDataServiceTest -DfailIfNoTests=false test`

Fixes #6611


Aias00 (approved):

yuluo-yx (approved):

yuluo-yx:
only record: 

image

PR #7104, opened 2026-09-17
fix(logging): retry index creation after check failure

### LINKED ISSUE #6547: [BUG] Elasticsearch existsIndex returns true on error, skipping index creation and dropping batches
## Description
In the `catch` of `existsIndex` the method returns `true`. `consume0` then treats the index as existing and skips `createIndex`/alias setup. On any transient error checking index existence (network blip, auth 401, ES temporarily unreachable) while the daily index does not yet exist, the day's index is never created; the subsequent `client.bulk(...)` then fails (if `auto_create_index` is disabled, common in secured clusters), so the **entire batch is dropped** on every failed check until the check happens to succeed.

## Location
```
shenyu-plugin-logging-elasticsearch/.../client/ElasticSearchLogCollectClient.java:134-142 (used at consume0 107-110)
```

## Impact
All log batches throw on insert; `AbstractLogCollector` catches and drops them, so no logs are persisted until the check happens to succeed.

## Suggested fix
Return `false` on error (fail-safe: attempt creation) or rethrow so the batch is retried; do not claim the index exists when the check itself failed.

## Related existing issue(s)
None

dengliming:
## Summary
- treat an Elasticsearch index-existence check failure as not confirmed to exist
- allow the existing caller to attempt index and alias creation instead of silently skipping it
- add regression coverage for an IOException from the indices client

## Testing
- `mvn -pl shenyu-plugin/shenyu-plugin-logging/shenyu-plugin-logging-elasticsearch -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=ElasticSearchLogCollectClientTest -DfailIfNoTests=false test`

Fixes #6547


Aias00 (approved):

PR #7110, opened 2026-09-18
fix(plugin): release response buffer once

### LINKED ISSUE #6722: [BUG] ResponseDecorator double-releases the joined DataBuffer (IllegalReferenceCountException / pooled-buffer corruption)
- Severity: Medium
- Location:
`shenyu-plugin/shenyu-plugin-base/src/main/java/org/apache/shenyu/plugin/base/support/ResponseDecorator.java:55` (first release) and `:58` (second release in `doFinally`)
- 
Description:
`writeWith` does `DataBufferUtils.join(body)` into a single buffer, copies bytes via `dataBuffer.read(bytes)`, then calls `DataBufferUtils.release(dataBuffer)` at line 55 (refcount → 0, buffer returned to pool). The returned Mono then attaches `.doFinally(signalType -> DataBufferUtils.release(dataBuffer))` at line 58, which fires on every terminal signal and calls `release` again on the same buffer. On the default Netty pooled-direct-buffer runtime, the second release either throws `IllegalReferenceCountException` or releases a buffer the pool has already handed to another writer (refcount bumped back to 1 on reuse → release drops to 0 while in use), corrupting an unrelated response. Triggered on every response through `ResponseDecorator` (wired in `ServerWebExchangeUtils.java:76`).
- 
Impact:
Post-write errors / connection resets / pooled-buffer corruption under load. On heap buffer factories the second release is a harmless no-op (why it survived).
- 
Suggested fix:
Drop the eager `release` at line 55 and keep only the `doFinally` release; or vice-versa.
- 
Confidence: High
- Related existing: none — #6444 is a different class.

---

dengliming:
## Summary
- remove the second release of the joined response DataBuffer
- keep releasing the source buffer immediately after copying its bytes
- add a regression test that verifies pooled buffers are released exactly once

## Testing
- `mvn -q -pl shenyu-plugin/shenyu-plugin-base -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=ResponseDecoratorTest -DfailIfNoTests=false test`

Fixes #6722


Aias00 (approved):
Approved as PMC (Aias00). Reviewed the diff — small, well-scoped fix with regression tests; mergeable.

PR #7122, opened 2026-09-19
fix(ai): preserve headers on transform fallback

### LINKED ISSUE #6756: [BUG] Response transformer mutates response headers before validating body extraction
- severity: Medium
- files: `shenyu-plugin/shenyu-plugin-ai/shenyu-plugin-ai-response-transformer/src/main/java/org/apache/shenyu/plugin/ai/transformer/response/AiResponseTransformerPlugin.java:359-371`
- description: In the decorator's inner `flatMap`, the code does `this.getHeaders().clear(); this.getHeaders().putAll(newHeaders);` BEFORE checking whether `newBody` is non-null. If body extraction fails (returns null), the else-branch returns the ORIGINAL body bytes but with response headers already replaced by whatever the AI emitted. Original `Content-Type`/`Content-Length` are gone.
- impact: Corrupted response headers (and Content-Length mismatch) on any AI transformation where body extraction fails.
- suggested_fix: Only mutate headers after both header AND body extraction succeed.
- confidence: High
- related_existing: none — SEC-9 covers gzip decompression in the same file, not this ordering bug.

---

dengliming:
## Summary
- keep original response headers when the transformed response has no valid body
- apply AI-provided headers only when the transformed body is accepted
- add regression coverage for the fallback path

## Testing
- `mvn -q -pl shenyu-plugin/shenyu-plugin-ai/shenyu-plugin-ai-response-transformer -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=AiResponseTransformerPluginTest -DfailIfNoTests=false test`

Fixes #6756


Aias00 (approved):
Approved as PMC (Aias00). Reviewed the diff — small, well-scoped fix with regression tests; mergeable.

Voyagerroc-Lab (commented):
Reviewed the response-transformer fallback header ordering. Original response headers are now left intact until a non-empty transformed body is accepted, so invalid/empty AI output cannot overwrite the upstream response metadata; the regression test covers the fallback and asserts the injected header is absent. I found no blocking issue.

PR #7133, opened 2026-09-19
OSPP_PR1:Add request-scoped context support for the Agent Gateway plugin

BobSong-dev:
## Background

Add request-scoped context support through a new Agent Gateway plugin without changing the existing AI Proxy or MCP implementation. The plugin provides an immutable request context that can be propagated through the Reactor chain and optionally returned as a response request ID.

## Changes

- Add the `shenyu-plugin-agent-gateway` plugin and Spring Boot starter;
- Register plugin metadata and database initialization entries;
- Support strict rule-handle validation, Reactor context propagation, request isolation, response correlation, and cache cleanup;
- Keep the existing AI Proxy and MCP plugins reusable and independently configurable.

## Verification

- Local JDK 17 `verify`: 13 reactor modules passed;
- RAT: `Unapproved: 0`, `Unknown: 0`, `Generated: 0`;
- RISC-V container simulation with Java 17 and Maven: 13 reactor modules passed;
- Admin → WebSocket → Gateway configuration synchronization passed;
- AI Proxy JSON, SSE, cancellation, and MCP coexistence were manually verified.

## Scope

- Full MCP JSON-RPC tool workflow is not included;
- Native RISC-V hardware and GitHub CI have not been run;
- Other database dialects have not been executed;
- The upcoming mentor version has not yet been merged.


Aias00 (commented):
## Review (PMC Aias00) — `OSPP_PR1: Add request-scoped context support for the Agent Gateway plugin`

I am leaving this as a **comment**, not a formal approve/request-changes, because the PR is still a draft and the description says the mentor version has not been merged yet. But there are two hard blockers below that will break a fresh install, plus a few design points worth settling before this leaves draft.

---

### 🔴 Blocking — rebase required, the SQL IDs now collide with master

The branch was cut around 2026-09-19; master has since landed the `sensitiveWord` plugin (2026-09-21), which consumed exactly the ID range this PR allocates. Verified against current `master`:

1. **`plugin.id = '67'` is already taken.** All six init scripts insert
   `('67', 'agentGateway', ... 'Ai', 198 ...)`, but master already has
   ```sql
   INSERT INTO `plugin` VALUES ('67', 'sensitiveWord', null, 'Ai', 197, ...);
   ```
   → duplicate primary key, schema init fails on a fresh install. Please move to the next free id (`68`) in **all six** files (`db/init/mysql|ob|og|oracle|pg` + `shenyu-admin/.../h2/schema.sql`).

2. **`resource` ids `1942847622591684620` / `…621` / `…622` are already used.** Master has
   ```sql
   INSERT INTO `plugin_handle` VALUES ('1942847622591684620', '67', 'failClosed', ...);
   INSERT INTO `plugin_handle` VALUES ('1942847622591684621', '67', 'words', ...);
   INSERT INTO `plugin_handle` VALUES ('1942847622591684622', '67', 'maxBodySize', ...);
   ```
   This PR reuses the same three values for its `resource` rows → duplicate primary key. Re-allocate, and please re-verify the `permission` ids `1942847622591684630–639` against master as well while you are there.

Good news: plugin **code `198` itself is free** and slots correctly immediately before `AI_PROXY(199, ...)`, so the enum addition is fine as-is.

---

### 🟠 Should fix before leaving draft

3. **`trafficType` is hard-restricted to `"LLM"`.** `AgentGatewayRuleHandleParser` returns `invalid("trafficType must be LLM")` for anything else, and an invalid handle becomes a **503 on live traffic**. The PR claims MCP coexistence and only lists "full MCP JSON-RPC tool workflow" as out of scope — but as written the plugin cannot even be *configured* for MCP later without another code change. Either open the enum now (`LLM`, `MCP`, …) or state plainly in the README that PR1 supports LLM only.
4. **Unknown fields reject live traffic.** Any key outside `{trafficType, responseRequestId}` → invalid handle → 503. Strict validation is right at *config-save* time, but failing requests because a stored handle carries an extra key is fragile (admin schema evolution, hand-edited JSON, copy-paste from docs). Suggest logging a warning and ignoring unknown fields on the request path, keeping the strict check in admin.
5. **`503` is the wrong status for a configuration error.** `reject()` sets `SERVICE_UNAVAILABLE`, which tells clients and upstream proxies to retry — but a bad rule handle will never start working. Use `500` (server-side misconfiguration) or `400`, and reserve 503 for real unavailability. The `AGENT_GATEWAY_CONFIG_INVALID` code prefix is a nice touch.
6. **`selector.getContinued() == FALSE` → 503.** `continued=false` conventionally means "stop the chain after this plugin", not "misconfigured". Failing the request here is surprising. Consider just proceeding, or document the constraint clearly.
7. **The `AtomicBoolean subscribed` guard is risky.**
   ```java
   if (!subscribed.compareAndSet(false, true)) {
       return Mono.error(new IllegalStateException("...cannot be subscribed twice"));
   }
   ```
   WebFlux subscribes once, but this makes the returned `Mono` non-idempotent, so any `retry`, `repeat`, `cache()` or re-subscribing composition turns into a hard error. If double subscription is not a real scenario, drop the guard; if it is, handle it without failing the request.

---

### 🟡 Non-blocking

8. `resolveHandle` re-parses JSON on **every** request when the handle is invalid (invalid results are never cached). Consider caching the invalid handle too, keyed by the raw string.
9. `restorePreviousContext` relies on `Map#remove(k, v)` / `Map#replace(k, old, new)`. `DefaultServerWebExchange` backs attributes with a `ConcurrentHashMap` so it works today, but those default methods throw `UnsupportedOperationException` on other `Map` implementations — worth a comment stating the assumption.
10. `AgentTrafficContext` has no `toString()`; adding one (requestId + trafficType + selector/rule) makes correlation in logs much easier.
11. The Verification section says GitHub CI has not been run. Apache requires a green CI run before merge — please make sure the full matrix passes (and rerun after the rebase).

---

### ✅ What looks good

- `AgentTrafficContext` is genuinely immutable, created per subscription, and never stored in a global map or thread local — real request isolation, and the test asserting 12 distinct request ids proves it.
- `contextWrite` is applied to `chain.execute(exchange)`, so downstream plugins see the context — correct Reactor usage, and the test asserts exchange context == reactor context.
- `doFinally` restores/removes the previous attribute, including on cancel — good cleanup hygiene.
- `AgentGatewayPluginDataHandler` cleans the handle cache on `removeRule`, `removeSelector` (prefix scan) and `removePlugin` — no leak on config removal.
- Response correlation via `beforeCommit` + `X-Shenyu-Agent-Request-Id`, and deliberately *not* trusting a client-supplied id.
- 26 tests across plugin / parser / data handler / starter config, covering isolation, restore, cancellation, rejection and the response-header opt-in.

Solid foundation — fix the ID collisions and the status-code semantics and this is in good shape.

db/init/mysql/schema.sql
```diff
@@ -1018,6 +1028,7 @@ INSERT INTO `plugin` VALUES ('52', 'aiPrompt', null, 'Ai', 170, 0, '2023-12-20 1
 INSERT INTO `plugin` VALUES ('53', 'aiRequestTransformer', NULL, 'Ai', 65, 0, '2023-12-20 18:02:53', '2023-12-20 18:02:53', null);
 
 INSERT INTO `plugin` VALUES ('61', 'mcpServer', null, 'MCP', 180, 0, '2023-12-20 18:02:53', '2023-12-20 18:02:53', null);
+INSERT INTO `plugin` VALUES ('67', 'agentGateway', null, 'Ai', 198, 0, '2026-09-19 00:00:00', '2026-09-19 00:00:00', null);
```
Aias00: 🔴 **Blocking — primary key collision with master.**

`plugin.id = '67'` was assigned to `sensitiveWord` on master (2026-09-21):
```sql
INSERT INTO `plugin` VALUES ('67', 'sensitiveWord', null, 'Ai', 197, 0, '2026-09-21 00:00:00', ...);
```
Inserting `('67', 'agentGateway', ...)` here fails with a duplicate primary key, so a fresh install cannot initialise the schema.

Please rebase and move to the next free id (`68`) — and apply the same change to all six init scripts (`db/init/mysql`, `ob`, `og`, `oracle`, `pg`, and `shenyu-admin/src/main/resources/sql-script/h2/schema.sql`).

The plugin **code** `198` itself is free and correctly ordered just before `AI_PROXY(199)`, so only the row id needs to change.

db/init/mysql/schema.sql
```diff
@@ -2070,6 +2083,17 @@ INSERT INTO `resource` VALUES ('1844026199075534867', '1844026199075534860', 'SH
 INSERT INTO `resource` VALUES ('1844026199075534868', '1844026199075534860', 'SHENYU.BUTTON.PLUGIN.RULE.DELETE', '', '', '', 2, 0, '', 1, 0, 'plugin:mcpServerRule:delete', 1, '2022-05-25 18:02:58', '2022-05-25 18:02:58');
 INSERT INTO `resource` VALUES ('1844026199075534869', '1844026199075534860', 'SHENYU.BUTTON.PLUGIN.SYNCHRONIZE', '', '', '', 2, 0, '', 1, 0, 'plugin:mcpServer:modify', 1, '2022-05-25 18:02:58', '2022-05-25 18:02:58');
 
+INSERT INTO `resource` VALUES ('1942847622591684620', '1346775491550474240', 'agentGateway', 'agentGateway', '/plug/agentGateway', 'agentGateway', 1, 0, 'pic-center', 0, 0, '', 1, '2026-09-19 00:00:00', '2026-09-19 00:00:00');
```
Aias00: 🔴 **Blocking — these `resource` ids are already used on master.**

`1942847622591684620`, `…621` and `…622` are already taken by `plugin_handle` rows belonging to `sensitiveWord` (plugin 67):
```sql
INSERT INTO `plugin_handle` VALUES ('1942847622591684620', '67', 'failClosed', ...);
INSERT INTO `plugin_handle` VALUES ('1942847622591684621', '67', 'words', ...);
INSERT INTO `plugin_handle` VALUES ('1942847622591684622', '67', 'maxBodySize', ...);
```
Please re-allocate this id block, and double-check the paired `permission` ids `1942847622591684630–639` while you are at it.

shenyu-plugin/shenyu-plugin-agent-gateway/src/main/java/org/apache/shenyu/plugin/agent/gateway/AgentGatewayPlugin.java
```diff
@@ -0,0 +1,119 @@
...
+    private void registerResponseRequestId(final ServerWebExchange exchange, final String requestId) {
+        exchange.getResponse().beforeCommit(() -> {
+            final HttpHeaders headers = exchange.getResponse().getHeaders();
+            headers.set(AgentGatewayConstants.REQUEST_ID_HEADER, requestId);
+            return Mono.empty();
+        });
+    }
+
+    private Mono<Void> reject(final ServerWebExchange exchange, final String reason) {
+        exchange.getResponse().setStatusCode(HttpStatus.SERVICE_UNAVAILABLE);
```
Aias00: 🟠 `503 SERVICE_UNAVAILABLE` is the wrong status for a **configuration** error.

503 tells clients and upstream proxies that the failure is transient and worth retrying, but a malformed rule handle will never start working — so this invites pointless retry storms. Please use `500 INTERNAL_SERVER_ERROR` (server-side misconfiguration) or `400`, and keep 503 only for genuine unavailability.

shenyu-plugin/shenyu-plugin-agent-gateway/src/main/java/org/apache/shenyu/plugin/agent/gateway/AgentGatewayPlugin.java
```diff
@@ -0,0 +1,119 @@
...
+            return reject(exchange, "selector continued=false is not supported");
+        }
+        final AgentGatewayRuleHandle handle = resolveHandle(rule);
+        if (!handle.isValid()) {
+            return reject(exchange, handle.getErrorMessage());
+        }
+        final AtomicBoolean subscribed = new AtomicBoolean();
+        return Mono.defer(() -> {
+            if (!subscribed.compareAndSet(false, true)) {
+                return Mono.error(new IllegalStateException("agent gateway execution cannot be subscribed twice"));
```
Aias00: 🟠 This guard makes the returned `Mono` non-idempotent: any second subscription (e.g. a `retry`, `repeat`, or `cache()` composed downstream) turns into an `IllegalStateException` and fails the request.

WebFlux subscribes once, so if double subscription is not a scenario you actually hit, I would drop the guard. If it *is* a real scenario, handling it should not mean failing the request — consider logging and reusing the already-built context instead.

BobSong-dev:
Thanks for the suggestion. We will add strict validation before saving or importing configurations in Admin for the agentGateway rule.handle. Only trafficType and responseRequestId will be allowed, with validation for required fields and value types. At runtime, unknown fields in legacy configurations will still be logged and ignored to preserve compatibility and avoid affecting existing requests.

BobSong-dev:
I’ve added strict validation in Admin for agentGateway rule handles. Rule creation, updates, and configuration imports now reject unknown fields and invalid values before they are saved. At runtime, unknown fields in existing configurations are still logged and ignored to avoid disrupting requests. The Admin tests pass locally (31 tests), and the change is included in commit 8dc237378.

PR #7140, opened 2026-09-20
Fixes #6775: Guard ExtensionLoader joins during concurrent initialization

### LINKED ISSUE #6775: [BUG] ExtensionLoader.getJoins() fast-path returns null / throws NPE under concurrent init
- Severity: Medium
- Location:
`shenyu-spi/src/main/java/org/apache/shenyu/spi/ExtensionLoader.java:174-194,55,331-369`
- 
Description:
`getJoins()` takes a fast path when `extensionClassesEntity.size() == cachedInstances.size()`, returning `cachedInstances.values().stream().sorted(HOLDER_COMPARATOR).map(e -> e.getValue())`. A `Holder` is inserted via `putIfAbsent` (line 153) *before* its `value`/`order` is populated inside the `synchronized(objectHolder)` block in `createExtension`. If thread A has just done `putIfAbsent` for the last name (so size reaches N) but not yet `setValue`/`setOrder`, a concurrent thread B calling `getJoins()` sees the size match, takes the fast path, reads `getValue()==null`/`getOrder()==null`. `HOLDER_COMPARATOR = Comparator.comparing(Holder::getOrder)` then throws `NullPointerException` (null `Integer` key); absent that, the returned list contains a `null` element.
- 
Impact:
Concurrent `getJoins()` (dynamic plugin/route loading) can throw NPE or return a list with `null`, causing NPE at the call site.
- 
Suggested fix:
Populate `value`/`order` before inserting the Holder, or gate the fast path on a `volatile boolean allInitialized` set after the slow path completes; or use `ConcurrentHashMap.compute` so a Holder is never observable half-built.
- 
Confidence: Medium

---

BobSong-dev:
Fixes #6775

## Background

ExtensionLoader#getJoins() treated cachedInstances.size() matching the number of extension classes as proof that every cached Holder was ready. However, getJoin() publishes a new Holder before populating its order and value. A concurrent getJoins() call could therefore sort or return an incompletely initialized holder, causing an NPE or a null element.

## Changes

- Track Holder initialization with a volatile flag published after order and value are set;
- Use the cached fast path only when every cached Holder is fully initialized;
- Add a concurrent regression test for getJoins() observing an incomplete Holder.

## Verification

- .\mvnw.cmd -pl shenyu-spi test -DskipITs: 20 tests passed, 0 failures, 0 errors, 0 skipped;
- Checkstyle: 0 violations;
- git diff --check: passed;


Aias00 (approved):
## Review — `Fixes #6775: Guard ExtensionLoader joins during concurrent initialization`

**Decision: APPROVE**

### The bug and why the fix works
`getJoins()` treated `cachedInstances.size() == extensionClassesEntity.size()` as proof that every cached `Holder` was ready. But `getJoin(name)` publishes a bare `new Holder<>()` into `cachedInstances` via `putIfAbsent` *before* `createExtension` fills in `order` and `value`. A concurrent `getJoins()` could therefore take the fast path and either sort a `Holder` with a `null` order or emit a `null` element into the returned list.

Tracking readiness with an explicit `volatile boolean initialized` set **last** in `createExtension(name, holder)` is the right fix, and the ordering matters:

```java
holder.setOrder(classEntity.getOrder());
holder.setValue(o);
holder.setInitialized(true);   // published last
```

Because `initialized` and `value` are both `volatile`, the write to `initialized` is a release — any thread that observes `initialized == true` is guaranteed (happens-before) to also observe the preceding `order` and `value` writes. Publishing the flag first would have reintroduced the same race, so thanks for getting the ordering right.

Once a `Holder` is initialized it never goes back (there is no removal path on `cachedInstances`), so the two consecutive stream passes — `allMatch(Holder::isInitialized)` then `map(Holder::getValue)` — are safe: the second pass cannot observe a value that disappeared.

### Test quality
`testMultiThreadGetJoinsWithUninitializedHolder` has real discriminating power. `HasDefaultSPI` resolves to exactly one join (`subHasDefaultSPI`), and the test seeds `cachedInstances` with one uninitialized `Holder` under that same key, so `size()` matches at 1:
- **Without the fix** → fast path taken → `[null]` returned → `assertNotNull(joins.get(0))` fails.
- **With the fix** → `allMatch(isInitialized)` is false → slow path → `getJoin(...)` creates the instance → 1 non-null join.

Good regression test.

### Non-blocking suggestions
1. **Scope.** The three infra changes bundled here (`.github/scripts/resolve-ci-modules.sh` forcing a full build on `shenyu-spi` changes, `shenyu-zookeeper.yml` `NodePort` → `ClusterIP`, and `set -euo pipefail` in `build_k8s_cluster.sh`) are unrelated to #6775. They are individually reasonable — `ClusterIP` is the more correct exposure for an in-cluster ZK, and the CI-script change prevents stale SNAPSHOT reuse — but please consider splitting them into a separate PR next time; unrelated infra edits in a bug-fix PR make backporting harder.
2. Forcing a full reactor build on every `shenyu-spi` touch will noticeably lengthen CI for SPI PRs. It is defensible given SPI is consumed nearly everywhere, just worth being aware of the cost.

Full CI is green (all four `k8s-ingress-*` jobs, `e2e`, `pr_build`, license header, `Analyze (java)`).

Thanks for the careful fix.

shenyu-examples/shenyu-examples-dubbo/shenyu-examples-apache-dubbo-service/k8s/shenyu-zookeeper.yml
```diff
@@ -68,7 +68,7 @@ metadata:
     app: shenyu-zk
     all: shenyu-examples-dubbo
 spec:
-  type: NodePort
+  type: ClusterIP
```
Aias00, line 71: Nit (non-blocking): this change is unrelated to #6775.

`NodePort` → `ClusterIP` for the ZooKeeper service is arguably more correct (ZK is only consumed in-cluster by the dubbo example pods, so exposing a NodePort is unnecessary surface), and the k8s ingress CI job passed with it.

But bundling infra/example changes into a bug-fix PR makes backporting and bisecting harder. Please consider moving this (and the `set -euo pipefail` + `resolve-ci-modules.sh` edits) into a separate PR next time.

PR #7142, opened 2026-09-20
fix(register): preserve distinct failback entries

### LINKED ISSUE #6777: [BUG] addToFail dedup key omits namespaceId and identity fields — drops distinct / newer failed registrations
- Severity: Medium
- Location:
`shenyu-register-center/shenyu-register-client/shenyu-register-client-api/src/main/java/org/apache/shenyu/register/client/api/FailbackRegistryRepository.java:126` (meta key), `:140` (uri key), `:154` (apiDoc key), `:169` (mcp key), `:175-184` (addToFail skip)
- 
Description:
`addToFail` builds the failback key from a field subset that omits `namespaceId` (and per type, methodName/serviceName/parameterTypes/contextPath/ruleName/appName/version): meta key = `rpcType://host:port/path`; uri key = `host:port:rpcType`; apiDoc key = `contextPath:apiPath:httpMethod:rpcType`; mcp key derives from the embedded `MetaDataRegisterDTO` without namespaceId. When a Holder already exists for a key, `addToFail` returns immediately (`:177-179`): (a) a *different* DTO colliding on the coarse key is never enqueued, and (b) a *newer* version (updated rpcExt/status/instanceInfo/version) is discarded, so `accept()` later persists the *stale* original. The DTOs' own `equals/hashCode` include namespaceId and far more fields — the dedup key is inconsistent with DTO identity.
- 
Impact:
Failed metadata/uri/apiDoc/mcp registrations for one namespace/variant are silently dropped from failback, or retried with stale content; never persisted until the client happens to re-register.
- 
Suggested fix:
Derive the key from the same fields the DTO's `equals/hashCode` use (or use the DTO's `hashCode()`); on collision, *replace* the stored Holder with the newer payload instead of skipping.
- 
Confidence: High
- Related existing: none. FUNC-E3 is the server-side URI batch namespace bug; this is the client-side failback dedup key. #4801 is the server-side `FallbackShenyuClientRegisterService` stale-retry (different class). FUNC-E1/E2 are exhaustion/race, not coarse-key dedup.

---

dengliming:
## Summary
- include namespace and type-specific identity fields in failback registration keys
- replace an existing failback holder with the latest payload for the same identity
- cover metadata, URI, API document, and MCP collision cases plus stale payload replacement

## Testing
- `mvn -q -pl shenyu-register-center/shenyu-register-client/shenyu-register-client-api -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=FailbackRegistryRepositoryTest -DfailIfNoTests=false test`

Fixes #6777


shenyu-register-center/shenyu-register-client/shenyu-register-client-api/src/main/java/org/apache/shenyu/register/client/api/FailbackRegistryRepository.java
```diff
@@ -166,19 +168,28 @@ protected <T> void addFailureMcpDocRegister(final T t) {
...
+        return String.join(":", value(dto.getNamespaceId()), value(dto.getRpcType()), value(dto.getAppName()),
+                value(dto.getContextPath()), value(dto.getServiceName()), value(dto.getMethodName()),
+                value(dto.getParameterTypes()), value(dto.getRuleName()), value(dto.getHost()), value(dto.getPort()), value(dto.getPath()));
+    }
+
+    private static String value(final Object value) {
+        return Objects.toString(value, "");
+    }
+
+    private void addToFail(final Holder t) {
```
Aias00, line 186: `put`-first-then-check is strictly better than `get`-then-`put` here, and replacing the stale payload is the right semantic. But the replacement path never re-arms the retry, which makes the new behaviour quietly lossy in one case.

`FailureRegistryTask` is constructed as `super(key, TimeUnit.SECONDS.toMillis(10), 18)`, i.e. `retryCount = 18`, `retryLimit = (18 < 0) == false`. In `AbstractRetryTask#run`:

```java
if (!retryLimit && tickCount > retryCount) {
    logger.warn("Final failed to execute task, key:{}, retried:{}, task over.", key, tickCount);
    return;                       // gives up - no reschedule, and no remove(key)
}
```

When it gives up, the `Holder` stays in `concurrentHashMap` forever (`remove(key)` is only called on success, and only `accept()` - which throws - precedes it). So the sequence is:

1. registration for identity X fails -> holder stored, task scheduled;
2. task retries 18x over ~180s, all failing -> logs "task over", returns, holder **still in the map**;
3. any later failure for X hits `oldObj != null` -> returns early -> **no new task is ever scheduled**.

With this change we now also log `Updated failback registration payload, ...`, which reads as "will be retried" - but nothing will retry it. Before the change the entry was at least created once; now the sticky-entry window is the normal path for any long outage.

Suggested fix in this PR, pick one:
- have `addToFail` re-arm a `FailureRegistryTask` when it replaces a payload whose previous task is no longer live (e.g. keep a `Set<String>` of live keys, or let `FailureRegistryTask` remove itself from the map when retries are exhausted and let `addToFail` treat "absent" as "schedule"); or
- add a terminal hook so `AbstractRetryTask#run` can remove the key when it gives up, making step 3 fall through to the "new failure" branch naturally.

The second is cleaner but touches shared timer code, so the first (re-arm on replace) is probably the smaller blast radius.

Minor: the new `logger.warn("Updated failback registration payload, ...")` fires on every duplicate failure, so a bulk re-registration (one entry per method) produces one WARN per method. Consider `debug` for the replacement case and keep `warn` for the first-time enqueue below.

Aias00 (changes requested):
## Summary

Widens the failback identity keys (namespace + type-specific fields), removes the `getHttpMethod().toString()` NPE, and changes `addToFail` from `get`-then-`put` to `put`-then-check so a repeat failure replaces the stored payload with the latest one.

## Requesting changes - replacing the payload never re-arms the retry

`put`-first-then-check is a genuine improvement over `get`-then-`put`, and storing the newest payload is the right semantic. But the replacement branch returns without scheduling anything, and that makes the new behaviour lossy in a case this PR makes *more* reachable.

`FailureRegistryTask` is built as `super(key, TimeUnit.SECONDS.toMillis(10), 18)` -> `retryCount = 18`, `retryLimit = (18 < 0) == false`. `AbstractRetryTask#run`:

```java
if (!retryLimit && tickCount > retryCount) {
    logger.warn("Final failed to execute task, key:{}, retried:{}, task over.", key, tickCount);
    return;                       // no reschedule, and no remove(key)
}
```

`remove(key)` only runs on success (`doRetry` calls `accept(key)` then `remove(key)`; `accept` throws on failure, so `remove` is skipped). So when the task exhausts its 18 retries (~180s) the `Holder` is **left in `concurrentHashMap` forever**, and:

1. first failure for identity X -> holder stored, task scheduled;
2. task fails 18x -> logs "task over", returns, holder still in map;
3. every later failure for X -> `oldObj != null` -> early return -> **no new task, ever**.

With this PR we now also log `Updated failback registration payload, ...` on that path, which reads as "this will be retried" when nothing will retry it. During a sustained admin outage - the exact scenario failback exists for - the registration is silently dropped for the rest of the process lifetime.

Suggested fix, either:
- re-arm: track live keys (e.g. a `Set<String>` next to `concurrentHashMap`, or have `FailureRegistryTask` deregister itself when it gives up) and schedule a new task in `addToFail` when replacing a payload whose previous task is dead; or
- let `AbstractRetryTask#run` invoke a terminal callback that removes the key when retries are exhausted, so step 3 falls through to the "new failure" branch.

The first has the smaller blast radius; the second is cleaner but touches shared timer code.

## Also worth a look (inline)

The new `logger.warn("Updated failback registration payload, {}", t.getPath())` fires on every duplicate failure, so a bulk re-registration emits one WARN per method. I would keep `warn` for the first-time enqueue below and use `debug` for the replacement case.

## What is good here

- Fine-grained metadata identity (namespace, appName, contextPath, serviceName, methodName, parameterTypes, ruleName, host, port, path) genuinely fixes the collision where two methods of one service shared the old `rpcType://host:portpath` key.
- `value(...)` wrapping every element is required - `String.join` throws NPE on a null element - so dropping `getHttpMethod().toString()` for `value(getHttpMethod())` removes a real NPE as well as widening the key.
- `testTypeSpecificFailureIdentitiesDoNotCollide` and `testNewerFailureReplacesStalePayload` cover the two behaviours that matter.

Everything else looks right; the re-arm issue above is the one thing I would fix before merge.

Aias00 (approved):
APPROVE — switching my earlier REQUEST_CHANGES. This is the fix I was hoping for, and you picked the better of the two shapes.

**The permanent-registration-loss bug is closed.** My concern was that once `FailureRegistryTask` (10s × 18) exhausted its retries, `AbstractRetryTask#run` logged "Final failed to execute task" and returned *without* removing the key, leaving the `Holder` in `concurrentHashMap` forever. Every later failure for the same identity then hit `addToFail`, saw a non-null existing entry, and returned early — so after a long admin outage the registration could never be re-attempted, while the new "Updated failback registration payload" log gave the impression that it would be.

You fixed it at the right layer: `AbstractRetryTask` now calls a new `onRetryExhausted(key)` hook when the retry limit is reached, and `FailureRegistryTask` overrides it to `registerRepository.remove(key)`. So the stale `Holder` is cleared and the next failure for that identity creates a fresh `Holder` **and** a fresh timer task. That is strictly better than my alternative of re-arming in `addToFail`, because the lifecycle now lives in one place and any future `AbstractRetryTask` subclass gets the same behaviour for free. The default hook is a no-op, so no other subclass changes behaviour.

**The log spam is fixed too** — the replace path now logs at `debug` rather than `warn`, which was my other point.

The rest still holds up: `concurrentHashMap.put(key, holder)` before the null check is the correct order, the widened identities (`namespaceId` on metadata/URI/MCP, `version` on API docs, and the full metadata identity for metadata and MCP) remove the collisions I noted, `Objects.toString(value, "")` also fixes the latent NPE where `ApiDocRegisterDTO#getHttpMethod().toString()` was called unguarded, and `metaDataIdentity` being shared by the metadata and MCP paths removes the duplication.

Tests are good: `testFailuresFromDifferentNamespacesDoNotCollide` and `testTypeSpecificFailureIdentitiesDoNotCollide` cover the key collision directly, `testNewerFailureReplacesStalePayload` covers the replace semantics (including that the retry replays the *latest* payload), and `testRetryExhaustedRemovesFailure` plus the new `AbstractRetryTaskTest` cover the exhaustion hook at both layers.

One small note for follow-up, not blocking: `remove()` on exhaustion gives up on that registration permanently for this process — correct and much better than leaking, but it does mean a client that never recovers stays unregistered until something re-triggers registration. If that turns out to matter, the natural next step is a periodic re-registration sweep rather than changing this behaviour.

PR #7143, opened 2026-09-20
fix(register): track all HTTP registrations

### LINKED ISSUE #6778: [BUG] HttpClientRegisterRepository uses static mutable URI/ApiDoc fields — only the last DTO is offlined on shutdown, with cross-instance leakage
- Severity: Medium
- Location:
`shenyu-register-center/shenyu-register-client/shenyu-register-client-http/src/main/java/org/apache/shenyu/register/client/http/HttpClientRegisterRepository.java:59,61` (static fields), `:127` (`uriRegisterDTO = registerDTO`), `:157` (`apiDocRegisterDTO = registerDTO`), `:171-180` (closeRepository)
- 
Description:
`uriRegisterDTO` and `apiDocRegisterDTO` are `static` mutable fields overwritten on every `doPersistURI`/`doPersistApiDoc`. `closeRepository()` (priority-1 shutdown hook) only re-registers the *last*-registered URI and the *last*-registered apiDoc; every earlier one is skipped. Because the fields are `static`, multiple repository instances (multiple register types in one JVM, re-init, tests) share and overwrite each other's DTOs. No memory barrier between writer (register thread) and reader (shutdown hook thread).
- 
Impact:
Graceful shutdown only offlines the most-recent apiDoc (URI is mitigated by the separate per-URI OFFLINE hook in `ShenyuClientURIExecutorSubscriber:105-115`); stale apiDoc state remains for all earlier docs. Cross-instance overwrite can offline the wrong DTO.
- 
Suggested fix:
Make the fields instance (non-static) and collect all registered DTOs (e.g. a `ConcurrentHashMap`/set) so `closeRepository` can offline each.
- 
Confidence: High
- Related existing: none — #6487 (heartbeat scheduler), #6559/FUNC-E4 (register partial-failure) are different.

---

dengliming:
## Summary
- replace static last-registration fields with instance-scoped concurrent maps
- retain every URI and API document identity for shutdown offline events
- add regression coverage for multiple registrations and repository isolation

## Testing
- `mvn -q -pl shenyu-register-center/shenyu-register-client/shenyu-register-client-http -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=HttpClientRegisterRepositoryTest -DfailIfNoTests=false test`

Fixes #6778


shenyu-register-center/shenyu-register-client/shenyu-register-client-http/src/main/java/org/apache/shenyu/register/client/http/HttpClientRegisterRepository.java
```diff
@@ -169,14 +171,28 @@ protected void doPersistMcpTools(final McpToolsRegisterDTO registerDTO) {
...
     public void closeRepository() {
-        if (Objects.nonNull(uriRegisterDTO)) {
-            uriRegisterDTO.setEventType(EventType.DELETED);
-            doRegister(uriRegisterDTO, Constants.URI_PATH, Constants.URI);
-        }
-        if (Objects.nonNull(apiDocRegisterDTO)) {
-            apiDocRegisterDTO.setEventType(EventType.OFFLINE);
-            doRegister(apiDocRegisterDTO, Constants.API_DOC_PATH, Constants.API_DOC_TYPE);
-        }
+        uriRegisterDTOs.values().forEach(registerDTO -> {
```
Aias00, line 174: The move from `static` fields to per-instance maps is correct and fixes real cross-instance leakage (two `HttpClientRegisterRepository` instances in one JVM previously overwrote each other's pending offline payload, and the `resetStatics()` reflection helper in the test was a symptom of that).

Two follow-ups, non-blocking:

1. `closeRepository()` iterates the maps but never clears them. A second call (or a close followed by any later re-registration + close) re-sends `DELETED`/`OFFLINE` for entries that were already offlined, and the repository keeps holding every DTO until it is GC'd. `uriRegisterDTOs.clear()` / `apiDocRegisterDTOs.clear()` after the loops would make close idempotent.

2. `uriIdentity` / `apiDocIdentity` here duplicate the identity logic that PR #7142 adds to `FailbackRegistryRepository` (`metaDataIdentity`). The two key spaces can drift apart silently - failback retries and shutdown offline events would then disagree about what "the same registration" means. Consider a small shared helper so both call sites derive identity from one place.

Also note `apiDocIdentity` has no namespace dimension (`ApiDocRegisterDTO` carries no `namespaceId`), so API docs with the same path in different namespaces still collapse to one entry - same limitation as #7142. Fine for now, but worth a comment in the code so it is a known gap rather than an oversight.

Aias00 (approved):
## Summary

Replaces the two `static` "last registration" fields in `HttpClientRegisterRepository` with per-instance `ConcurrentHashMap`s, so `closeRepository()` offlines *every* URI / API doc the instance registered instead of only the most recent one, and two repositories in the same JVM no longer clobber each other.

## Why this is correct

`private static URIRegisterDTO uriRegisterDTO` / `private static ApiDocRegisterDTO apiDocRegisterDTO` were process-wide, so:
- with multiple registered URIs, only the last one received the `DELETED` event on close (the rest stayed registered upstream);
- two `HttpClientRegisterRepository` instances (multi-client apps, tests) shared one slot.

The `resetStatics()` reflection helper that this PR deletes from the test was itself the smell that exposed the design problem. Instance-scoped maps with an identity key are the right shape, and the identity (`namespaceId / protocol / appName / contextPath / rpcType / host / port` for URI, `contextPath / apiPath / httpMethod / rpcType / version` for API doc) covers the fields that actually distinguish registrations. Verified all referenced getters exist on `URIRegisterDTO` / `ApiDocRegisterDTO`.

Tests are the right ones to add: `closeRepositoryShouldUnregisterEveryUriAndApiDoc` (4 + 4 invocations instead of 2 + 2) and `repositoriesShouldTrackRegistrationsIndependently` (proves instance isolation, which is exactly what the static field broke).

## Non-blocking follow-ups (inline)

1. `closeRepository()` never clears the maps, so a second close re-sends `DELETED`/`OFFLINE`, and the DTOs stay reachable until the repository is GC'd. Clearing after the loops makes close idempotent.
2. `uriIdentity` / `apiDocIdentity` duplicate the identity logic PR #7142 adds to `FailbackRegistryRepository` as `metaDataIdentity`. Retry keys and shutdown-offline keys can silently drift apart; a small shared helper would keep them in sync.
3. `ApiDocRegisterDTO` carries no `namespaceId`, so API docs with the same path in different namespaces still collapse into one entry. Acceptable, but worth a code comment so it reads as a known limitation.

Approving - the core change is a real bug fix with matching regression coverage, and the three items above are follow-up material.

PR #7144, opened 2026-09-20
fix(register): propagate unregister failures

### LINKED ISSUE #6779: [BUG] doUnregister swallows all offline exceptions — offline failures never retried, leaving stale upstreams
- Severity: Medium
- Location:
`HttpClientRegisterRepository.java:232-246` (`doUnregister`), called from `offline()` (`:131-133`) and the per-URI shutdown hook `ShenyuClientURIExecutorSubscriber:109`
- 
Description:
`doUnregister` catches `Exception` per server, logs, and never throws — it lacks the `if (i == serverList.size()) throw new RuntimeException(e)` that `doRegister` (`:205-207`) and `doHeartbeat` (`:225-227`) have. `offline()` is not wrapped by `FailbackRegistryRepository.persist*`, so no failback retry is enqueued. This is the *only* offline mechanism (the per-URI shutdown hook also routes through `offline()`). If all admin servers are unreachable at shutdown, the offline is silently lost.
- 
Impact:
Stale upstream entries after client shutdown when admin is temporarily unavailable; gateway may route traffic to dead instances during the health-check grace window.
- 
Suggested fix:
Throw on last-server failure (mirroring `doRegister`) and/or route offline through a failback retry; alternatively document as best-effort.
- 
Confidence: Medium
- Related existing: none — #6559/FUNC-E4 concerns `doRegister` partial-failure; #6487 concerns the heartbeat scheduler. `doUnregister` is a distinct method with no retry.

---

dengliming:
## Summary
- propagate the final admin-server failure from URI unregister operations
- retain the original exception and stack trace in logs
- add regression coverage for failed offline requests

## Testing
- `mvn -q -pl shenyu-register-center/shenyu-register-client/shenyu-register-client-http -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=HttpClientRegisterRepositoryTest -DfailIfNoTests=false test`

Fixes #6779


shenyu-register-center/shenyu-register-client/shenyu-register-client-http/src/main/java/org/apache/shenyu/register/client/http/HttpClientRegisterRepository.java
```diff
@@ -240,7 +242,10 @@ private <T> void doUnregister(final T t) {
                 RegisterUtils.doUnregister(GsonUtils.getInstance().toJson(t), concat, accessToken);
                 // considering the situation of multiple clusters, we should continue to execute here
             } catch (Exception e) {
-                LOGGER.error("Unregister admin url :{} is fail. cause:{}", server, e.getMessage());
+                LOGGER.error("Unregister admin url :{} is fail.", server, e);
+                if (i == serverList.size()) {
+                    throw new RuntimeException(e);
```
Aias00, line 263: This makes `doUnregister` consistent with `doRegister` / `doHeartbeat`, which I agree with in principle. But the only production caller is a JVM shutdown hook, and propagating here has a concrete side effect worth checking before merge:

`ShenyuClientURIExecutorSubscriber` (shenyu-client-core) registers:

```java
ShutdownHookManager.get().addShutdownHook(new Thread(() -> {
    ...
    shenyuClientRegisterRepository.offline(offlineDTO);   // <-- now can throw

    // shutdown heartbeat executor
    if (!executor.isTerminated()) {
        executor.shutdown();                              // <-- skipped if the line above throws
    }
}), 2);
```

`ShutdownHookManager` wraps each hook in `try { hook.run(); } catch (Throwable ex) { LOG.error(...); }`, so the throw is swallowed and logged - but it aborts the rest of *that* hook, so the heartbeat executor is never shut down. Previously the unregister failure was logged and the cleanup still ran.

Net effect: the "propagation" never reaches anything that can act on it (there is no failback for `offline`, unlike `persistURI`), and it costs us the executor shutdown. Options:
- move the `executor.shutdown()` into a `finally` in the subscriber (or a separate hook) so cleanup is exception-proof, or
- keep `doUnregister` non-throwing and just make the failure visible another way.

If you go with propagation, please also cover the multi-server ordering semantics noted in the review - `i == serverList.size()` means "the *last* server failed", not "every server failed".

Aias00 (changes requested):
## Summary

`doUnregister` now rethrows when the last admin server in `serverList` fails, matching the existing `doRegister` / `doHeartbeat` behaviour, and logs the exception object instead of only `e.getMessage()`.

## Requesting changes - the propagation lands in a shutdown hook

Consistency with `doRegister`/`doHeartbeat` is the right instinct, but `offline()` has a different blast radius than `persistURI()`: there is no failback path behind it. The only production caller is a JVM shutdown hook in `ShenyuClientURIExecutorSubscriber` (shenyu-client-core):

```java
ShutdownHookManager.get().addShutdownHook(new Thread(() -> {
    final URIRegisterDTO offlineDTO = new URIRegisterDTO();
    BeanUtils.copyProperties(uriRegisterDTO, offlineDTO);
    offlineDTO.setEventType(EventType.OFFLINE);
    shenyuClientRegisterRepository.offline(offlineDTO);   // now throws

    // shutdown heartbeat executor
    if (!executor.isTerminated()) {
        executor.shutdown();                              // skipped on throw
    }
}), 2);
```

`ShutdownHookManager` wraps each hook in `try { hook.run(); } catch (Throwable ex) { LOG.error(...); }`, so the exception is swallowed and logged - it does not surface anywhere that can react to it. What it *does* do is abort the remainder of that hook, so `executor.shutdown()` no longer runs and the heartbeat executor leaks. Before this change the failure was logged and cleanup still happened.

Net: the propagated exception is never actionable, and it costs us clean shutdown. Please pick one:
- make the subscriber resilient (`finally` around `executor.shutdown()`, or split cleanup into its own hook) so propagation is safe, **or**
- keep `doUnregister` non-throwing and surface the failure some other way.

The new test `offlineShouldThrowWhenEveryServerFails` is good coverage for whatever you decide; it just needs to be paired with the caller fix.

## Also worth fixing (inline)

`i == serverList.size()` encodes "the *last* server failed", not "every server failed". With `serverList = [A, B]`, if A succeeds and B fails we throw even though the unregister partly succeeded. That is the same semantics `doRegister`/`doHeartbeat` already have, so this PR is consistent - but the sibling test is named `doPersistUriShouldThrowWhenEveryServerFails` and only uses a single-server config, so the mismatch is currently untested. Either track a failure count and throw only when all failed, or align the naming/expectation.

Minor: a bare `RuntimeException` is used. `ShenyuException` would be more idiomatic, though `doRegister`/`doHeartbeat` also use `RuntimeException`, so changing it here alone would be inconsistent - your call.

Aias00 (approved):
APPROVE — switching my earlier REQUEST_CHANGES. Both points are addressed, and the second one is fixed properly rather than papered over.

**1. Shutdown cleanup now survives an `offline()` failure.** Previously the hook called `offline(...)` and then `executor.shutdown()` as two sequential statements, so a throw from `offline` skipped the shutdown — and this hook is the only production caller of `offline`, with `ShutdownHookManager` merely catching and logging, so nothing was left to act on the failure. Now:

```java
try {
    shenyuClientRegisterRepository.offline(offlineDTO);
} finally {
    if (!executor.isTerminated()) {
        executor.shutdown();
    }
}
```

`testOfflineFailureStillShutsDownHeartbeatExecutor` pins exactly this (mock `offline` to throw, assert `executor.isShutdown()` afterwards). Extracting `offlineAndShutdown` to make it reachable from the test is a reasonable trade.

**2. "all servers failed" is now actually all servers.** The old condition was `if (i == serverList.size())`, which is true when the *last* server fails regardless of the others. It now counts:

```java
failureCount++;
LOGGER.error("Unregister admin url :{} is fail.", server, e);
if (failureCount == serverList.size()) {
    throw new RuntimeException(e);
}
```

and `offlineShouldNotThrowWhenOnlyLastServerFails` covers precisely the case the old check got wrong — two servers, only the second failing, assert no throw and that *both* were attempted. `offlineShouldThrowWhenEveryServerFails` covers the opposite. Good pair.

Also appreciated: the log line now passes the exception `e` instead of only `e.getMessage()`, so the stack trace survives — which matters a lot for a failure that is happening during shutdown.

The propagation behaviour itself still matches `doRegister`/`doHeartbeat`, which is what I asked for originally.

PR #7145, opened 2026-09-20
fix(consul): use concurrent watcher state maps

### LINKED ISSUE #6783: [BUG] ConsulSyncDataService uses non-thread-safe HashMap for shared state accessed by 7 concurrent watcher threads
- Severity: Medium
- Location:
`shenyu-sync-data-center/shenyu-sync-data-consul/src/main/java/org/apache/shenyu/sync/data/consul/ConsulSyncDataService.java:63,65` (HashMap fields), `:98` (7-thread `ScheduledThreadPoolExecutor`), `:103-111` (7 scheduled `watcherData0` tasks)
- 
Description:
`consulIndexes` (`Map<String, Long>`) and `cacheConsulDataKeyMap` (`Map<String, List<ConsulData>>`) are plain `HashMap`. The executor has 7 threads and schedules 7 independent `watchConfigKeyValues` tasks (one per watcher path). Each task performs `consulIndexes.get`, `consulIndexes.containsValue` (full-map scan), `consulIndexes.put`, `cacheConsulDataKeyMap.get`, `cacheConsulDataKeyMap.put` on these shared maps. `HashMap` is not safe for concurrent access — structural modifications (put during resize) can corrupt internal structure or lose data; `containsValue` racing with `put` produces inconsistent reads.
- 
Impact:
Under normal multi-path consul operation, concurrent access corrupts data, loses index updates (watcher misses config changes or re-processes them), inconsistent gateway cache state.
- 
Suggested fix:
Change both to `ConcurrentHashMap`.
- 
Confidence: High
- Related existing: none — #5545 (consul client metadata register fails) is a different component/root cause.

---

dengliming:
## Summary
- replace shared watcher index and cache HashMaps with ConcurrentHashMaps
- preserve missing-index behavior without storing null values
- add regression coverage for concurrent watcher state maps

## Testing
- `mvn -q -pl shenyu-sync-data-center/shenyu-sync-data-consul -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=ConsulSyncDataServiceTest -DfailIfNoTests=false test`

Fixes #6783


shenyu-sync-data-center/shenyu-sync-data-consul/src/main/java/org/apache/shenyu/sync/data/consul/ConsulSyncDataService.java
```diff
@@ -60,9 +60,9 @@ public class ConsulSyncDataService extends AbstractPathDataSyncService {
      */
     private static final Logger LOG = LoggerFactory.getLogger(ConsulSyncDataService.class);
 
-    private final Map<String, Long> consulIndexes = new HashMap<>();
+    private final Map<String, Long> consulIndexes = new ConcurrentHashMap<>();
```
Aias00, line 63: Swapping `HashMap` for `ConcurrentHashMap` is the right call here: `watcherData0(...)` schedules one self-rescheduling watcher per path on a `ScheduledThreadPoolExecutor(7)`, so up to seven threads mutate these two maps concurrently.

One thing this does *not* fix, worth tracking as a follow-up rather than blocking: `watchConfigKeyValues` still does a check-then-act across the whole map -

```java
if (!this.consulIndexes.containsValue(newIndex) && !currentIndex.equals(ConsulConstants.INIT_CONFIG_VERSION_INDEX)) { ... }
this.consulIndexes.put(watchPathRoot, newIndex);
```

`containsValue` is weakly consistent and the read-modify-write is not atomic, so two watchers that observe the same `newIndex` can both pass the guard and publish the same event twice. `ConcurrentHashMap` removes the corruption risk (which is the real bug), but not the duplicate-publish race. If you want to close that too, `compute`/`merge` on a per-path entry would be the natural shape.

shenyu-sync-data-center/shenyu-sync-data-consul/src/test/java/org/apache/shenyu/sync/data/consul/ConsulSyncDataServiceTest.java
```diff
@@ -131,8 +132,19 @@ public void testWatchConfigKeyValues() throws NoSuchMethodException, IllegalAcce
...
     }
+
+    @Test
+    public void testWatcherStateUsesConcurrentMaps() throws NoSuchFieldException, IllegalAccessException {
+        Field consulIndexesField = ConsulSyncDataService.class.getDeclaredField("consulIndexes");
+        consulIndexesField.setAccessible(true);
+        Field cacheDataField = ConsulSyncDataService.class.getDeclaredField("cacheConsulDataKeyMap");
+        cacheDataField.setAccessible(true);
+
+        Assertions.assertTrue(consulIndexesField.get(consulSyncDataService) instanceof ConcurrentHashMap);
```
Aias00: This assertion locks the test to a concrete implementation type rather than to behaviour. It will fail if someone later changes the field to `ConcurrentMap<String, Long>`, wraps it with `Collections.synchronizedMap`, or switches to a different concurrent map - even though the production behaviour would still be correct. It also proves nothing about concurrency: a map can be a `ConcurrentHashMap` and the surrounding code can still be unsafe.

`remove(watchPathRoot)` above is the valuable part of this test - it actually drives the `Objects.isNull(currentIndex) -> INIT_CONFIG_VERSION_INDEX` branch, whereas the old `put("/null", null)` wrote a key the code never read. I would keep that and either drop this test or make it behavioural, e.g. mutate the two maps from N threads and assert no lost updates / no exception. If you keep a structural check, at least assert assignability to `ConcurrentMap` instead of `instanceof ConcurrentHashMap`.

Aias00 (dismissed):
## Summary

Replaces the two shared watcher-state `HashMap`s with `ConcurrentHashMap`s in `ConsulSyncDataService`, and drops the now-illegal `null` value from the test.

## Why this is correct

`watcherData0(...)` is called once per watched path (selector / rule / proxy-selector / discovery-upstream / app-auth / meta-data) and each call schedules a self-rescheduling `watchConfigKeyValues` task on a single `ScheduledThreadPoolExecutor(7)`. Those tasks run concurrently and all read/put into `consulIndexes` and `cacheConsulDataKeyMap`. Unsynchronized `HashMap` access from those threads is a genuine data-race / potentially-corrupted-map bug, so `ConcurrentHashMap` is the right fix.

The test change is also required, not cosmetic: `ConcurrentHashMap` rejects `null` values, so `consulIndexesSource.put("/null", null)` would now throw NPE. Replacing it with `remove(watchPathRoot)` is arguably better coverage than what it replaced - the old entry used key `/null`, which `watchConfigKeyValues` never reads, so it exercised nothing; removing `watchPathRoot` actually drives the `Objects.isNull(currentIndex) -> INIT_CONFIG_VERSION_INDEX` branch.

## Non-blocking follow-ups (inline)

1. The `containsValue(newIndex)` guard followed by `put(...)` is still a non-atomic check-then-act across the whole map, so two watchers seeing the same index can publish the same event twice. Pre-existing; `ConcurrentHashMap` removes the corruption risk, which is the actual bug being fixed here. See inline comment.
2. `testWatcherStateUsesConcurrentMaps` asserts a concrete implementation type via reflection rather than behaviour. It will break on any future equivalent implementation and proves nothing about concurrency. The `remove(watchPathRoot)` change above is the valuable part of the test; I would keep that and treat the type assertion as optional.

Approving - the production change is correct and the risk is low, and #1/#2 are worth tracking but not worth blocking a race fix on.

Aias00 (approved):

PR #7131, opened 2026-09-19
fix(modify-response): preserve non-json bodies

dengliming:
## Summary
- bypass body parsing when a rule only changes response headers or status
- skip JSON body mutations for explicitly non-JSON content types
- preserve non-JSON response bytes while still applying header changes

## Testing
- `mvn -q -pl shenyu-plugin/shenyu-plugin-modify-response -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=ModifyResponsePluginTest -DfailIfNoTests=false test`

Fixes #6771

PR #7148, opened 2026-09-20
fix: fix ci

yuluo-yx:
case by https://github.com/apache/shenyu/pull/7131


yuluo-yx:
pr run successed, total cost 18m.

Aias00 (approved):

PR #7152, opened 2026-09-21
Fixes #6819: harden Dockerfiles for example and integrated test images

### LINKED ISSUE #6819: [Task] — Integrated-test & examples Dockerfiles run as root, no HEALTHCHECK, no LABEL
- severity: Medium; files: all of `shenyu-integrated-test/*/Dockerfile` and `shenyu-examples/*/Dockerfile` (verified via grep — zero `USER`/`HEALTHCHECK`/`LABEL` matches)
- description: e.g. `shenyu-integrated-test-http/Dockerfile` ends with `CMD java -jar ${APP_NAME}.jar` with no `USER`, `HEALTHCHECK`, or OCI `LABEL`. The release dist Dockerfiles (`shenyu-dist/shenyu-{admin,bootstrap}-dist/docker/Dockerfile`) correctly set `USER shenyu`, but these do not.
- impact: Containers run as root; no liveness probe; no image metadata.
- suggested_fix: Add a non-root `USER`, `HEALTHCHECK`, and `LABEL org.opencontainers.image.*`.
- confidence: High

---

BobSong-dev:
## Background

The Dockerfiles under shenyu-examples and shenyu-integrated-test run as root and do not provide OCI image metadata or container liveness checks. This makes the example and integration-test images harder to operate safely and consistently.

Fixes #6819

## Changes

- Add OCI image title, source, and license labels to all 39 target Dockerfiles.
- Create a dedicated shenyu system user and run each image as that non-root user.
- Add a portable process-level liveness HEALTHCHECK without introducing extra runtime packages.
- Keep the upload-plugin custom artifact image buildable by checking the packaged plugin artifact instead of treating it as an application service.

## Verification

- git diff --check
- Static validation confirmed all 39 target Dockerfiles contain LABEL, USER shenyu, and HEALTHCHECK, with no build-stage RUN instructions after the user switch.
- Dockerfile BuildKit checks were attempted for representative CentOS, Alpine, and custom-plugin images


shenyu-examples/shenyu-examples-http/Dockerfile
```diff
@@ -17,13 +17,21 @@
...
 RUN mkdir -p ${LOCAL_PATH}
 
 ADD target/${APP_NAME}.jar ${LOCAL_PATH}
 
 WORKDIR ${LOCAL_PATH}
+RUN groupadd --system shenyu && \
+    useradd --system --gid shenyu --create-home --home-dir /home/shenyu shenyu && \
+    chown -R shenyu:shenyu ${LOCAL_PATH}
+USER shenyu
+HEALTHCHECK --interval=30s --timeout=5s --start-period=30s --retries=3 CMD ["sh", "-c", "kill -0 1"]
```
Aias00: `kill -0 1` asserts nothing about the application.

PID 1 is by definition the container main process, so while the container is up this probe can never fail; it reports "the process exists", which Docker guarantees anyway. I verified this locally (Docker 29.4.0) with the same USER + HEALTHCHECK pattern on a busybox image:

    $ docker inspect hc1 --format '{{.State.Health.Status}}'
    healthy
    $ docker exec hc1 sh -c 'nc -z 127.0.0.1 8189 || echo "port 8189 NOT listening (app not ready/broken)"'
    port 8189 NOT listening (app not ready/broken)

The container reports healthy while nothing is listening on the service port, and it stays healthy through startup, a deadlock, or a refusing JVM.

That is a false signal rather than no signal, because this repository already consumes health status: `docker compose up --wait` is being adopted as a readiness gate (#7151 for the storage e2e scripts) and `healthcheck.sh` gates the integrated tests. Once one of these images is pulled into a `--wait`-based pipeline, this probe declares it ready ~30s after start regardless of what the JVM is doing.

Please either make it a real readiness probe (the actuator health endpoint where the image exposes one, otherwise a TCP connect against the application port using whatever the base image provides), or drop the HEALTHCHECK line and land only the non-root user + labels. Note that the custom-plugin image uses `test -s /opt/shenyu-custom-plugin.jar`, which is a build-time invariant - same tautology, different flavour.

Aias00 (changes requested):
Running these images as a non-root user and adding OCI metadata labels is a genuine improvement, and doing it consistently across all 22 example and integrated-test images (`addgroup`/`adduser` on the alpine bases, `groupadd`/`useradd` on the centos7 ones) is what a hardening PR should look like. Ownership of ${LOCAL_PATH} looks complete, the copied entries is chowned where relevant, and every exposed port is > 1024 so nothing needs root.

I am requesting changes on exactly one point.

[blocking] HEALTHCHECK ... CMD ["sh", "-c", "kill -0 1"] asserts nothing

PID 1 is the container main process by definition, so this probe can never fail while the container is up - it says "a process exists", which Docker already guarantees. Locally (Docker 29.4.0) with the same USER + HEALTHCHECK pattern:

    $ docker inspect hc1 --format '{{.State.Health.Status}}'
    healthy
    $ docker exec hc1 sh -c 'nc -z 127.0.0.1 8189 || echo "port 8189 NOT listening (app not ready/broken)"'
    port 8189 NOT listening (app not ready/broken)

Healthy while nothing is listening on the service port, and equally healthy through startup, a deadlock or a refusing JVM.

That matters because this repository already reads health status: `docker compose up --wait` is being adopted as the readiness gate (#7151 for the storage e2e scripts) and `healthcheck.sh` gates the integrated tests. The moment one of these images enters such a pipeline it is declared ready ~30s after start whatever the JVM is doing - a false signal, which is worse than none.

Please choose one:
  (a) a real readiness probe per image - the actuator health endpoint where the image exposes one, otherwise a TCP connect against the application port using whatever the base image provides (and keep it working under `USER shenyu`);
  (b) drop the HEALTHCHECK lines and land only the non-root user + labels.

Same issue in a different flavour in `shenyu-integrated-test-upload-plugin/shenyu-custom-plugin`: `test -s /opt/shenyu-custom-plugin.jar` is a build-time invariant and cannot fail at runtime either.

[nit] The two alpine images (`shenyu-examples-apache-dubbo-service-xml` and `shenyu-examples-apache-dubbo-service`) create the user with `adduser -S shenyu -G shenyu` without a home directory, while the centos7 variants use `--create-home --home-dir /home/shenyu`. Please add `-h /home/shenyu` so `$HOME` is not `/` and behaviour matches across images.

CI note (not a blocker): the integrated-test jobs are red here, but not because of these changes. The failure is

    Unable to build image [apache/shenyu-integrated-test-upload-plugin-case:latest]:
    "invalid from flag value apache/shenyu-integrated-test-custom: pull access denied for
     apache/shenyu-integrated-test-custom, repository does not exist or may require 'docker login'"

plus fail-fast cancellation of the remaining cases - an image publishing/environment problem, please just re-run. The e2e-case jobs that do not depend on that base image are green.

Aias00 (approved):
APPROVE — switching my earlier REQUEST_CHANGES.

My blocking point was that `HEALTHCHECK ... kill -0 1` is a tautology: I verified locally on Docker 29.4.0 that a container reports `healthy` even when nothing is listening on the service port, which would have fed false signals into the `docker compose up --wait` readiness gate being introduced at the same time.

That is fixed. The probes are now real:

```dockerfile
HEALTHCHECK --interval=30s --timeout=5s --start-period=30s --retries=3 \
  CMD ["sh", "-c", "wget -q -O - http://127.0.0.1:9195/actuator/health | grep -q UP"]
```

with `/dev/tcp/...` TCP probes for the two images that do not expose an actuator endpoint. The pipeline returns `grep`'s status, so a refused connection (empty body) correctly fails, and `--start-period=30s --retries=3` gives ~120s to come up, which fits inside the `--wait-timeout 300` used by the storage e2e scripts.

CI is fully green, including all of `shenyu-integrated-test-*`, `shenyu-integrated-test-k8s-ingress-*` and `build-docker-images`.

Two things to keep an eye on, neither blocking:

1. `wget` now has to exist in every image you touched. The ones CI builds and runs are proven, but there are ~45 Dockerfiles here and not all of them are exercised — if any image is distroless or otherwise missing `wget`/`busybox`, its container will never reach `healthy` and anything using `--wait` on it will time out. Worth a pass over the images that CI does not build.
2. `/actuator/health` returns `UP` by default even when a downstream dependency (database, registry) is unavailable, unless management health details are configured. That is fine for a "process is serving" probe, which is what a compose readiness gate needs — just don't read it as "fully ready to take traffic".

Thanks for the follow-through; the non-root user and OCI label work from the original commit still looks good.

PR #7158, opened 2026-09-22
fix(plugin): bound metadata path cache

### LINKED ISSUE #6565: [BUG] MetaDataCache: unbounded CACHE (maximumSize=Integer.MAX_VALUE) + MAPPING Set never shrunk
## Description
`CACHE` is a `WindowTinyLFUMap` built with `maximumSize = Integer.MAX_VALUE`, which disables size-based eviction entirely; there is no `expireAfterWrite`. The class comment at line 148 even acknowledges OOM risk yet configures an unbounded size. `MAPPING` (`ConcurrentMap<String, Set<String>>`) is populated by `initCache`: for wildcard metadata paths the reverse-mapping `ConcurrentSkipListSet` under each `metaPath` collects every distinct request path ever matched. `clean(key)` removes CACHE entries but **never clears the Set itself and never removes it from MAPPING**; `remove(MetaData)` only calls `clean(path)`.

## Location
```
shenyu-plugin-base/.../cache/MetaDataCache.java:48 (CACHE), 53 (MAPPING), 99-107 (clean), 147-157 (initCache)
```

## Impact
`MetaDataCache.obtain(path)` is called every request from the global plugin and divide/context-path/rewrite plugins. High-cardinality request paths (e.g. `/api/users/{userId}` with millions of distinct IDs) cause monotonic, unbounded heap growth in both CACHE and the MAPPING `ConcurrentSkipListSet` (memory-heavy per entry) that survives metadata refreshes. Slow OOM over days/weeks for public-facing REST gateways.

## Suggested fix
Set a real configurable `maximumSize` and `expireAfterAccess`; in `clean(key)`, after removing CACHE entries, also `MAPPING.get(key).clear()` and `MAPPING.remove(key)`.

## Related existing issue(s)
#6506 is the CachePlugin (response caching); this is the metadata path cache. #6479 is discovery-upstream cache staleness; this is selector/metadata cache growth.

dengliming:
## Summary
- cap resolved metadata path entries at the shared cache limit
- remove the unbounded reverse path mapping and invalidate entries from the bounded cache directly
- make cache size observations apply pending Caffeine maintenance
- add regression coverage for capacity and wildcard invalidation

## Testing
- `mvn -q -pl shenyu-plugin/shenyu-plugin-global -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=MetaDataCacheTest -DfailIfNoTests=false test`

Fixes #6565


moremind (approved):

Aias00 (approved):
APPROVE — the unbounded cache is real, and the replacement `clean()` preserves the old semantics.

**The leak is real.** `CACHE` was built as

```java
new WindowTinyLFUMap<>(1 << 16, Integer.MAX_VALUE, Boolean.FALSE)
```

i.e. `initialCapacity` 65536 with `maximumSize` `Integer.MAX_VALUE` — so despite the comment ("The extreme case will lead to OOM, that's why use LRU") the LRU never actually evicted anything. And the auxiliary `MAPPING` (`metaPath -> Set<path>`) was itself an unbounded `ConcurrentMap` of `ConcurrentSkipListSet`, so it was a second leak of the same shape. Bounding both to `Constants.CACHE_MAX_COUNT` and dropping `MAPPING` is the right call.

**The scan-based `clean()` is semantically equivalent.** In `obtain(path)`, the registration key is `metaPath = value.getPath()` (or `DIVIDE_CACHE_KEY` on a miss), so comparing `entry.getValue().getPath()` against the key — with the `NULL` sentinel mapped back to `DIVIDE_CACHE_KEY` — reproduces exactly what `MAPPING` used to return. I confirmed `DIVIDE_CACHE_KEY` and `NULL` are the fields used for the negative-cache path, and that `clean(data.getPath())` is called from `remove` while `clean()` (no arg) is the divide-plugin flush. `WindowTinyLFUMap#entrySet()` delegates to `cache.asMap().entrySet()`, so `removeIf` is backed by Caffeine and works.

The new tests are good: `testMetadataPathCacheIsBounded` inserts twice the bound and asserts `size()` stays under it, and `testRemoveCleansWildcardPathEntries` covers the wildcard (`/home/**`) case that the old `MAPPING` existed for. Adding `@AfterEach` cleanup also fixes cross-test pollution through these static caches, which matters now that a test deliberately fills the cache.

Three things worth a follow-up, none blocking:

1. **`cleanUp()` inside `size()`** forces eviction work on every call. That is what makes the bound assertion deterministic, and `size()` is not on a request path today — but if anything starts calling `size()` per request (metrics, admin endpoints) it becomes a per-call maintenance cost. A comment noting why it is there would help.
2. **1000 may be too small for this particular cache.** `CACHE` is keyed by concrete request path, so a deployment with more than a thousand hot paths will thrash, and every miss falls back to a full scan of `META_DATA_MAP` with `PathMatchUtils.match`. Bounding it is unambiguously better than `Integer.MAX_VALUE`, but `Constants.CACHE_MAX_COUNT` is a generic constant — this cache may deserve its own (larger, or configurable) bound.
3. `initCache(path, value, metaPath)` is public and takes `metaPath` explicitly, while the new `clean()` derives the key from `value.getPath()`. They agree for every call site inside `obtain`, but an external caller passing a `metaPath` that differs from `value.getPath()` would now register an entry that `clean()` cannot find.

PR #7159, opened 2026-09-22
fix(tars): clear companion caches on invalidation

### LINKED ISSUE #6567: [BUG] Tars ApplicationConfigCache: four unbounded ConcurrentHashMaps never evicted (class-loader leak)
## Description
Only the `LoadingCache<String, TarsInvokePrxList> cache` (line 82) has a `maximumSize`. Four companion maps (`ctxPathCache`, `prxClassCache`, `prxParamCache`, `refreshUpstreamCache`) have no maximumSize, no expiry, and no removal path. `invalidate(contextPath)` only calls `cache.invalidate(path)`. `prxClassCache` holds ByteBuddy-generated `Class<?>` objects loaded via `ClassLoadingStrategy.Default.INJECTION`; each retained `Class<?>` pins its class-loader and prevents metaspace GC.

## Location
```
shenyu-plugin-tars/.../cache/ApplicationConfigCache.java:92,94,96,98 (maps), 349-352 (invalidate only touches bounded LoadingCache)
:269 (ClassLoadingStrategy.Default.INJECTION)
```

## Impact
Long-lived deployments that add/remove Tars services (blue-green, ephemeral environments) leak generated Class objects + loaders forever -> eventual `Metaspace OOM`. Worst kind of leak (prevents metaspace reclamation).

## Suggested fix
In `invalidate(contextPath)`, also remove `ctxPathCache.remove(contextPath)` and, for each MetaData in that list, `prxClassCache.remove(metaData.getPath())` and `prxParamCache` entries. Wire `TarsMetaDataHandler.remove` to the full cleanup.

## Related existing issue(s)
None

dengliming:
## Summary
- remove context metadata and upstream snapshots when a Tars context is invalidated
- evict generated proxy classes and method parameter metadata for every removed path
- add regression coverage for all companion cache entries

## Testing
- `mvn -q -pl shenyu-plugin/shenyu-plugin-proxy/shenyu-plugin-rpc/shenyu-plugin-tars -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=ApplicationConfigCacheTest -DfailIfNoTests=false test` (with the module's hard-coded Surefire skip temporarily disabled locally)

Fixes #6567


Aias00 (approved):
APPROVE — this closes a real leak, and the key arithmetic checks out.

`invalidate(contextPath)` previously only called `cache.invalidate(...)` on the Guava `LoadingCache`:

```java
List<MetaData> metaDataList = ctxPathCache.getOrDefault(contextPath, new ArrayList<>());
metaDataList.forEach(metaData -> cache.invalidate(metaData.getPath()));
```

`ctxPathCache`, `prxClassCache`, `prxParamCache` and `refreshUpstreamCache` were never touched, so after a metadata removal the class definition, the method parameter info and the upstream list all stayed resident — stale data plus an unbounded growth path as services registered and deregistered.

The new version removes all four companion entries. I verified the parts that are easy to get wrong:

- **Parameter-cache key prefix is correct.** The cache is written with `getClassMethodKey(clazzName, methodName)` where `clazzName = PrxInfoUtil.getPrxName(metaData)` and `getClassMethodKey` is `String.join("_", className, methodName)`. So entries are `getPrxName(md) + "_" + methodName`, and removing by the prefix `getPrxName(md) + "_"` matches every method of that proxy. (Note `getPrxName` itself embeds the method name, which is why iterating over every `MetaData` under the context path rather than computing one prefix is the right shape.)
- **Thread safety is fine.** All four fields are `ConcurrentHashMap`, so `keySet().removeIf(...)` is a weakly-consistent bulk removal with no `ConcurrentModificationException`. `invalidate` is not on the request hot path, so the O(n) scan over at most `CACHE_MAX_COUNT` entries is acceptable.
- **`remove` instead of `getOrDefault` is right** — the previous code left the `ctxPathCache` entry in place forever.

`testInvalidateRemovesCompanionCaches` is a solid regression test: it populates all four caches through the real key format (`PrxInfoUtil.getPrxName(metaData) + "_" + methodName`), invalidates, and asserts all four are empty plus that the `LoadingCache` hands back a fresh instance. The loader returns `new TarsInvokePrxList(null, null, null)` rather than throwing, so the `assertNotSame` is meaningful.

Two observations, neither blocking:

1. `prxClassCache` is keyed by `metaData.getPath()` while the generated class is named `getPrxName(metaData)` (path **plus** method name), so two methods on the same path collide on one class-cache entry. That is pre-existing and out of scope here, but it is adjacent to what you are fixing.
2. `refreshUpstreamCache` is written with `selectorData.getName()` (`initPrxClass`) and read/removed with a context path. This change is consistent with the existing code, which already assumes the two are equal — just flagging that the invariant is load-bearing and undocumented.

PR #7160, opened 2026-09-22
fix(admin): scope long-poll cache refreshes

### LINKED ISSUE #6569: [BUG] HttpLongPollingDataChangedListener: single-thread scheduler + global synchronized(this) full-cache refresh
## Description
One scheduler thread runs the periodic `refreshLocalCache`, every `DataChangeTask` notification to all long-polling clients, and every `LongPollingClient` timeout future. `checkCacheDelayAndUpdate` takes `synchronized(this)` — a process-wide lock across all groups/namespaces — and inside it calls `super.refreshLocalCache()` which loops all namespaces and rebuilds all eight caches under the lock. Every long-polling compare that detects a stale md5 turns into a full-DB scan holding the global lock, blocking every other concurrent long-polling compare.

## Location
```
shenyu-admin/.../listener/http/HttpLongPollingDataChangedListener.java:111 (new ScheduledThreadPoolExecutor(1, ...)), 261 (synchronized(this) in checkCacheDelayAndUpdate)
```

## Impact
With many gateways holding long polls, one stale md5 triggers a global full-DB rebuild that blocks every other long-poll compare. O(total_config) work per stale-detect under a global lock.

## Suggested fix
Increase scheduler parallelism; narrow the lock to per-namespace; make `refreshLocalCache` namespace-scoped.

## Related existing issue(s)
None

dengliming:
## Summary
- refresh only the stale client namespace instead of rebuilding every namespace
- serialize delayed refreshes per namespace instead of behind one global listener lock
- allow long-poll notifications and timeout work to run on multiple scheduler threads
- drain notification queues before partitioning so parallel batches do not mutate a shared list
- add focused regression coverage

## Testing
- `mvn -q -pl shenyu-admin -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=HttpLongPollingDataChangedListenerTest -DfailIfNoTests=false test`

Fixes #6569


shenyu-admin/src/main/java/org/apache/shenyu/admin/listener/http/HttpLongPollingDataChangedListener.java
```diff
@@ -80,6 +79,8 @@
 @SuppressWarnings("all")
 public class HttpLongPollingDataChangedListener extends AbstractDataChangedListener {
 
+    private static final int SCHEDULER_THREADS = 4;
```
Aias00: Two concerns with widening the scheduler here.

`LongPollingClient#sendResponse` is not idempotent:

```java
void sendResponse(final List<ConfigGroupEnum> changedGroups) {
    if (Objects.nonNull(asyncTimeoutFuture)) {
        asyncTimeoutFuture.cancel(false);
    }
    generateResponse((HttpServletResponse) asyncContext.getResponse(), changedGroups);
    asyncContext.complete();
}
```

A client that watches several groups is registered once per `(namespaceId, group)` key, so the same `LongPollingClient` instance can sit in more than one `clientsMap` queue. Previously the scheduler had a single thread, so two notifications (say PLUGIN and RULE) could never `sendResponse` the same client concurrently — the tasks were serialised.

With `SCHEDULER_THREADS = 4`, two `doRun` batches can now run in parallel, and if they both contain that client you get two concurrent `generateResponse` writes followed by two `asyncContext.complete()` calls. The second `complete()` throws `IllegalStateException`, and the writes can interleave in the response body.

Options:

- make `sendResponse` idempotent (an `AtomicBoolean` guard, or a CAS on a `responded` flag on the client), which is the robust fix and also protects the timeout path; or
- keep this PR focused on the namespace scoping and move the thread-count bump to its own change with its own justification.

Also worth noting: `refreshLocks` is keyed by namespace and never evicted. That is fine in practice (the number of namespaces is small and bounded), but if namespaces can be created dynamically it grows without limit — a `ConcurrentHashMap` of a handful of monitors is not a problem today, just not something that should silently become unbounded later.

Aias00 (changes requested):
REQUEST CHANGES — the namespace scoping is a clear win; the scheduler change bundled with it is the problem.

**The part I want to keep is genuinely good.** `checkCacheDelayAndUpdate` used to call `super.refreshLocalCache()`, which looped over *every* namespace and refreshed plugin/auth/rule/selector/metadata/proxy-selector/discovery/AI-key caches for all of them — so a single long-polling client in one namespace could trigger a full refresh of the whole cluster's config. Replacing it with `refreshLocalCache(serverCache.getNamespaceId())` is the right fix, and replacing `synchronized (this)` with a per-namespace monitor means two namespaces no longer serialise against each other. I checked: `synchronized (this)` appears exactly once in this class (line 261), so nothing else depends on that monitor and scoping it is safe.

Extracting `refreshLocalCache(String)` in `AbstractDataChangedListener` while keeping the no-arg version as a loop over namespaces is clean, and `testClientNewerRefreshesOnlyItsNamespace` pins the behaviour properly by overriding the hook and asserting which namespace was refreshed.

The `doRun` batching change is also fine — `drainTo` up front plus dropping `iter.remove()` preserves the existing remove-then-send order.

**My objection is `SCHEDULER_THREADS = 4`.** `LongPollingClient#sendResponse` does `asyncContext.complete()` with no idempotency guard, and one client can be registered in several `clientsMap` queues (one per group it watches). With a single-threaded scheduler those sends were always serialised; with four threads two batches can hit the same client concurrently, producing interleaved response writes and a second `asyncContext.complete()` that throws `IllegalStateException`. Details are in the inline comment.

So please either:

- add an idempotency guard to `sendResponse` (an `AtomicBoolean`/CAS on a `responded` flag — it also protects the timeout path, which has the same race today), or
- drop the thread bump from this PR. It is not related to "scope long-poll cache refreshes", and the scoping fix stands on its own.

Smaller note: `testLongPollingSchedulerUsesMultipleThreads` asserts `scheduler.getCorePoolSize() == 4`, i.e. it pins an implementation detail. If you keep the bump, that is an acceptable smoke test; just be aware it will need updating the moment the constant changes.

Aias00 (approved):
## Review (resolving my earlier request-changes)

The refresh is now correctly scoped per-namespace: `refreshLocalCache()` delegates to a new `refreshLocalCache(namespaceId)`, and the global `synchronized (this)` is replaced by a per-namespace `refreshLock` (`refreshLocks.computeIfAbsent(namespaceId, ...)`). `testClientNewerRefreshesOnlyItsNamespace` proves only the changed namespace is refreshed.

On the double-`complete()` / `IllegalStateException` concern I raised earlier: I checked the surrounding code and it isn't reachable here. `clientsMap` is keyed by `namespaceId` (one queue per client-namespace), and `doRun` `drainTo`s the queue before calling `sendResponse` — so the same `LongPollingClient` (and its `asyncContext`) is completed at most once per cycle. The enhanced-for loop replacing the `Iterator` is safe because `drainTo` already removed the elements. So the per-namespace lock is the right, sufficient fix.

The scheduler is a single-threaded `ScheduledThreadPoolExecutor`, so partitioned `doRun` calls also run sequentially. Approving. (If `build (shenyu-integrated-test-http)` stays red it's almost certainly infra — this change is admin-only — please re-run.)

shenyu-admin/src/main/java/org/apache/shenyu/admin/listener/http/HttpLongPollingDataChangedListener.java
```diff
@@ -258,12 +260,13 @@ private boolean checkCacheDelayAndUpdate(final ConfigDataCache serverCache, fina
         if (latest != serverCache) {
             return !StringUtils.equals(clientMd5, latest.getMd5());
         }
```
Aias00, line 262: 🟢 Verified this resolves the earlier concern without an explicit `sendResponse` guard: `clientsMap` is keyed by `namespaceId` (one queue per client-namespace) and `doRun` `drainTo`s the queue before `sendResponse`, so the same `asyncContext` is completed at most once per cycle. The per-namespace `refreshLock` is the right, sufficient fix.

PR #7161, opened 2026-09-22
perf(loadbalancer): cache consistent hash rings

### LINKED ISSUE #6571: [BUG] HashLoadBalancer rebuilds consistent-hash ring + allocates MD5 MessageDigest per virtual node per request
## Description
`doSelect` allocates a fresh `ConcurrentSkipListMap` and, for every upstream x 5 virtual nodes, calls `hash("SHENYU-" + upstream.getUrl() + "-HASH-" + i)` which does `MessageDigest.getInstance("MD5")` (JCA lookup + new instance), `key.getBytes(UTF_8)`, and `md5.digest()` (new `byte[]`). The ring is never cached; it is fully rebuilt on every selection even though the upstream set only changes on health-check/admin events.

## Location
```
shenyu-loadbalancer/.../spi/HashLoadBalancer.java:44-56 (doSelect), 58-77 (hash)
```

## Impact
Per hash-LB request: 1 SkipListMap + 5*N MD5 instances + 5*N String concatenations + 5*N `byte[]` digests. For N=10 upstreams that's 50 MD5 hashes/request. Very high CPU + allocation on the sticky/consistent-hash path; scales linearly with RPS and upstream count.

## Suggested fix
Build and cache the ring (e.g. `ConcurrentHashMap<String, TreeMap<Long,Upstream>>` keyed by selectorId, invalidated when `UpstreamCacheManager.submit`/health-check mutates the list). Use a `ThreadLocal<MessageDigest>` or a faster non-crypto hash (murmur3/xxHash). Precompute virtual-node keys when the upstream list changes.

## Related existing issue(s)
None

dengliming:
## Summary
- cache bounded consistent-hash rings by ordered upstream URL set
- store upstream indexes in cached rings so refreshed upstream objects are returned
- reuse one MD5 digest per worker thread instead of allocating per virtual node
- add regression coverage for ring reuse and refreshed instances

## Testing
- `mvn -q -pl shenyu-loadbalancer -am -DskipTests=false -Dcheckstyle.skip=false -Dtest=HashLoadBalancerTest,HashLoadBalanceTest -DfailIfNoTests=false test`

Fixes #6571


shenyu-loadbalancer/src/main/java/org/apache/shenyu/loadbalancer/spi/HashLoadBalancer.java
```diff
@@ -39,33 +43,38 @@ public class HashLoadBalancer extends AbstractLoadBalancer {
...
     @Override
     public Upstream doSelect(final List<Upstream> upstreamList, final LoadBalanceData data) {
-        final ConcurrentSkipListMap<Long, Upstream> treeMap = new ConcurrentSkipListMap<>();
-        upstreamList.forEach(upstream -> IntStream.range(0, VIRTUAL_NODE_NUM).forEach(i -> {
-            long addressHash = hash("SHENYU-" + upstream.getUrl() + "-HASH-" + i);
-            treeMap.put(addressHash, upstream);
-        }));
+        final List<String> ringKey = new ArrayList<>(upstreamList.size());
+        upstreamList.forEach(upstream -> ringKey.add(upstream.getUrl()));
+        final SortedMap<Long, Integer> treeMap = ringCache.computeIfAbsent(List.copyOf(ringKey), this::buildRing);
```
Aias00, line 55: Two notes on the cache itself.

**1. `computeIfAbsent` here is a non-atomic check-then-act.** `WindowTinyLFUMap` extends `AbstractMap` and only overrides `put/get/remove/clear/size/entrySet`, so this resolves to the default `Map.computeIfAbsent`:

```java
if ((v = get(key)) == null) {
    V newValue;
    if ((newValue = mappingFunction.apply(key)) != null) { put(key, newValue); return newValue; }
}
```

The underlying Caffeine cache makes the individual `get`/`put` safe, so there is no corruption, but concurrent callers can each build a ring and the last `put` wins. That is harmless here because `buildRing` is deterministic — just wasted work on first touch. Worth a comment so nobody later adds something non-idempotent to `buildRing`.

**2. The bounds are coupled, and 1000 rings is not obviously the right number.** `Constants.CACHE_MAX_COUNT` is 1000, and each entry holds `upstreams × VIRTUAL_NODE_NUM` map entries. For a deployment with, say, 200 upstreams behind one service that is 1000 entries per ring, and up to 1000 distinct upstream sets retained per load-balancer instance. That is bounded, but it is a much larger steady-state footprint than the previous "build it and throw it away" behaviour. Given the ring is keyed on the full URL list, any churn in the upstream set (a single node flapping) produces a new key each time, so the working set is driven by how often the upstream list changes.

Neither is a correctness problem — just please make the sizing deliberate rather than reusing `CACHE_MAX_COUNT` because it was there.

Aias00 (approved):
APPROVE — good optimisation, and the stale-instance fix is the part I care about most.

The original rebuilt a `ConcurrentSkipListMap` on every single `doSelect`, which meant `upstreams × 5` MD5 hashes per request, plus a fresh `MessageDigest.getInstance("MD5")` for each one — that was expensive on the request path. Caching the ring is the right call.

The more important half is switching the ring value from `Upstream` to the **index** into the URL list. Caching `Upstream` instances would have pinned whatever objects were live when the ring was built, so every upstream refresh (status, warmup, weight, last-selected timestamp) would have been invisible to the load balancer, and callers would have mutated stale objects — the classic cache-the-instance bug. Storing indices and resolving against the caller's `upstreamList` keeps that correct, and `shouldReuseRingWithoutReturningStaleUpstreamInstances` pins exactly that (`assertNotSame` on the selected instance between two equal-but-distinct lists, and `assertSame` against the list that was passed in). Good test.

Also checked: the key is `List.copyOf(upstreamUrls)`, so two lists with the same URLs in the same order reuse a ring and the stored indices stay valid; a different order or membership produces a different key and a new ring. `LoadBalanceData#ip` defaults to `"127.0.0.1"` so `hash(data.getIp())` cannot NPE in the new test.

I left one inline comment with two things to think about:

- `computeIfAbsent` on `WindowTinyLFUMap` is the default `Map` implementation, i.e. a non-atomic `get` then `put`. Caffeine makes the individual operations safe so nothing corrupts, and `buildRing` is deterministic so a duplicate build is only wasted work — but it deserves a comment before someone adds something non-idempotent there.
- The ring cache is bounded by `Constants.CACHE_MAX_COUNT` (1000) while each entry holds `upstreams × 5` entries, and any churn in the upstream set mints a new key. Bounded, but a very different steady-state footprint from "build and discard" — worth sizing deliberately rather than reusing that constant by default.

One more small thing: the `ThreadLocal<MessageDigest>` is never cleared. In practice shenyu runs a bounded set of event-loop threads so this is a fixed, small cost, and the retained `MessageDigest` comes from the JDK provider rather than a plugin classloader, so it is not a reload leak — just noting it since a `static ThreadLocal` on a shared SPI class tends to outlive expectations.

PR #7162, opened 2026-09-22
fix(httpclient): bound the default connection pool

### LINKED ISSUE #6573: [BUG] Default outbound connection pool is UNBOUNDED (ELASTIC, maxConnections=Integer.MAX_VALUE) with no eviction defaults
## Description
The default outbound proxy pool type is ELASTIC with `maxConnections=Integer.MAX_VALUE` and no background eviction. The gateway opens a new connection for every concurrent request with no ceiling, and idle connections to dead backends are never reaped.

## Location
```
shenyu-spring-boot-starter/.../httpclient/HttpClientFactory.java:192-197 (buildElasticConnectionPool)
HttpClientProperties.java:401 (default PoolType type = ELASTIC)
HttpClientFactory.java:160-162 (maxIdleTime/maxLifeTime/evictionInterval default null)
shenyu-bootstrap/src/main/resources/application.yml (httpclient.pool block commented out)
```

## Impact
Under a traffic burst the gateway opens unbounded TCP connections (FD exhaustion) and never evicts idle connections to decommissioned upstreams. Combined with keepAlive=false, connections are used once and discarded, churning FDs/ephemeral ports.

## Suggested fix
Change default to `FIXED` with a sane `maxConnections`, or cap the ELASTIC pool; ship default `maxIdleTime`/`evictionInterval` and uncomment/tune the pool block in `application.yml`.

## Related existing issue(s)
#6298/#6270/#5987/#6018 are FD leaks (resources not released); this is the pool itself being mis-sized/unbounded — a distinct configuration defect.

dengliming:
Fixes #6573.

Changes the default Reactor Netty connection pool from ELASTIC to FIXED so outbound connections are bounded by the existing maxConnections default. Explicit ELASTIC configuration remains supported.

Tests:
- HttpClientPropertiesTest
- HttpClientPluginConfigurationTest
- Checkstyle


shenyu-plugin/shenyu-plugin-httpclient/src/main/java/org/apache/shenyu/plugin/httpclient/config/HttpClientProperties.java
```diff
@@ -396,9 +396,9 @@ public void setMaxInMemorySize(final Integer maxInMemorySize) {
     public static class Pool {
 
         /**
-         * Type of pool for HttpClient to use, defaults to ELASTIC.
+        * Type of pool for HttpClient to use, defaults to FIXED.
          */
-        private PoolType type = PoolType.ELASTIC;
+        private PoolType type = PoolType.FIXED;
```
Aias00, line 401: Switching the default here changes more than the pool type — it also switches on an unbounded pending queue with a 45s wait.

`HttpClientFactory#buildConnectionProvider` routes to two very different configurations:

```java
// ELASTIC (old default)
builder.maxConnections(Integer.MAX_VALUE)
       .pendingAcquireTimeout(Duration.ofMillis(0));

// FIXED (new default)
builder.maxConnections(pool.getMaxConnections())            // default 500
       .pendingAcquireTimeout(Duration.ofMillis(pool.getAcquireTimeout()))  // default 45000
       .pendingAcquireMaxCount(-1);                          // unlimited pending
```

So a deployment that upgrades without touching any config goes from "open another connection" to "queue without limit and hang for up to 45 seconds" once 500 concurrent connections are in use — and `pendingAcquireMaxCount(-1)` means the queue itself is unbounded, so this fails by accumulating latency and memory rather than by rejecting.

Bounding the pool is the right instinct (an ELASTIC pool is an unbounded resource), but the two defaults have to be chosen together. Concretely:

- lower the default `acquireTimeout` for a gateway — 45s exceeds almost every client and upstream timeout, so a saturated pool turns into requests that hang long after the caller gave up; something in the 1-5s range fails fast and lets the load balancer/retry logic react;
- and/or give `pendingAcquireMaxCount` a real bound instead of `-1`, so saturation produces a rejection instead of an ever-growing queue.

Either is fine, but please make it an explicit decision rather than inheriting Reactor Netty's 45s default. This also deserves a release note: it is a behavioural default change that affects every deployment that does not configure `shenyu.httpclient.pool.*`.

Aias00 (changes requested):
REQUEST_CHANGES (non-blocking to re-approve — one decision to make first).

I agree with the goal: an `ELASTIC` pool (`maxConnections(Integer.MAX_VALUE)`) is an unbounded resource, and a gateway should not default to that. The one-line change and the two updated tests are clean.

My concern is what the new default brings with it. `HttpClientFactory#buildFixedConnectionPool` applies:

```java
builder.maxConnections(pool.getMaxConnections())                            // 500
       .pendingAcquireTimeout(Duration.ofMillis(pool.getAcquireTimeout()))  // 45000
       .pendingAcquireMaxCount(-1);                                          // unlimited
```

versus the ELASTIC path, which used `maxConnections(Integer.MAX_VALUE)` and `pendingAcquireTimeout(0)`.

So any deployment upgrading without touching `shenyu.httpclient.pool.*` goes from "open another connection" to "queue without limit, wait up to 45 seconds" once 500 connections are in flight. `pendingAcquireMaxCount(-1)` makes the queue unbounded as well, which means saturation manifests as unbounded latency and retained pending subscriptions rather than as a rejection — the exact failure mode a gateway most wants to avoid. And 45s is longer than virtually every client and upstream timeout, so those requests hang long after the caller has already given up.

What I'd like before merge:

- **pick a gateway-appropriate `acquireTimeout` default** (something in the 1-5s range, or whatever you judge right) instead of inheriting Reactor Netty's 45s, and/or
- **bound `pendingAcquireMaxCount`** so saturation rejects rather than queues forever;

and in either case call this out in the release notes — it is a behavioural default change for every deployment that does not configure the pool explicitly.

If you'd rather keep the 45s default as-is and just document it, say so in a reply and I'll switch this to an approve; I just don't want it to land silently.

Minor, unrelated: the javadoc on `Pool#name` still says "defaults to proxy" while the field is `"shenyu-proxy"`. Pre-existing, but you are already touching this class.

Aias00 (approved):
## Review (resolving my earlier request-changes)

Both points from the earlier review are addressed:
- `buildFixedConnectionPool` no longer calls `.pendingAcquireMaxCount(-1)`, so the fixed pool now uses Reactor Netty's **bounded** default (`2 × maxConnections`) instead of an unbounded pending-acquire queue. The new `testFixedConnectionPoolUsesBoundedPendingQueue` asserts `pendingAcquireMaxCount` is not set.
- `acquireTimeout` defaults to `3000L` (3s) instead of the Reactor Netty default, and `RELEASE-NOTES.md` documents the default flip (`ELASTIC` → `FIXED`) plus the `shenyu.httpclient.pool.type=ELASTIC` escape hatch.

Tests updated accordingly. Approving.

